PR #1, opened 2017-08-21
Integrate repo-tools.

jmdobry:
This is the initial integration of repo tools into this library. I released 2.0.0-beta.1 (it's now at 2.0.0-beta.3) of the repo tools, and we can use its 2.0.0 branch to refine it as we discover more pieces of these repos that should be automated.

Currently repo-tools:

- Generates the root README.md file
- Generates the samples/README.md file
- Generates the CODE_OF_CONDUCT.md file
- Generates the LICENSE file.
- Lints files using semistandard (samples file only for now)


jmdobry:
Looks like I'll need to produce a Node4-compatible build of repo tools.

README.md
```diff
@@ -1,161 +1,157 @@
...
-### Instantiate a client
+**Table of contents:**
 
-```js
-var gcs = require('@google-cloud/storage')({
-  projectId: 'grape-spaceship-123',
-  keyFilename: '/path/to/keyfile.json'
-});
-```
+* [QuickStart](#quickstart)
```
jmdobry: Note, this was modeled after: https://cloud.google.com/natural-language/docs/quickstart-client-libraries

LICENSE
```diff
@@ -187,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2014-2017 Google Inc.
+   Copyright 2017 Google Inc.
```
lukesneeringer: Is it possible to set this as `2014-${current}`?

jmdobry: Will do: GoogleCloudPlatform/nodejs-repo-tools#22

package.json
```diff
@@ -63,15 +67,16 @@
     "mime-types": "^2.0.8",
     "once": "^1.3.1",
     "pumpify": "^1.3.3",
+    "semistandard": "^11.0.0",
```
stephenplusplus: Was this meant to be a devDep?

README.md
```diff
@@ -1,161 +1,157 @@
...
+[enable_api]: https://console.cloud.google.com/flows/enableapi?apiid=storage-api.googleapis.com
+[auth]: https://cloud.google.com/docs/authentication/getting-started
+
+### Installing the client library
+
+    npm install --save @google-cloud/storage
+
+### Using the client library
+
+    // Imports the Google Cloud client library
```
stephenplusplus: I think this file was generated in some form-- if that's the case, instead of using indents here, is it possible to have the generator wrap the code blocks with ticks, and in this case, ```js so that it's syntax highlighted?

package.json
jmdobry: Yes

README.md
jmdobry: Will do: GoogleCloudPlatform/nodejs-repo-tools#23

PR #11, opened 2017-09-12
Upgrade repo-tools and regenerate.

jmdobry:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)
+
+- [ ] Tests and linter pass
+- [ ] Code coverage does not decrease (if any source code was changed)
```
stephenplusplus: This one is a bit of duplication from our coverage reporter that will comment if it decreases. The same almost applies with the first one (test and lint failures), except GitHub doesn't alert the OP if a CI task fails. 

One thing that could be helpful is instructing the user how these checks are performed, i.e. "Run npm lint"

jmdobry: I don't have strong opinions on this, but the file to change is here: https://github.com/GoogleCloudPlatform/nodejs-repo-tools/blob/2.0.0/templates/pr_template.tpl

lukesneeringer: Same. No strong opinion; feel free to edit as you see fit. :-)

PR #13, opened 2017-09-20
Add ESLint and Prettify

lukesneeringer:
@jmdobry PR to nodejs-repo-tools will be be coming this afternoon.

Commit-by-commit might be an easier sanity check path.


lukesneeringer:
A couple things I want to call out:

  * I moved the samples tests back into `samples/package.json`, and I have the `samples-tests` command in the main repo cd in and call it. This was partially to make the linter happy (it thought that things like ava were not installed because they were not in `samples/package.json`) but I actually think the separation makes a lot of sense.
  * I <3 prettier. (There is an Atom plugin to just run it on save, which is really nice.)
  * I exempted the samples tests from the Node 4 and Node 6 compatibility checks. (In principle I am willing to do that for system tests also, but at the moment it is not necessary.)

package.json
```diff
@@ -67,39 +67,43 @@
...
     "semistandard": "^11.0.0",
-    "sinon": "^3.2.1",
     "tmp": "^0.0.33",
-    "uuid": "^3.0.1"
+    "uuid": "^3.1.0"
   },
   "scripts": {
     "docs": "node_modules/.bin/jsdoc -c .jsdoc.js",
-    "lint": "npm run samples-lint",
+    "lint": "eslint src/ samples/ system-test/ test/",
```
jmdobry: Note that we don't actually have to modify repo-tools itself to change the linter. We can just do the following:

1. Add the following to `.cloud-repo-tools.json`:

```
"lint": {
  "cmd": "eslint"
}
```

2. Change the `lint` script to the following:

```
"lint": "repo-tools lint src/ samples/ system-test/ test/"
```

samples/acl.js
```diff
@@ -267,27 +258,26 @@ function printFileAclForUser (bucketName, filename, userEmail) {
 
   const options = {
     // Specify the user
-    entity: `user-${userEmail}`
+    entity: `user-${userEmail}`,
```
jmdobry: What's the deal with these trailing commas?

lukesneeringer: I set prettify to `trailingComma: es5`. That is an item that is configurable in prettify.

jmdobry: What I meant is, can we undo the trailing commas?

lukesneeringer: Hmm, I do not consider it binding for open source, but:
https://google.github.io/styleguide/jsguide.html#features-arrays-trailing-comma
https://google.github.io/styleguide/jsguide.html#features-objects-use-trailing-comma

jmdobry: alright

src/acl.js
```diff
@@ -264,22 +264,25 @@ Acl.prototype.add = function(options, callback) {
...
-    json: {
-      entity: options.entity,
-      role: options.role.toUpperCase()
-    }
-  }, function(err, resp) {
-    if (err) {
-      callback(err, null, resp);
-      return;
-    }
+  this.request(
```
stephenplusplus: Is there a style rule that forces this that we can ignore? The way that was removed is generally what I've seen in other codebases. I'd rather factor out the object argument so we can at least end up with:

```js
this.request(reqOpts, function(err, resp) {
  // ...
});
```

src/bucket.js
```diff
@@ -1691,7 +1730,8 @@ Bucket.prototype.upload = function(localPath, options, callback) {
   }
 
   function upload() {
-    fs.createReadStream(localPath)
+    fs
```
stephenplusplus: This is a bit unpleasing to the eye.

src/file.js
```diff
@@ -1048,9 +1065,7 @@ File.prototype.download = function(options, callback) {
       .on('error', callback)
       .on('finish', callback);
   } else {
-    fileStream
-      .on('error', callback)
-      .pipe(concat(callback.bind(null, null)));
+    fileStream.on('error', callback).pipe(concat(callback.bind(null, null)));
```
stephenplusplus: This is less readable to me than the typical notation where new lines help show the directions a stream goes and which stream gets the event handlers.

stephenplusplus:
One of my favorite code style rules is "be consistent", and this definitely does that. My only issue is some of the style choices I've pointed out. Is there any part of these new tooling processes that auto-formats the code? If that's the case, I don't mind the new rules so much. But if not, the new approach feels a bit robotic, which can deflect contributions (at some point, they'll get sick of a yelling linter and just stop), or at best, makes the review process longer-- they likely won't follow the rules they haven't seen elsewhere, which leaves an extra step for us to do the style-conformity for them.

jmdobry:
I believe the code is now being formatted by [prettier](https://github.com/prettier/prettier), which does have a small amount of configurability, should we desire.

package.json
```diff
@@ -67,39 +67,43 @@
     "mime-types": "^2.0.8",
     "once": "^1.3.1",
     "pumpify": "^1.3.3",
+    "safe-buffer": "^5.1.1",
```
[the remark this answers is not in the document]
jmdobry: It's actually used in the sample, I think it can be removed from this top-level package.json file.

system-test/storage.js
```diff
@@ -18,6 +18,7 @@
 
 var assert = require('assert');
 var async = require('async');
+var Buffer = require('safe-buffer').Buffer;
```
stephenplusplus: @jmdobry here is where `safe-buffer` is used in the non-samples code.

package.json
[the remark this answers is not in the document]
lukesneeringer: It's used in the actual code too.

@stephenplusplus: According to ESLint, `new Buffer(something)` is deprecated, and now should be `Buffer.from(something)`. However, that syntax is only supported in Node 4.5.0 and above.

`safe-buffer` uses the Node implementation if available, and provides its own for Node < 4.5.0.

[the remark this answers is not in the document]
lukesneeringer: Note: That means we can remove `safe-buffer` in April once we go to `engines: '>6'` should we desire (and remember).

src/bucket.js
lukesneeringer: Concur.

src/acl.js
lukesneeringer: Making the change you described would get that result. (Prettier tries to maximize what it will fit on a line.) I actually would prefer the code you wrote in your comment over _both_ what was there before _and_ what Prettier spit out. (I agree that what was there before was better than what Prettier spit out.)

package.json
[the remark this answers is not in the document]
stephenplusplus: Ah, thanks. Is it used in the source, or just the system test? If just the tests, we can move it to `devDependencies`.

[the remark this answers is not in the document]
lukesneeringer: Both.

PR #16, opened 2017-09-21
Perform the entire CircleCI workflow on tags.

lukesneeringer:
This _should_ make it such that CircleCI runs on tag builds (and thus can push to npm).


lukesneeringer:
Nixing in favor of not being on a fork.

PR #17, opened 2017-09-21
Perform the entire CircleCI workflow on tags.

lukesneeringer:
This _should_ make it such that CircleCI runs on tag builds (and thus can push to npm).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,22 +20,40 @@ workflows:
   version: 2
   tests:
     jobs:
-      - node4
-      - node6
-      - node7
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -107,7 +119,7 @@ jobs:
             cd samples/
             npm install
             npm link @google-cloud/storage
-            cd ..    
+            cd ..
```
stephenplusplus: This is my favorite part.

lukesneeringer: :-p

PR #28, opened 2017-09-21
Update samples.

jmdobry:


jmdobry:
They pass.

samples/requesterPays.js
```diff
@@ -0,0 +1,217 @@
...
+
+'use strict';
+
+function enableRequesterPays(bucketName) {
+  // [START storage_enable_requester_pays]
+  // Imports the Google Cloud client library
+  const Storage = require(`@google-cloud/storage`);
+
+  // The name of the bucket to enable requester-paying for, e.g. "my-bucket"
+  // const bucketName = "my-bucket";
```
stephenplusplus: Supposed to be commented?

samples/requesterPays.js
```diff
@@ -0,0 +1,217 @@
...
+ * For more information, see the README.md under /storage and the documentation
+ * at https://cloud.google.com/storage/docs.
+ */
+
+'use strict';
+
+function enableRequesterPays(bucketName) {
+  // [START storage_enable_requester_pays]
+  // Imports the Google Cloud client library
+  const Storage = require(`@google-cloud/storage`);
```
stephenplusplus: This might be old-school JS conventions, but I think CaptialCase was meant for constructors, thus you have to use `new` in front of them. The tooling and style linters have come a long way, so ignore me if they ignore it now, too.

samples/system-test/requesterPays.test.js
```diff
@@ -0,0 +1,123 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const fs = require(`fs`);
```
stephenplusplus: Just curious-- is this the new thing to do, using the ticks in place of all string quote marks?

samples/requesterPays.js
jmdobry: Yes. If you execute the `requesterPays.js` file, then `bucketName` is passed into the `enableRequesterPays` function on line 26, ala `node requesterPays.js enable "some-bucket-name"`.

Someone looking at the sample on [cloud.google.com](https://cloud.google.com) won't see the wrapper function, and even if they did, just seeing a `bucketName` argument doesn't tell them very much. So, for those looking at the sample in the docs, who might copy and paste the sample (without the wrapper function) into their app, all they have to do us uncomment the commented variable initializations to make it work. Those comments also serve to give users examples of real values.

It's a larger conversation whether the samples should be useful command line programs. If we were to do away with that, then we could do away with the wrapper functions and just hard-code the variable values. That would, however, make the samples harder to test. The samples' tests in their current form double as system tests for the client library.

samples/requesterPays.js
jmdobry: The way our libraries are implemented, `Storage` _is_ a constructor. `new Storage()` works just fine. If I had documented the client libraries, I would have used `new` when creating `Storage` instances.

Consider this:

```js
// what is "storage"?
const storage = require('@google-cloud/storage');

// oh, I guess it's a factory function, which gives me...something
const client = storage();
```
and compare to:
```js
// Capital "Storage", must be a class
const Storage = require('@google-cloud/storage');

// Okay, so I make a new instance
storage instanceof Storage; // true

// And if it's a class, I ought to be able to do this
class MyStorage extends Storage {
  getBuckets(options, callback) {
    // do something custom
    return super.getBuckets(options, callback);
  }
}
```

If/when we switch to ES2015 classes, then the class constructor will no longer be able to serve as a factory function:

```js
// TypeError: Class constructor Storage cannot be invoked without 'new'
const storage = Storage();
```

samples/system-test/requesterPays.test.js
jmdobry: In our tests, we found ourselves doing a lot of string interpolation, as well as putting both single and double quotes in strings, so we found it easier to simply use the ES2015 string templates throughout the tests.

I've seen JS devs starting to use the string templates all over, calling them the new, better JavaScript strings.

samples/requesterPays.js
stephenplusplus: I'm with you on all points, but I'm looking at this in the examples I've found:

```js
// Instantiates a client
const storage = Storage({
  projectId: projectId,
});
```

So, there's just no `new`, and I think that's what we want?

samples/system-test/requesterPays.test.js
stephenplusplus: Cool, thanks!

PR #42, opened 2017-09-28
handle validation at the same rate as destination pipe

### LINKED ISSUE #24: storage - readStream hangs on long running time pipe
_From @tomgrossman on August 26, 2017 9:49_

#### Environment details
  - OS: Ubuntu 16.04
  - Node.js version: 6.11.2
  - npm version: 3.10.10
  - google-cloud-node version: @google-cloud/storage@1.2.1 

Some more details:
The file size I'm reproducing the issue on is ~2.3 gb.
The issue is happening **only** when using **pipe** and **only** when the download speed is below 5 mb\s.
When the download speed is faster or when using `.on('data', (data) => res.write(data))`, there is no problem.

#### Steps to reproduce

1. require `@google-cloud/storage`
2. use the following route function, where bucketFile is a pure file instance from the storage lib:

```javascript
function DownloadOva (req, res) {
    console.log(new Date() + ' start');

    res.on('finish', function () {
        console.log(new Date() + ' finish');
    });

    let bucketFile = OvaFileHelpers.GetBucketFile('agent_ova/agent.ova');
    bucketFile.createReadStream({validation: false})
        .on('error', (err) => {
            console.log(new Date() + ' error: ' + err.stack);
        })
        .on('response', (streamResponse) => {
            console.log(new Date() + ' streamResponse');
            res.setHeader('Content-Length', streamResponse.headers['content-length']);
            res.setHeader('Content-Type', streamResponse.headers['content-type']);
            res.setHeader('Content-disposition', 'attachment; filename="agent_download.ova"');
        })
        .on('end', () => {
            console.log(new Date() + ' end');
            res.end();

            return true;
        })
        // .on('data', (data) => {
        //     Response.write(data);
        // })
        .pipe(res);
}
```
3. limit the download speed in chrome using the throttling option in the network developer tools to 5000 kb/s
4. call the route

**Expected results**:
all console logs outputs should be written to console.
**Actual results**:
The file is downloaded successfully
the 'end' and 'finish' events are never called (no console log outputs and doesn't get into the breakpoint). 
Can't call the route again until restarting the application

**more details**:
When the download speed is not limited, it works fine (on high speed connection). but some users with low speed connections, faced this issue (that's why I reproduced this issue).
When using the commented `.on('data' ...` and commenting out the `pipe`, it also works fine also on slow download.

It seems that when the `pipe` takes too long (seems like 3-4 minutes), it looses all the event listening functions and gets hangs even though the download finish successfully on the client side.

Thanks!

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2569_

stephenplusplus:
Fixes #24 

Carried over from https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2581

### DIFF
--- a/src/file.js
+++ b/src/file.js
@@ -456,6 +456,8 @@ File.prototype.createReadStream = function(options) {
   var self = this;
   var rangeRequest = is.number(options.start) || is.number(options.end);
   var tailRequest = options.end < 0;
+
+  var validateStream; // Created later, if necessary.
   var throughStream = streamEvents(through());
 
   var crc32c = true;
@@ -484,8 +486,11 @@ File.prototype.createReadStream = function(options) {
   // returned to the user.
   function makeRequest() {
     var reqOpts = {
+      forever: false,
       uri: '',
-      gzip: true,
+      headers: {
+        'Accept-Encoding': 'gzip',
+      },
       qs: {
         alt: 'media',
       },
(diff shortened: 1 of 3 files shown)


src/file.js
```diff
@@ -503,27 +508,41 @@ File.prototype.createReadStream = function(options) {
...
+        throughStream.emit('response', res);
+        common.util.handleResp(null, res, null, onResponse);
+      })
+      .resume();
+
+    throughStream.on('error', () => {
+      // An error can occur before the request stream has been created
+      // (during authentication)
+      if (requestStream.abort) {
+        requestStream.abort();
```
callmehiphop: I get lost around here, why do we need to call abort? It seems like if an error occurred it was either because of authentication (in which case there would be nothing to abort) or a request error, which the request failed so I'm not sure what's left to abort?

stephenplusplus: Old code, don't remember for sure. I'll remove it.

stephenplusplus: Done.

PR #49, opened 2017-10-03
Upgrade repo-tools and regenerate scaffolding.

jmdobry:


LICENSE
```diff
@@ -186,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2014-2017 Google Inc.
+   Copyright [yyyy] [name of copyright owner]
```
stephenplusplus: Oops.

jmdobry: No, it turns out this is how the license file is supposed to appear.

stephenplusplus: Attribution implied by the publisher?

stephenplusplus: Oh, I see now. This is the bottom of the file talking about how to attribute yourself when you copy the files. *I* was copywrong.

jmdobry: It was news to me too. I learned about this during some discussions with the folks who own https://opensource.google.com/docs/releasing/#checklist

PR #61, opened 2017-10-16
Fix docs typo

Crazometer:
Change `Fileile` to `File`

- [ ] Tests and linter pass


stephenplusplus:
Thanksanks!

PR #63, opened 2017-10-21
Kiricon/upload from url

### LINKED ISSUE #58: Feature Request: Upload files to a bucket by passing url in place of a local path.
Create a new method for the `Bucket` class called `uploadFromURL`. Said method would work the same way as the `upload` method, but in place of a local path would take a url (`http://exapmle.com/img.jpg`). 

This eliminates the need for a user to download the file, upload it, then delete it him or her self. 

If approved I would build the feature my self and submit a pull request.

Kiricon:
Fixes #58 

- [x] Tests and linter pass

A new method called `uploadFromURL` has been added to the `Bucket` object.
The method does the same thing the `upload` method does, save the `localpath` argument is now meant to be a url string.

Due to the file being remotely hosted, the `option`  property `resumable` it not being taken in to account.

I'm sure of the best way to go about writing them up considering the majority of shared functionality between the `upload` and `uploadFromURL` methods.

Let me know what I can improve and I'll be glad to make the change. This is a feature I've been needing for a very long time.

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -27,6 +27,7 @@ var path = require('path');
 var snakeize = require('snakeize');
 var util = require('util');
 var request = require('request');
+var url = require('url');
 
 var Acl = require('./acl.js');
 var File = require('./file.js');
@@ -2241,7 +2242,11 @@ Bucket.prototype.upload = function(pathString, options, callback) {
     });
   } else {
     // Resort to using the name of the incoming file.
-    newFile = this.file(path.basename(pathString), {
+    var destination = path.basename(pathString);
+    if (isURL) {
+      destination = new url.parse(pathString).pathname;
+    }
+    newFile = this.file(destination, {
       encryptionKey: options.encryptionKey,
     });
   }
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -2048,7 +2048,7 @@ Bucket.prototype.setUserProject = function(userProject) {
  * @see [Upload Options (Simple or Resumable)]{@link https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#uploads}
  * @see [Objects: insert API Documentation]{@link https://cloud.google.com/storage/docs/json_api/v1/objects/insert}
  *
- * @param {string} localPath The fully qualified path to the file you wish to
+ * @param {string} path The fully qualified path or url to the file you wish to
  *     upload to your bucket.
  * @param {object} [options] Configuration options.
  * @param {string|File} [options.destination] The place to save
@@ -2056,7 +2056,7 @@ Bucket.prototype.setUserProject = function(userProject) {
  *     using the string as a filename. When given a File object, your local file
  *     will be uploaded to the File object's bucket and under the File object's
  *     name. Lastly, when this argument is omitted, the file is uploaded to your
- *     bucket using the name of the local file.
+ *     bucket using the name of the local file or the path of the url relative to it's domain.
  * @param {string} [options.encryptionKey] A custom encryption key. See
  *     [Customer-supplied Encryption Keys](https://cloud.google.com/storage/docs/encryption#customer-supplied).
  * @param {boolean} [options.gzip] Automatically gzip the file. This will set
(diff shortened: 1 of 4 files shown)


src/bucket.js
```diff
@@ -2227,6 +2230,68 @@ Bucket.prototype.upload = function(localPath, options, callback) {
...
+  var contentType = mime.contentType(fileName);
+
+  if (contentType && !options.metadata.contentType) {
+    options.metadata.contentType = contentType;
+  }
+
+  // Determine protocal to use for get get request
+  var request = urlObject.protocol === 'https:' ? https : http;
+  try {
+    request.get(urlpath, function(response) {
```
stephenplusplus: Let's just use the `request` module.

stephenplusplus:
Thank you very much for the PR! What do you think about accepting a URL path right to `.upload()`? There are a couple ways  we could differentiate between a URL and a local path that I can think of;

- `isUrl = /^https*:/.test(filePath)`
- `fs.exists(filePath)`
- A module on npm that handles this for us?

Kiricon:
@stephenplusplus I like that idea a lot, initially I wasn't sure if it would be confusing or not but it would save a ton of repetition in the code base, docs and unit tests. If I close out this PR and open a new one with those changes changes mentioned would you be willing to take a look at that one?

(Honestly would be easier for me to just open a new PR, the first one didn't take a too long).

stephenplusplus:
Sure, whatever is easiest for you. Although, if you just push to this branch and keep this PR alive, we would have a smaller footprint if we ever need to look back and read how this feature was introduced and the discussions we had about it. That's a big "if", though, so whatever you prefer!

Kiricon:
@stephenplusplus Fair point you bring up, I'll see what I can do to preserver the audit trail for this feature.

Kiricon:
@stephenplusplus Implemented the changes we had talked about along with adding some docs. 

I hard reset the branch to remove all changes and rewrite the code but now it looks like the cirlceci server is now drying to pull old data from before the reset.

When you get the chance rebuilding without cache should fix the build errors.

lukesneeringer:
I just kicked off a new build and it failed again. Trying to ascertain why.

lukesneeringer:
Rebuilding without cache does not seem to be helping. I will open an issue with CircleCI. In the meantime, could we trouble you to just start another PR with a new branch name? Lousy for audit trail but we can get your change merged. :-D

lukesneeringer:
Tests pass. :-)
Leaving this in @stephenplusplus's capable hands as far as what else (if anything) needs to happen for this to be merged.

Kiricon:
@lukesneeringer Thank you very much for taking the time out of your day to help debug CircleCI. I really appreciate it!

src/bucket.js
```diff
@@ -2211,11 +2222,15 @@ Bucket.prototype.setUserProject = function(userProject) {
...
  * Example of uploading an encrypted file:
  */
-Bucket.prototype.upload = function(localPath, options, callback) {
+Bucket.prototype.upload = function(pathString, options, callback) {
   if (global.GCLOUD_SANDBOX_ENV) {
     return;
   }
 
+  var isURL =
+    !fs.existsSync(path.basename(pathString)) &&
```
stephenplusplus: Unfortunately for the simplicity of the code, let's stay away from the sync method and use `fs.exists` instead. I'm open to not using `fx.exists` at all, and instead just testing the beginning of the string for the http/https protocol. What do you think?

src/bucket.js
```diff
@@ -2238,22 +2253,30 @@ Bucket.prototype.upload = function(localPath, options, callback) {
     });
   } else {
     // Resort to using the name of the incoming file.
-    newFile = this.file(path.basename(localPath), {
```
stephenplusplus: Let's actually stick with `path.basename` for both cases. That way, it will turn "http://www.example.com/my/custom/file.jpg" into just "file.jpg" instead of "my/custom/file.jpg".

src/bucket.js
```diff
@@ -2263,11 +2286,28 @@ Bucket.prototype.upload = function(localPath, options, callback) {
 
       upload();
     });
+  } else if (isURL) {
+    uploadFromURL();
   }
 
   function upload() {
```
stephenplusplus: We can probably condense these methods:

```js
function upload() {
  var sourceStream;

  if (isUrl) {
    sourceStream = request.get(pathString);
  } else {
    sourceStream = fs.createReadStream(pathString);
  }

  sourceStream
    .on('error', callback)
    .pipe(newFile.createWriteStream(options))
    .on('error', callback)
    .on('finish', function() {
      callback(null, newFile);
    });
}
```

Kiricon:
@stephenplusplus Made the 3 changes you recommend and updated the unit tests to reflect them.

Not sure why the lint check on CircleCI is failing, looks like a configuration issue.

stephenplusplus:
@Kiricon I pushed a few changes to add some tests and add some resumable-detection when the path is a URL. Can you please give it a sanity check?

stephenplusplus:
Thank you @Kiricon!

PR #71, opened 2017-10-25
requester pays

### LINKED ISSUE #67: files.js:createReadStream doesn't include userProject when calling getMetadata
This is reposted from an internal issue. (Googlers: b/68164122)

---------------------------

In project A, create requester-pays bucket my-bucket and with IAM roles allUsers:legacyBucketOwner,objectAdmin 

In project B add user U to a role with serviceusage.services.use permission.

Use the requesterPays sample to read an object in my-bucket with credentials for user U.

  * Expected: download succeeds
  * Actual: download fails

Hacking the invocation of `getMetadata` to include the `userProject` worked for me.

### LINKED ISSUE #68: In files.js:copy, rewrite calls only includes userProject on the first request
This is re-posted from an internal ticket. (Googlers: b/68164387)

----------------------------------

This causes copy requests to fail.

Repro: Copy a large object across two requester pays buckets in different locations, while acting as a user in a different project. I hacked the source to set `query.maxBytesRewrittenPerCall = 1048576` so that I could test with a 2 MiB object.

callmehiphop:
Fixes #67
Fixes #68

- [x] Tests and linter pass

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -603,7 +603,7 @@ Bucket.prototype.createNotification = function(topic, options, callback) {
       method: 'POST',
       uri: '/notificationConfigs',
       json: snakeize(body),
-      qs: query
+      qs: query,
     },
     function(err, apiResponse) {
       if (err) {
@@ -1813,7 +1813,7 @@ Bucket.prototype.notification = function(id) {
  */
 Bucket.prototype.request = function(reqOpts, callback) {
   if (this.userProject && (!reqOpts.qs || !reqOpts.qs.userProject)) {
-    reqOpts.qs = extend(reqOpts.qs, { userProject: userProject });
+    reqOpts.qs = extend(reqOpts.qs, {userProject: this.userProject});
   }
 
   return common.ServiceObject.prototype.request.call(this, reqOpts, callback);
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -49,12 +49,16 @@ var RESUMABLE_THRESHOLD = 5000000;
  *
  * @param {Storage} storage A {@link Storage} instance.
  * @param {string} name The name of the bucket.
+ * @param {object} [options] Configuration object.
+ * @param {string} [options.userProject] User project.
  *
  * @example
  * var storage = require('@google-cloud/storage')();
  * var bucket = storage.bucket('albums');
  */
-function Bucket(storage, name) {
+function Bucket(storage, name, options) {
+  options = options || {};
+
   var methods = {
     /**
      * Create a bucket.
@@ -105,6 +109,13 @@ function Bucket(storage, name) {
    * @type {string}
    */
   this.storage = storage;
+
+  /**
+   * A user project to apply to each request from this bucket.
+   * @name Bucket#userProject
+   * @type {string}
+   */
+  this.userProject = options.userProject;
 
   /**
    * Cloud Storage uses access control lists (ACLs) to manage object and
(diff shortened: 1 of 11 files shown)


src/bucket.js
```diff
@@ -1971,6 +2006,22 @@ Bucket.prototype.setStorageClass = function(storageClass, options, callback) {
...
+ * @param {string} userProject The user project.
+ *
+ * @example
+ * var storage = require('@google-cloud/storage')();
+ * var bucket = storage.bucket('albums');
+ *
+ * bucket.setUserProject('grape-spaceship-123');
+ */
+Bucket.prototype.setUserProject = function(userProject) {
+  this.userProject = userProject;
```
stephenplusplus: @frankyn I might be crazy, but I thought I remember reading in an early Requester Pays design requirement doc that we shouldn't store a `userProject` on a Bucket instance. Instead, it has to be supplied each time a method is called. Am I making that up?

stephenplusplus: To elaborate, this would allow this kind of thing:

```js
var bucket = gcs.bucket('some-bucket', { userProject: 'my-project' })
bucket.getMetadata() // `userProject` passed to API automatically
```

As opposed to:

```js
var bucket = gcs.bucket('some-bucket')
bucket.getMetadata({ userProject: 'my-project' })
```

callmehiphop: Oh, that might be true, I never saw any of the design docs for this.

frankyn: @stephenplusplus I recall that a user needs to opt-in by specifying a user project. I don't recall seeing a design document.
However,  Ruby and PHP for example allow the user to opt into UserProject once when getting the bucket resource and follow-up operations using this bucket resource continue to use this value.

callmehiphop: @stephenplusplus based on that, do you think we should consider removing the query string parameter and only support the bucket option?

frankyn: @callmehiphop What did you mean by removing the query string parameter?

callmehiphop: Sorry, what I meant was that we should remove the option to specify a user project at the method level and only allow users to configure a user project at the bucket level.

frankyn: I think having both options is helpful when a service account has access to multiple project ids used for billing. Another reason is `UserProject` can be set for all Storage API requests and not only for a Requester Pays bucket.

system-test/storage.js
```diff
@@ -74,14 +95,14 @@ describe('storage', function() {
           return;
         }
 
-        async.eachLimit(buckets, 10, deleteBucket, done);
+        async.eachLimit(buckets, 10, deleteBucket, function() {
```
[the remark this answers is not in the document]
callmehiphop: This was intentional.. regardless of whether or not the buckets delete properly I still want to delete the topic. So I'm not really sure what to even do if an error occurred here.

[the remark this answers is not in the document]
stephenplusplus: New direction: we should change this from a `topic.delete()` to a `pubsub.getTopics({ filter: 'unique-prefix-' }).forEach(delete)`. Future test runs would catch any stale topics.

I see that as;

```js
async.parallel([
  deleteBuckets,
  deleteTopics
], done)
```

callmehiphop:
@stephenplusplus @lukesneeringer I'm a little confused as to why Codecov is reporting a decrease in coverage, when I try to view the diffs it says everything is 100%. Any ideas what is going on here??

system-test/storage.js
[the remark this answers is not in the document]
callmehiphop: PTAL

test/bucket.js
```diff
@@ -173,31 +173,43 @@ describe('Bucket', function() {
...
-        bind: function(context) {
-          return context;
-        },
-      };
+      after(function() {
+        Bucket.prototype.request = _request;
+      });
+
+      it('should create an ACL object', function() {
+        Bucket.prototype.request = {
```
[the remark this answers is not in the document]
callmehiphop: Done.

PR #79, opened 2017-11-03
docs: Add link to upstream extensionHeaders docs.

### LINKED ISSUE #76: extensionHeaders in method File.getSignedUrl do not work as expected
_From @teemeh on August 25, 2017 14:41_

It's only noted in the documentation that the "extensionHeaders" needs to be an object with not a single example or link. While not using such key-value pair in config object, I can easily retrieve the file, setting a self-defined header in way described below, which I believe is the most common way, doesn't allow me to retrieve the file anymore with the error message "The request signature we calculated does not match the signature you provided."

#### Environment details

  - OS: ubuntu
  - Node.js version: v6.11.2
  - npm version: irrelavent
  - google-cloud-node version: latest

#### Steps to reproduce

1.`var myHeaders = { myHeader: "myHeader"};`
2.Then inside the config object, we have the key-value pair `extensionHeaders: myHeaders,`.
The config object then is fed to File.getSignedUrl().

No matter how I try (node.js request module or curl), the uncommented key-value pair (step 2) will always cause error, whereas the error is gone as soon as I comment that one line out in step 2 (logically, it doesn't matter if I delete the customized request header when doing the request). Not sure if it's the server's problem or google-cloud-node.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2559_

stephenplusplus:
Fixes #76 

This adds a link to the upstream docs that explain `extensionHeaders`.

### DIFF
--- a/src/file.js
+++ b/src/file.js
@@ -1551,6 +1551,10 @@ File.prototype.getSignedPolicy = function(options, callback) {
  *     given is passed to `new Date()`.
  * @param {object} [config.extensionHeaders] If these headers are used, the
  *     server will check to make sure that the client provides matching values.
+ *     See [Canonical extension headers](https://cloud.google.com/storage/docs/access-control/signed-urls#about-canonical-extension-headers)
+ *     for the requirements of this feature, most notably:
+ *       - The header name must be prefixed with `x-goog-`
+ *       - The header name must be all lowercase
  * @param {string} [config.promptSaveAs] The filename to prompt the user to
  *     save the file as when the signed url is accessed. This is ignored if
  *     `config.responseDisposition` is set.


src/file.js
```diff
@@ -1551,6 +1551,10 @@ File.prototype.getSignedPolicy = function(options, callback) {
  *     given is passed to `new Date()`.
  * @param {object} [config.extensionHeaders] If these headers are used, the
  *     server will check to make sure that the client provides matching values.
+ *     See [Canonical extension headers](https://cloud.google.com/storage/docs/access-control/signed-urls#about-canonical-extension-headers)
+ *     for the requirements of this feature, most notably:
+ *       - The header name must be prefixed with `x-goog-`
```
stephenplusplus: @jmdobry does this indentation / list format work for the docs?

PR #81, opened 2017-11-03
Add file#rotateEncryptionKey().

### LINKED ISSUE #22: Storage - file.js. Any way to set Source Encryption Key?
_From @conundrum888 on May 21, 2017 14:26_

Hi,
In File.js of the Storage package, the function setEncryptionKey() helps set the "x-goog-encryption-key" header. Is there a function to set the "x-goog-copy-source-encryption-key header"? Can a function be added for this please?

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2322_

stephenplusplus:
Fixes #22 

This adds `file.rotateEncryptionKey(newKey, [callback])`.

### DIFF
--- a/samples/encryption.js
+++ b/samples/encryption.js
@@ -64,7 +64,7 @@ function uploadEncryptedFile(bucketName, srcFilename, destFilename, key) {
   // const destFilename = 'Remote destination for file, e.g. file_encrypted.txt';
 
   // See the "Generating your own encryption key" section above.
-  // const key = 'A customer-supplied key';
+  // const key = 'A base64 encoded customer-supplied key';
 
   const options = {
     // The path to which the file should be uploaded, e.g. "file_encrypted.txt"
@@ -106,7 +106,7 @@ function downloadEncryptedFile(bucketName, srcFilename, destFilename, key) {
   // const destFilename = 'Local destination for file, e.g. ./file.txt';
 
   // See the "Generating your own encryption key" section above.
-  // const key = 'A customer-supplied key';
+  // const key = 'A base64 encoded customer-supplied key';
 
   const options = {
     // The path to which the file should be downloaded, e.g. "./file.txt"
@@ -144,8 +144,8 @@ function rotateEncryptionKey(bucketName, fileName, oldKey, newKey) {
   // const fileName = 'Nome of a file in the bucket, e.g. my-file';
 
   // See the "Generating your own encryption key" section above.
-  // const oldKey = 'The current customer-supplied encryption key';
-  // const newKey = 'A new customer-supplied encryption key';
+  // const oldKey = 'The current base64 encoded customer-supplied encryption key';
+  // const newKey = 'A new base64 encoded customer-supplied encryption key';
 
   storage
     .bucket(bucketName)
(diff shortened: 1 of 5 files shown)


samples/encryption.js
```diff
@@ -122,13 +122,34 @@ function downloadEncryptedFile(bucketName, srcFilename, destFilename, key) {
...
+  const Storage = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage();
+
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // const fileName = 'Nome of a file in the bucket, e.g. my-file';
```
frankyn: I think oldKey and newKey should be apart of this const list of variables.

samples/encryption.js
```diff
@@ -122,13 +122,34 @@ function downloadEncryptedFile(bucketName, srcFilename, destFilename, key) {
...
+
+  storage
+    .bucket(bucketName)
+    .file(fileName, {
+      encryptionKey: Buffer.from(oldKey, 'base64'),
```
frankyn: As a user I would think the Buffer.from(oldKey, 'base64') would be hidden and I would only consider passing a base64 string without creating a Buffer object. 

Could you provide more clarity here?

samples/encryption.js
stephenplusplus: I don't believe so. I've been following the format of the previous examples in this file, such as this method: https://github.com/stephenplusplus/nodejs-storage/blob/18c92a222262708b0a6a9866eaed3eea8431798b/samples/encryption.js#L89

frankyn: I feel that this is a typo given other samples such as https://github.com/stephenplusplus/nodejs-storage/blob/18c92a222262708b0a6a9866eaed3eea8431798b/samples/files.js#L57 which define a commented variable for users determined by method parameters.

stephenplusplus: I see now, and I would have to agree!

stephenplusplus: Added in the `const` declarations 👍

samples/encryption.js
```diff
@@ -63,10 +63,14 @@ function uploadEncryptedFile(bucketName, srcFilename, destFilename, key) {
   // const srcFilename = 'Local file to upload, e.g. ./local/path/to/file.txt';
   // const destFilename = 'Remote destination for file, e.g. file_encrypted.txt';
 
+  // See the "Generating your own encryption key" section above.
+  // const key = 'A customer-supplied key';
```
frankyn: How about: A base64 encoded customer-supplied key? That way know what the sample code is expecting without reading the code.

stephenplusplus: The "Generating your own encryption key" is from the docs, which is where these samples are read from: https://cloud.google.com/storage/docs/using-encryption-keys#storage-generate-encryption-key-nodejs

I think it's a good idea to add more context by saying "A base64 encoded customer-supplied key", but I think it would help to leave the reference to the "Generating your own encryption key" section, which can elaborate.

samples/encryption.js
stephenplusplus: Our API was written to accept either a Buffer or 32 characters that we make the key from. It does not accept a base64 string, which is why the sample code has to do the `Buffer.from` conversion.

samples/encryption.js
stephenplusplus: I popped in some more "base64" references.

Issue #69, opened 2017-10-24
CONTENT_DOWNLOAD_MISMATCH on prematurely closed streams


rexxars:
#### Issue

When you use the `bucket.file(...).createReadStream()` and pipe it to a target target stream, and that target stream is ended/closed before the Google Cloud Storage client is finished reading the remote file, the validation fails because it tries to compare the partially received data with the remote hash.

I've created a repository containing a simplified example of the problem. In the real-world, we experienced this problem when an HTTP server behind a reverse proxy was piping from GCS to the HTTP response, and the upstream proxy closed the socket.

#### Environment details

  - OS: Ubuntu Linux 16.04
  - Node.js version: 8.7.0
  - npm version: 5.4.2
  - @google-cloud/storage version: 1.4.0

#### Steps to reproduce
  1. `git clone git@github.com:rexxars/gcs-premature-eos-error.git`
  2. `cd gcs-premature-eos-error`
  3. `npm install`
  4. Define the following environment variables (alternatively, modify `test.js` directly):
      - `GCS_KEY_PATH` - Path to a Google Cloud key file with access to a bucket
      - `GCS_BUCKET` - Name of Google Cloud Storage bucket to fetch from
      - `GCS_FILE_PATH` - Path to a file inside of the bucket
  5. `npm test`

After 60 seconds, the script will exit with the following error:
```
Error: The downloaded data did not match the data from the server. To be sure the content is the same, you should download the file again.
    at onComplete (.../node_modules/@google-cloud/storage/src/file.js:638:29)
    at .../node_modules/@google-cloud/storage/src/bucket.js:1439:7
    at Object.handleResp (.../node_modules/@google-cloud/common/src/util.js:135:3)
    at .../node_modules/@google-cloud/common/src/util.js:465:12
    at Request.onResponse [as _callback] (.../node_modules/retry-request/index.js:179:7)
    at Request.self.callback (.../node_modules/request/request.js:186:22)
    at emitTwo (events.js:125:13)
    at Request.emit (events.js:213:7)
    at Request.<anonymous> (.../node_modules/request/request.js:1163:10)
    at emitOne (events.js:115:13)
```

stephenplusplus:
Thanks for the reproduction case! I have received the error from the test, and will be looking into it.

stephenplusplus:
A quick question about:

[...]

PR #110, opened 2017-12-26
Attach end event listener to last stream in pipe.

### LINKED ISSUE #108: file.createReadStream & all dependencies return potentially broken data with gzip/validation stream on
#### Environment details

  - OS: any
  - Node.js version: any
  - npm version: 5
  - @google-cloud/storage version: >= 1.4.0

#### Steps to reproduce

https://github.com/googleapis/nodejs-storage/blob/3921796b26b15abccfba0556f05962fbb617870e/src/file.js#L584-L592

It's possible for 'end' of rawResponseStream to fire before all of the data has been processed by gzip / validation stream and end transformStream prematurely, effectively loosing data in the process.

'end' must only be called after it has been confirmed that further stream had emitted 'end' event

https://github.com/googleapis/nodejs-storage/blob/3921796b26b15abccfba0556f05962fbb617870e/src/file.js#L594-L609

so basically something like can return incomplete data:

`file.createReadStream('name-of-file.txt').pipe(transform).pipe(fs.createWriteStream())`

Essentially gunzip will take sometime, will be performed automatically, but 'end' will fire before all the data has been passed further to transform resulting in data corruption

stephenplusplus:
Fixes #108

### DIFF
--- a/src/file.js
+++ b/src/file.js
@@ -585,9 +585,7 @@ File.prototype.createReadStream = function(options) {
         rawResponseStream = rawResponseStream.pipe(pumpify.obj(throughStreams));
       }
 
-      rawResponseStream
-        .on('end', onComplete)
-        .pipe(throughStream, {end: false});
+      rawResponseStream.on('end', onComplete).pipe(throughStream, {end: false});
     }
 
     // This is hooked to the `complete` event from the request stream. This is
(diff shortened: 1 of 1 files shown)


SimenB:
Does this fix #69 as well? Just ran into this today (we're on 1.5.1, though)

stephenplusplus:
Hey @SimenB. In theory, this could resolve #69, but in that particular issue, I wasn't able to recreate the problem. This fix is in `1.5.2`, so feel free to give it a try. If there are still issues, please create a new issue with details of the environment and a reproduction case so we can dig into it further.

PR #133, opened 2018-02-08
add 'auto' options for contentType and gzip

calvinmetcalf:
does 3 related things for creating files

- lets contentType be defined directly in options instead of in options.metadata
- adds an 'auto' option for contentType that will infer it from the file name
- adds an 'auto' option gzip which uses content type to figure out if a file is compressible

for contentType at least 'auto' might be a sensible default but that's likely a breaking change.

Happy to break this into multiple pulls if part of this makes sense but other parts don't

- [x] Tests and linter pass


calvinmetcalf:
rebased

stephenplusplus:
I think this is all awesome, thank you!

It would be nice to default to auto, but like you said, we might have to defer that to our next major release. I'll make an issue to remind us about it when we're getting ready.

I made some small style changes and test tweaks. Feel free to take a look and let me know if you think I broke anything.

calvinmetcalf:
yeah those changes are fine, yeah making the contentType default to auto would put this tool more inline with gsutils (gzip should probably not default to auto)

PR #158, opened 2018-03-22
Add location and storageClass metadata to createBucket sample

frankyn:
- [x] Tests and linter pass


frankyn:
Thanks @stephenplusplus!

PR #162, opened 2018-03-28
Add `directory` option to `getFiles()`

### LINKED ISSUE #26: add a bucket#getDirectories() helper method
_From @esprehn on September 5, 2017 21:1_

#### Environment details

  - OS: OS X
  - Node.js version: v6.10.3
  - npm version: 3.10.10
  - google-cloud-node version: 1.2.1

#### Steps to reproduce

  1. require `google-cloud`

```js
bucket.getFiles({ delimiter: '/' }, function(err, files, nextQuery, api) {
  console.log(err, files, nextQuery, api);
});
```

Will print `null [] undefined undefined`.

This is very confusing, it seems you need to pass autoPaginate: false to get the api response back.

Looking at the code I don't know where this happens, the code does:

```js
    var nextQuery = null;
    if (resp.nextPageToken) {
      nextQuery = extend({}, query, {
        pageToken: resp.nextPageToken
      });
    }

    callback(null, files, nextQuery, resp);
```

And if you console.log(resp) here it's non-undefined,  but by the time it gets to your callback it's undefined. Where did the extra arguments end up?

Note also that while other apis return the ApiResponse in the second position of the array, getFiles does't so you can't do:

```js
const [files, api] = await bucket.getFiles({ delimiter: '/' });
console.log(api.prefixes);
```

and instead you must pass a callback.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2594_

stephenplusplus:
Fixes #26

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -1163,6 +1163,8 @@ Bucket.prototype.get = function(options, callback) {
  *     names, aside from the prefix, contain delimiter will have their name
  *     truncated after the delimiter, returned in `apiResponse.prefixes`.
  *     Duplicate prefixes are omitted.
+ * @property {string} [directory] Filter results based on a directory name, or
+ *     more technically, a "prefix".
  * @property {string} [prefix] Filter results to objects whose names begin
  *     with this prefix.
  * @property {number} [maxApiCalls] Maximum number of API calls to make.
@@ -1257,6 +1259,13 @@ Bucket.prototype.getFiles = function(query, callback) {
   if (!callback) {
     callback = query;
     query = {};
+  }
+
+  query = extend({}, query);
+
+  if (query.directory) {
+    query.prefix = `${query.directory}/`.replace(/\/*$/, '/');
+    delete query.directory;
   }
 
   this.request(
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -1963,30 +1963,37 @@ describe('storage', function() {
   });
 
   describe('list files', function() {
+    var DIRECTORY_NAME = 'directory-name';
+
     var NEW_FILES = [
       bucket.file('CloudLogo1'),
       bucket.file('CloudLogo2'),
       bucket.file('CloudLogo3'),
+      bucket.file(`${DIRECTORY_NAME}/CloudLogo4`),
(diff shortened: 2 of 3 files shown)


callmehiphop:
@stephenplusplus does this actually fix #26? This seems like a nice addition, but I thought the request was for a `getDirectories` method?

stephenplusplus:
The thread got a bit complicated-- I wasn't sure the best way to honor the request for "getting directories" meant "get directory names" or "get files from a specified directory".

I started this PR as just a new function that would return an array of directory name strings. After implementing, it felt strange, since our `get___` methods are supposed to return custom objects. And it brought up questions like: should it return nested directories, like "a/b/c" or just "a/"? I took it a step further and thought if they want to know the directories, they probably want the files from them-- should we return arrays of files?

I left a comment in the PR, hoping to get some of the original folks in the discussion to chime in. I'm not sure which way to go at this point.

callmehiphop:
@stephenplusplus got it. Should we let this hang out a while longer until some of the users get back to us?

stephenplusplus:
Yeah, that works. And if we don't hear back after too long, I suppose it's an inactive feature request that we can close.

The change in this PR could still be considered useful, in which case, it can be merged on its own merit.

stephenplusplus:
@callmehiphop @alexander-fenster do we want this?

fenixsolorzano:
It would be super helpful to get a list of file names, getFiles is downloading the whole file.

stephenplusplus:
@fenixsolorzano `getFiles()` shouldn't be downloading the contents of the file, it just gives you a `File` object, which you can use to perform operations on, such as downloading, copying, etc.

If you want just the names, this should work:

```js
const [files] = await storage.bucket('stephen-has-a-new-bucket').getFiles()
console.log(files.map(file => file.name))
```

fenixsolorzano:
Hmm, that's kinda weird, because I have ~1,600 images stored in a bucket and when I used `getFiles()` in a cloud function with 2GB of memory it ran out of memory

JustinBeckwith:
That sounds like a different problem 😂. Can you file a new bug with details on that one?

PR #163, opened 2018-03-28
Fix stale test.

stephenplusplus:
A 404 error for an object used to return an error message that said something like "Not found". It now returns "No such object". I removed our test that checked for specific messaging, and instead just confirm that the code is 404.


stephenplusplus:
Ping @alexander-fenster @callmehiphop

PR #168, opened 2018-04-03
don't require contentType for bucket.combine

### LINKED ISSUE #166: Don't throw an error if contentType can't be determined in bucket.combine
Currently an error is thrown if this library can't detect the content type from the filename with `bucket.combine(files, "destination")`. The GCS API doesn't require a content type though[1], so I don't think this library should throw.

https://github.com/googleapis/nodejs-storage/blob/31e4654497cb02ce8dd2111c64f0c7c739d0b19c/src/bucket.js#L338-L340

#### Environment details

  - `@google-cloud/storage` version: master

[1] As far as I can tell at least. I tried the API explorer on the [doc page](https://cloud.google.com/storage/docs/json_api/v1/objects/compose) and it didn't cause an error. GCS responded with an object without a content-type. From the other docs, it sounds like it serves the contents with `application/octet-stream` in this scenario.

zbjornson:
Fixes #166

- [x] Tests and linter pass

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -274,7 +274,6 @@ util.inherits(Bucket, common.ServiceObject);
  * @throws {Error} if a non-array is provided as sources argument.
  * @throws {Error} if less than two sources are provided.
  * @throws {Error} if no destination is provided.
- * @throws {Error} if content type can't be determined for the destination file.
  *
  * @param {string[]|File[]} sources The source files that will be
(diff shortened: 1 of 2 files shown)


stephenplusplus:
🙇 thank you! I marked as `do not merge` until we confirm in #166 this is what we want (I assume it is, just want to be sure).

PR #179, opened 2018-04-18
Introduce KMS

stephenplusplus:
This adds support for KMS.

#### To Dos

- [x] [Merge & Release support in gcs-resumable-upload](https://github.com/stephenplusplus/gcs-resumable-upload/pull/54)
- [x] System tests
- Unit tests
  - [x] bucket.js
  - [x] file.js

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -280,6 +280,11 @@ util.inherits(Bucket, common.ServiceObject);
  * @param {string|File} destination The file you would like the
  *     source files combined into.
  * @param {object} [options] Configuration options.
+ * @param {string} [options.kmsKeyName] Resource name of the Cloud KMS key, of
+ *     the form
+ *     `projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`,
+ *     that will be used to encrypt the object. Overwrites the object metadata's
+ *     `kms_key_name` value, if any.
  * @param {string} [options.userProject] The ID of the project which will be
  *     billed for the request.
  * @param {CombineCallback} [callback] Callback function.
@@ -1037,8 +1042,12 @@ Bucket.prototype.exists = function(options, callback) {
  * @param {object} [options] Configuration options.
  * @param {string|number} [options.generation] Only use a specific revision of
  *     this file.
- * @param {string} [options.key] A custom encryption key. See
+ * @param {string} [options.encryptionKey] A custom encryption key. See
  *     [Customer-supplied Encryption Keys](https://cloud.google.com/storage/docs/encryption#customer-supplied).
+ * @param {string} [options.kmsKeyName] The name of the Cloud KMS key that will
+ *     be used to encrypt the object. Must be in the format:
+ *     `projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`.
+ *     KMS key ring must use the same location as the bucket.
  * @returns {File}
  *
  * @example
@@ -1279,6 +1288,10 @@ Bucket.prototype.getFiles = function(query, callback) {
 
         if (query.versions) {
           options.generation = file.generation;
+        }
+
+        if (file.kmsKeyName) {
+          options.kmsKeyName = file.kmsKeyName;
         }
 
         const fileInstance = self.file(file.name, options);
@@ -1927,6 +1940,15 @@ Bucket.prototype.setLabels = function(labels, options, callback) {
  * }, function(err, apiResponse) {});
  *
  * //-
+ * // Enable KMS encryption for objects within this bucket.
+ * //-
+ * bucket.setMetadata({
+ *   encryption: {
+ *     defaultKmsKeyName: 'projects/grape-spaceship-123/...'
+ *   }
+ * }, function(err, apiResponse) {});
+ *
+ * //-
  * // If the callback is omitted, we'll return a Promise.
  * //-
  * bucket.setMetadata(metadata).then(function(data) {
@@ -2066,6 +2088,9 @@ Bucket.prototype.setUserProject = function(userProject) {
  *     [Customer-supplied Encryption Keys](https://cloud.google.com/storage/docs/encryption#customer-supplied).
  * @param {boolean} [options.gzip] Automatically gzip the file. This will set
  *     `options.metadata.contentEncoding` to `gzip`.
+ * @param {string} [options.kmsKeyName] The name of the Cloud KMS key that will
+ *     be used to encrypt the object. Must be in the format:
+ *     `projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`.
  * @param {object} [options.metadata] See an
  *     [Objects: insert request body](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON).
  * @param {string} [options.offset] The starting byte of the upload stream, for
@@ -2277,12 +2302,14 @@ Bucket.prototype.upload = function(pathString, options, callback) {
     // Use the string as the name of the file.
     newFile = this.file(options.destination, {
       encryptionKey: options.encryptionKey,
+      kmsKeyName: options.kmsKeyName,
     });
   } else {
     // Resort to using the name of the incoming file.
     const destination = path.basename(pathString);
     newFile = this.file(destination, {
       encryptionKey: options.encryptionKey,
+      kmsKeyName: options.kmsKeyName,
     });
   }
 
--- a/src/file.js
+++ b/src/file.js
@@ -205,7 +205,7 @@ util.inherits(File, common.ServiceObject);
  * @param {string} [options.destinationKmsKeyName] Resource name of the Cloud
  *     KMS key, of the form
  *     `projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`,
- *     that will be used to encrypt the object. Overrides the object metadata's
+ *     that will be used to encrypt the object. Overwrites the object metadata's
  *     `kms_key_name` value, if any.
  * @param {string} [options.keepAcl] Retain the ACL for the new file.
  * @param {string} [options.predefinedAcl] Set the ACL for the new file.
--- a/src/file.js
+++ b/src/file.js
@@ -85,6 +85,9 @@ const GS_URL_REGEXP = /^gs:\/\/([a-z0-9_.-]+)\/(.+)$/;
  * @param {object} [options] Configuration options.
  * @param {string} [options.encryptionKey] A custom encryption key.
  * @param {number} [options.generation] Generation to scope the file to.
+ * @param {string} [options.kmsKeyName] Cloud KMS Key used to encrypt this
+ *     object, if the object is encrypted by such a key. Limited availability;
+ *     usable only by enabled projects.
  * @param {string} [options.userProject] The ID of the project which will be
  *     billed for all requests made from File object.
  * @example
@@ -98,6 +101,8 @@ function File(bucket, name, options) {
 
   this.bucket = bucket;
   this.storage = bucket.parent;
+
+  this.kmsKeyName = options.kmsKeyName;
   this.userProject = options.userProject || bucket.userProject;
 
   Object.defineProperty(this, 'name', {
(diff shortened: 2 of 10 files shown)


src/bucket.js
```diff
@@ -1037,8 +1047,11 @@ Bucket.prototype.exists = function(options, callback) {
  * @param {object} [options] Configuration options.
  * @param {string|number} [options.generation] Only use a specific revision of
  *     this file.
- * @param {string} [options.key] A custom encryption key. See
+ * @param {string} [options.encryptionKey] A custom encryption key. See
```
stephenplusplus: (Note: This isn't a breaking change, it was just a docs goof)

src/file.js
```diff
@@ -2071,21 +2129,38 @@ File.prototype.request = function(reqOpts, callback) {
...
+ *     key name.
  * @returns {File}
  *
  * @example <caption>include:samples/encryption.js</caption>
  * region_tag:storage_rotate_encryption_key
  * Example of rotating the encryption key for this file:
  */
-File.prototype.rotateEncryptionKey = function(encryptionKey, callback) {
-  var newFile = this.bucket.file(this.id, {encryptionKey});
+File.prototype.rotateEncryptionKey = function(key, options, callback) {
```
stephenplusplus: This is not a great design here, so opinions welcome. The problem is that `rotateEncryptionKey` never expected that there would be two types of keys. To support a `kms` key, I created an `options` object, which has to specify `options.kms = true`.

stephenplusplus:
cc @frankyn

src/bucket.js
```diff
@@ -1281,6 +1289,10 @@ Bucket.prototype.getFiles = function(query, callback) {
           options.generation = file.generation;
         }
 
+        if (file.kmsKeyName) {
+          options.kmsKeyName = file.kmsKeyName;
```
callmehiphop: Should we be checking that we're not overwriting an already specified `kmsKeyName`?

stephenplusplus: Just a guess, but the diff in this PR makes it look like we have an option `[options.kmsKeyName] The name of the...` to this method, but that's actually for a separate method, `bucket.file()`.

But if that's not what maybe got you mixed up, there can be a default KMS key name on a Bucket, (set via `setMetadata`), but if a File has its own KMS key name, I believe we need to use that one.

LMK if I misunderstood!

system-test/storage.js
```diff
@@ -1778,6 +1778,290 @@ describe('storage', function() {
...
+            storage.request(
+              {
+                method: 'POST',
+                uri: `${keyRingsBaseUrl}/${keyRingId}/cryptoKeys/${cryptoKeyId}:setIamPolicy`,
+                json: {
+                  policy: {
+                    bindings: [
+                      {
+                        role: 'roles/cloudkms.cryptoKeyEncrypterDecrypter',
+                        members: `serviceAccount:${
```
frankyn: I recommend to use instead of writing the form `project-id@gs-project-accounts.iam.gserviceaccount.com` as this will change to `service-project-number@gs-project-accounts.iam.gserviceaccount.com`.
https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount/get

stephenplusplus: I'm not sure how to get the project number-- would the user running our tests have to explicitly give it to us, or is there an easier way? Auth for the tests comes from a service account JSON key file.

The API endpoint you gave only returns an email address, which uses the project ID:

```json
{
 "email_address": "nth-circlet-705@gs-project-accounts.iam.gserviceaccount.com",
 "kind": "storage#serviceAccount"
}
```

frankyn: The robot service account will be updated for GCS later this quarter and will be available using the serviceAccount.get request. My recommendation is to use the serviceAccount.get request to get the robot service account which will provide the correct format. You can ignore the new format for now.

stephenplusplus: Done! https://github.com/googleapis/nodejs-storage/pull/179/commits/a1dce4f5d6041338b26862ee0aaba0be2e4ce3a7

src/file.js
```diff
@@ -2071,22 +2120,30 @@ File.prototype.request = function(reqOpts, callback) {
...
+ * @param {boolean} [options.kmsKeyName] A Cloud KMS key name.
  * @returns {File}
  *
  * @example <caption>include:samples/encryption.js</caption>
  * region_tag:storage_rotate_encryption_key
  * Example of rotating the encryption key for this file:
  */
-File.prototype.rotateEncryptionKey = function(encryptionKey, callback) {
-  var newFile = this.bucket.file(this.id, {encryptionKey});
+File.prototype.rotateEncryptionKey = function(options, callback) {
```
frankyn: does this feel idiomatic to you @stephenplusplus, it was a suggestion.

stephenplusplus: Unfortunately, I think we're in a lose-lose situation. I think what you suggested is better than having to provide `{ kms: true }`. Having to do two things increases the odds of forgetting.

Have other languages found a better pattern when they implemented this?

frankyn: @ace-n or @jmdobry, could you help provide feedback on the surface change? I'm not a Node.js community expert and want to make sure.

@stephenplusplus, Ruby had a similar helper method, but added additional parameters for [kms key](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2059/files#diff-8c864fd34aa2f2f960b3b13353152551R923). However, I don't think this is how the Node.js client is meant to be used given most functions have an `options` parameter for this kind of metadata.

test/file.js
```diff
@@ -2799,15 +2956,44 @@ describe('File', function() {
...
+
+    it('should default to customer-supplied encryption key', function(done) {
+      var encryptionKey = 'encryption-key';
+
+      file.bucket.file = function(id, options) {
+        assert.strictEqual(options.encryptionKey, encryptionKey);
+        done();
+      };
+
+      file.rotateEncryptionKey(encryptionKey, assert.ifError);
```
frankyn: If the design is kept for rotateEncryptionKey(), there should exist a test that shows how to use a CSEK with a `{}` along with a string option.

src/file.js
```diff
@@ -2071,22 +2120,30 @@ File.prototype.request = function(reqOpts, callback) {
...
  *
- * @param {string|buffer} encryptionKey An AES-256 encryption key.
+ * @param {string|buffer|object} options An AES-256 encryption key. If you'd
+ *     like to use a Cloud KMS key name, you must specify an options object with
+ *     the property name: `kmsKeyName`.
+ * @param {string|buffer} [options.encryptionKey] An AES-256 encryption key.
+ * @param {boolean} [options.kmsKeyName] A Cloud KMS key name.
  * @returns {File}
  *
  * @example <caption>include:samples/encryption.js</caption>
```
frankyn: Sample should be updated to use `options.encryptionKey` if this design is kept.

src/file.js
```diff
@@ -2071,22 +2120,30 @@ File.prototype.request = function(reqOpts, callback) {
...
  * file.
  *
  * @see [Customer-supplied Encryption Keys]{@link https://cloud.google.com/storage/docs/encryption#customer-supplied}
  *
- * @param {string|buffer} encryptionKey An AES-256 encryption key.
+ * @param {string|buffer|object} options An AES-256 encryption key. If you'd
+ *     like to use a Cloud KMS key name, you must specify an options object with
+ *     the property name: `kmsKeyName`.
+ * @param {string|buffer} [options.encryptionKey] An AES-256 encryption key.
+ * @param {boolean} [options.kmsKeyName] A Cloud KMS key name.
```
frankyn: I don't believe this is a boolean if the design is kept.

src/file.js
```diff
@@ -2071,22 +2120,30 @@ File.prototype.request = function(reqOpts, callback) {
...
 
 /**
- * The Storage API allows you to use a custom key for server-side encryption.
  * This method allows you to update the encryption key associated with this
  * file.
  *
  * @see [Customer-supplied Encryption Keys]{@link https://cloud.google.com/storage/docs/encryption#customer-supplied}
  *
- * @param {string|buffer} encryptionKey An AES-256 encryption key.
+ * @param {string|buffer|object} options An AES-256 encryption key. If you'd
```
frankyn: If design is kept, should clarify that if a `string` is provided, an AES-256 encryption is expected. As it's not clear from first reading the documentation.

src/file.js
```diff
@@ -745,26 +776,42 @@ File.prototype.createReadStream = function(options) {
...
-  );
+  this.storage.authClient.getProjectId(function(err, projectId) {
+    // Ignore error.
+
+    if (self.kmsKeyName) {
+      self.kmsKeyName = common.util.replaceProjectIdToken(
+        self.kmsKeyName,
+        projectId
+      );
+    }
```
frankyn: Why is this kmsKeyName doing a replacement for projectId?

src/bucket.js
frankyn: Want to make sure this is resolved.

test/file.js
stephenplusplus: That's actually covered in this test:

```js
it('should create new File correctly', function(done) {
  var options = {};

  file.bucket.file = function(id, options_) {
    assert.strictEqual(id, file.id);
    assert.strictEqual(options_, options);
    done();
  };

  file.rotateEncryptionKey(options, assert.ifError);
});
```

That tests that *any* object goes through, which includes one with `encryptionKey` or any arbitrary property name.

Then we have "should default to customer-supplied encryption key" for strings, and "should accept a Buffer for customer-supplied encryption key" for Buffers

src/file.js
stephenplusplus: The format of KMS key name that we're anticipating is:

```
`projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`
```

Our library doesn't typically require the user to specify their project ID, which is why I was doing the replacement. However, if we're always requiring this full resource path, then it would be highly unlikely that the user wouldn't be providing their own project ID. That was an oversight on my part -- removed :)

src/bucket.js
stephenplusplus: Yes, I think this was just confusion on how the PR diff was displaying the changes. We use the object's KMS key name that comes back from the API.

test/file.js
stephenplusplus: Just for context, here's the full source code of the method we're testing:

```js
File.prototype.rotateEncryptionKey = function(options, callback) {
  if (!is.object(options)) {
    options = {
      encryptionKey: options,
    };
  }

  var newFile = this.bucket.file(this.id, options);
  this.copy(newFile, callback);
};
```

Since there's no specific handling when `options` is an object, we don't have any special tests other than that the original user-provided object makes its way to the `this.bucket.file` constructor.

We have system tests that cover that rotating works with CSEK and KMS.

stephenplusplus:
@frankyn thanks for pointing out those bloopers. I believe all is taken care of now.

test/file.js
stephenplusplus: We don't explicitly specify `{ encryptionKey: '...' }` in a unit test, because there is no line of source code that would care if we put "encryptionKey" or "kmsKeyName" or "madeUpKey", etc. Our testing strategy goes line-by-line to test that the user input goes to the right place.

In this case, the source code takes the user-provided options object and passes it directly to `this.bucket.file(this.id, options)`. So our test confirms that `this.bucket.file` receives the exact user-provided object. It's the the job of `this.bucket.file` to confirm the right thing happens next, which we have a separate collection of tests for.

To test that first block of code in the function:

```js
if (!is.object(options)) {
  options = {
    encryptionKey: options,
  };
}
```

We have these two tests:

```js
it('should default to customer-supplied encryption key', function(done) {
  var encryptionKey = 'encryption-key';

  file.bucket.file = function(id, options) {
    assert.strictEqual(options.encryptionKey, encryptionKey);
    done();
  };

  file.rotateEncryptionKey(encryptionKey, assert.ifError);
});

it('should accept a Buffer for customer-supplied encryption key', function(done) {
  var encryptionKey = crypto.randomBytes(32);

  file.bucket.file = function(id, options) {
    assert.strictEqual(options.encryptionKey, encryptionKey);
    done();
  };

  file.rotateEncryptionKey(encryptionKey, assert.ifError);
});
```

Those confirm that what we list in our JSDocs params work as advertised. The first gives it a string, then confirms it is interpreted as an `encryptionKey`. The second gives a buffer, and confirms the same.

frankyn: Ah okay, so the test verifies that any `{}` is received correctly in a request. I thought there was a test for `{ encryptionKey: '...' }` that I wasn't seeing. Thanks!

frankyn:
LGTM, please wait to merge PR. A change is coming that will remove global KMS keys and most likely will break your tests.

stephenplusplus:
I didn't realize this would be a public feature at the time of launch. Are there any public docs we can link to?

frankyn:
Not yet, but when it happens docs will be available. Update previous comment: A change is coming that will remove global KMS keys and tests will need to use regional KMS keys, that match the bucket location.

stephenplusplus:
@frankyn I've made the location update.

stephenplusplus:
Anyone?

stephenplusplus:
@crwilcox thanks for catching that! Fixed 👍

stephenplusplus:
@alexander-fenster do you know what might be going on here? https://circleci.com/gh/googleapis/nodejs-storage/4586 

Nightlies are succeeding.

alexander-fenster:
Regenerating `package-lock.json` should help. Just remove it, reinstall modules, and commit the changed lock file. If `lint` task fails after that, do the same for samples as well (but it might be not needed).

stephenplusplus:
Errors persist: https://circleci.com/gh/googleapis/nodejs-storage/4607

alexander-fenster:
Interesting. Looking...

alexander-fenster:
OK so what I did was, as usual, removing `node_modules` and `package-lock.json` in both main package and samples and reinstalling everything. CI all green now.

stephenplusplus:
Thanks! Somehow, the generated lock files were different for me, I guess.

stephenplusplus:
P.S. can you approve this while you're here?

PR #169, opened 2018-04-03
properly encode signed URLs

zbjornson:
Fixes #167

- [x] Tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

This uses the native `url` module to build the URL, which I think is less brittle than the existing string-building method.

PR #187, opened 2018-05-02
0.17.0

stephenplusplus:
# Do not merge
I think we're broken.

## Features

- (#165): Check if `$HOME` is writable before starting a resumable upload.
- (#174): Introduce a `directory` option to `getFiles()`.

## Fixes

- (#160): Return `apiResponse` argument from `bucket.upload()`. (Thanks for pointing this out, @scragg0x!)
- (#168): Don't require a content type for `bucket.combine()`. (Thanks, @zbjornson!) 
- (#169): Properly encode signed URLs. (Thanks again to @zbjornson!)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/storage",
-  "version": "1.6.0",
+  "version": "1.7.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Closing due to problems: https://github.com/googleapis/nodejs-storage/pull/188

PR #190, opened 2018-05-02
Fix system tests.

stephenplusplus:
## Do not merge: https://github.com/googleapis/nodejs-storage/pull/190#issuecomment-386419909

CI has been failing: https://circleci.com/gh/googleapis/nodejs-storage/4098.

### DIFF
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -1229,6 +1229,20 @@ describe('storage', function() {
         );
 
         it(
+          'file#createResumableUpload',
+          doubleTest(function(options, done) {
+            file.createResumableUpload(options, function(err, uri) {
+              assert.ifError(err);
+
+              file.createWriteStream({uri})
+                .on('error', done)
+                .on('finish', done)
+                .end('Test data');
+            });
+          })
+        );
+
+        it(
           'file#download',
           doubleTest(function(options, done) {
             file.download(options, done);
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -1393,6 +1393,16 @@ describe('storage', function() {
           })
         );
 
+        // @TODO: There may be a backend bug here.
+        // Reference: https://github.com/googleapis/nodejs-storage/pull/190#issuecomment-388475406
+        it.skip(
+          'iam#testPermissions',
+          doubleTest(function(options, done) {
+            const tests = ['storage.buckets.delete'];
+            bucketNonWhitelist.iam.testPermissions(tests, options, done);
+          })
+        );
+
         it(
           'notification#get',
           doubleTest(function(options, done) {
--- a/system-test/storage.js
+++ b/system-test/storage.js
@@ -1229,13 +1229,6 @@ describe('storage', function() {
         );
 
         it(
-          'file#createResumableUpload',
-          doubleTest(function(options, done) {
-            file.createResumableUpload(options, done);
-          })
-        );
-
-        it(
           'file#download',
           doubleTest(function(options, done) {
             file.download(options, done);
@@ -1282,7 +1275,7 @@ describe('storage', function() {
               }
 
               // Re-create the file. The tests need it.
-              file.save('newcontent', done);
+              file.save('newcontent', options, done);
             });
           })
         );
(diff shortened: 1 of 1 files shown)


system-test/storage.js
```diff
@@ -1282,7 +1275,7 @@ describe('storage', function() {
               }
 
               // Re-create the file. The tests need it.
-              file.save('newcontent', done);
+              file.save('newcontent', options, done);
```
stephenplusplus: This fixed 3/5 failures.

system-test/storage.js
```diff
@@ -1228,13 +1228,6 @@ describe('storage', function() {
           })
         );
 
-        it(
```
stephenplusplus: I removed this because the URL can still be created, even without providing a userProject on a requester pays bucket. There was either an API change that now allows this, or something else was busted on our end.

system-test/storage.js
```diff
@@ -1385,14 +1378,6 @@ describe('storage', function() {
...
```
stephenplusplus: I removed this because testing permissions is still allowed, even without providing a userProject on a requester pays bucket. There was either an API change that now allows this, or something else was busted on our end.

stephenplusplus:
@frankyn Our system tests around requester pays started failing. I left comments on the diff-- could you take a look? Thanks!

frankyn:
acking* I'm at a conference this week and will have delayed responses.

stephenplusplus:
@frankyn any availability to give this a quick look?

frankyn:
Hey @stephenplusplus, taking a look now.

system-test/storage.js
frankyn: When the URL is created, are you still able to upload?

stephenplusplus: Yes. We test this through `file.save()` flow, which performs a resumable upload.

frankyn:
Pinged GCS team for more context, resumable upload should be required to define user project unless I misunderstood. The testIamPermission is not as clear. I'll ping when I have an update.

system-test/storage.js
stephenplusplus: Sorry, I reversed the case we were talking about. The resumable upload does fail when a userProject isn't provided. I added a more thorough test to cover it now.

stephenplusplus:
Thanks @frankyn!

system-test/storage.js
frankyn: So this is no longer an issue?

frankyn: thanks for clarifying. So, It isn't an issue until you attempt to the upload.

stephenplusplus: Right-- after you get the URI, everything works as you would expect (fails without a userProject, and succeeds with).

lukesneeringer:
@googlebot The reason for overriding the CLA status is because you are wrong and should feel bad.

system-test/storage.js
frankyn: Thanks, waiting on more clarity on the case here. This is mildly confusing, but sounds as if it's WAI.

JustinBeckwith:
*gentle ping* what's the status on this one folks?

system-test/storage.js
stephenplusplus: I'm not sure if you meant clarity and confusion from my explanation, so just in case:

Our resumable uploads work in two phases: 1) Create a resumable URI, 2) Send data to it. When we first implemented requester pays support, step 1 would fail if the bucket was a requester pays bucket and no `userProject` was sent in the URL parameters. Somewhere along the way, our system tests started to fail, because step 1 began succeeding under the same conditions.

Ultimately, the way it works now seems correct-- a user can create a resumable URI without deciding what project will be billed for it, but is required to decide by the time data is actually uploaded. Nonetheless, since it's different than the way it was, I want to make sure :)

stephenplusplus:
Thanks for your help, @frankyn!

I've left the test in, but our test runner will skip it. There's also a `@TODO` with more information about why it's skipped.

Once tests pass, I'll merge this in.

Issue #222, opened 2018-06-05
Setting metadata seems really difficult to reason with.


gauntface:
I was just trying to set the metadata to files and frankly couldn't figure out what was going on. I'm still not 100% but I think this invalid:
```
file.setMetadata({
  'example-key': 'example-value',
});
```

Instead it's meant to be:

```
file.setMetadata({
  metadata: {
    'example-key': 'example-value',
  }
});
```

Reason being that custom metadata *has* to moved out of the top level set of metadata values.

If this is the case, I would love two changes made to the API to make this easy to reason with:

1. Enforce a strict check on the metadata value such that custom values *cannot* be used on the top level metadata object.
1. *Bit of a stretch but I would like it* Add a setCustomMetadata() method to avoid this weirdness around wrapping metadata with a metadata key.

gauntface:
Another point of confusion I hit was setting custom metadata when saving a file.

This silently drops the metadata (same as above):
```
file.save("hello", {
    contentType: 'application/json',
    metadata: {
        'example-key': 'example-value',
    }
})
```

This works:

```
file.save("hello", {
    contentType: 'application/json',
    metadata: {
        metadata: {
          'example-key': 'example-value',
        }
    }
})
```

jeffposnick:
Chiming in here—irrespective of *why* there is that double-`metadata` key, there's a lack of examples that illustrate how to use `metadata.metadata` and `gzip: true` to set values that are useful for CDN-type hosting setups.

I filed https://github.com/googleapis/nodejs-storage/pull/225 to update one of the examples to include those fields. Feel free to close that and make changes elsewhere if you think there's a better place to illustrate it.

stephenplusplus:
> Enforce a strict check on the metadata value such that custom values cannot be used on the top level metadata object.

I don't think we can go this route, as we'd have to whitelist property names, which could evolve without our knowing.

> Bit of a stretch but I would like it Add a setCustomMetadata() method to avoid this weirdness around wrapping metadata with a metadata key.

I'm neutral on this. I don't mind it, but I don't know if it will make anything more clear for users. The confusion is mostly that two different metadata concepts exist. Once they know that, it's not too hard to communicate the right type to us. But I agree it's weird, so if we want to just solve the weirdness, this is okay.

gauntface:
@stephenplusplus yeah, it kind of feels like "file options" and "file metadata" are being merged into one place but have different meaning / uses.

stephenplusplus:
[...]

PR #225, opened 2018-06-11
gzip and Cache-Control headers in upload sample

jeffposnick:
This is related to https://github.com/googleapis/nodejs-storage/issues/222

Setting `gzip: true` seems like a a very common thing to want to do if you're using Cloud Storage to handle HTTP traffic. Ditto for wanting to change the `Cache-Control:` response headers—they should either be 1 year if the content will never change, or 0 if the content will change. (See https://jakearchibald.com/2016/caching-best-practices/)

- [ ] Tests and linter pass


jeffposnick:
(I believe the [linting failure](https://circleci.com/gh/googleapis/nodejs-storage/5167) is due to something that I fixed in the [last commit](https://github.com/googleapis/nodejs-storage/pull/225/commits/c1a1f0ae8a037dc5ffafca71d77dd18bad21c483), which did not trigger another CI run.)

jeffposnick:
As it turns out, my sample was incorrect—pointing to the fact that this is really hard to get right without a complete example.

After some more digging, I found https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1087#issuecomment-175105508, and I'll adjust the example to match that.

PR #242, opened 2018-06-20
Major semver: drop node 4 support

jkwlui:
Blocking #241 (refactor es6 class)

- [ ] Tests and linter pass


kinwa91:
Something is still going wrong with my package-lock.json's (lint breaks), need some help please @googleapis/yoshi-nodejs

JustinBeckwith:
The answer here is comical.  You need to:

```
rm -rf package-lock.json node_modules samples/package-lock.json samples/node_modules
npm install
npm run prettier
```

That should fix it :)

JustinBeckwith (approved):
Assuming you get CI passing :)

ofrobots (approved):
nit: Use https://conventionalcommits.org/ for commit messages.

LGTM once CI is green.

kinwa91:
8252c7e fixed it. Still no idea why.

JustinBeckwith:
I don't think you meant to delete `samples/package-lock.json` :) . Actually, I just gave you bad instructions.  Please run `npm install` in the `samples` directory again to regenerate the file. 

On the common upgrade thing...  don't look a gift horse in the mouth? :)

.circleci/config.yml
```diff
@@ -139,7 +121,7 @@ jobs:
             NPM_CONFIG_PREFIX: /home/node/.npm-global
   docs:
     docker:
-      - image: 'node:8'
+      - image: 'node:10'
```
alexander-fenster: why is that? 8 is still LTS, 10 is current. I guess we test on LTS.

alexander-fenster: (same comment for all `node:10` below)

package.json
```diff
@@ -65,7 +65,7 @@
     "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json"
   },
   "dependencies": {
-    "@google-cloud/common": "^0.19.2",
+    "@google-cloud/common": "^0.20.0",
```
alexander-fenster: This is unrelated to dropping `node4` support and should go to a separate PR.

alexander-fenster:
Here goes the Right Way of regenerating lock files :)

```sh
npm install && npm link && cd samples && npm link ../ && npm install && cd .. && npm run lint
git status # sanity check: one or two lock files must be updated
git commit -a -m 'updated package lock files'
```

The `npm link` in main folder and `npm link ../` in samples folder are important to make it build the right dependency tree.

alexander-fenster:
(note: I have updated the commands in my previous comment)

package.json
alexander-fenster: And if you don't make this change here, you can safely revert your lock files as lock files are supposed to be changed only if package dependencies are changed :)

kinwa91:
@alexander-fenster it was one of the things that I tried as well. I even went as far as removing both `package-lock.json` and SSH into the (passing) build and grabbing the files and recommitting them.

kinwa91:
I'm suspecting it's a problem with my setup. Could you see if you would generate a different lock file?

kinwa91:
```shell
➜  ~ node -v
v10.4.1
➜  ~ npm -v
6.1.0
```

PR #248, opened 2018-06-21
Updated generateSignedUrl example to use a unix timestamp

iamtmrobinson:
The existing generateSignedUrl example results in the following error when trying to view the resource as the config object should contain a unix timestamp.
```
<Error>
  <Code>SignatureDoesNotMatch</Code>
  <Message>
    The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.
  </Message>
</Error>
```

- [x] Tests and linter pass

### DIFF
--- a/samples/files.js
+++ b/samples/files.js
@@ -330,7 +330,7 @@ function generateSignedUrl(bucketName, filename) {
   // These options will allow temporary read access to the file
   const options = {
     action: 'read',
-    expires: '03-17-2025',
+    expires: new Date("March 17 2025").getTime(),
   };
 
   // Get a signed URL for the file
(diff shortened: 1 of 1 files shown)


stephenplusplus:
Thanks for this, however, the source code already handles the string input correctly. The code behind the method:

```js
// input: "03-17-2025"

const expires = new Date(config.expires);
// new Date('03-17-2025')

const expiresInSeconds = Math.round(expires / 1000); // The API expects seconds.
// 1742184000
```

From the changes in this PR:

```js
// input: new Date('March 17 2025').getTime()

const expires = new Date(config.expires);
// new Date(new Date('March 17 2025').getTime());

const expiresInSeconds = Math.round(expires / 1000);
// 1742184000
```

iamtmrobinson:
Hmm, that's strange because it's the only thing I changed for my implementation to work 🤔 But your example given certainly works!

PR #251, opened 2018-06-22
system-tests: re-enable iam#testPermissions requesterPays test.

stephenplusplus:
re: #190 
cc: @frankyn 

## Do not merge
Re-enabling for debugging.

This re-enables the failing system test.


frankyn:
Thanks @stephenplusplus! Looks like they pass now. I recalled that you experienced an issue with resumable uploads not throwing the similar error. Was this resolved before this issue? As only one test was skipped for testPermissions?

stephenplusplus:
Sorry for the confusion here. The test that fails is a system test, which doesn't run on PRs. So to run it, you have to clone, `npm install`, set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to a service account key file, then `npm run system-test`.

The old behavior for `iam.testPermissions()` was:

- Did the user provide a `userProject`? Succeed.
- Did the user not provide a `userProject`? Fail.

The new behavior that caused us to skip the test still exists, which is:

- Did the user provide a `userProject`? Succeed.
- Did the user not provide a `userProject`? Succeed.

> I recalled that you experienced an issue with resumable uploads not throwing the similar error.

The test we used to have that failed was simply for creating a resumable upload URI. It didn't test actually uploading. Instead of skipping the test, I refactored it to perform an upload from the created URI.

The API call to create the URI still experiences the same behavior described above for `testPermissions`. It will succeed regardless of if a `userProject` is provided or not.

frankyn:
Thanks @stephenplusplus, I'm assuming I should need to set two different projects, one that owns the requester pays project and the service account defined in `GOOGLE_APPLICATION_CREDENTIALS` while the other project is used for userProject. So I should set `GCN_STORAGE_2ND_PROJECT_ID` and `GCN_STORAGE_2ND_PROJECT_KEY` as well?

stephenplusplus:
Yes, sorry, you're right. You need to set both `GCN_STORAGE_2ND_PROJECT_ID` and `GCN_STORAGE_2ND_PROJECT_KEY`.

JustinBeckwith:
👋 Just checking in - what's the status on this?  Do we still need it?

frankyn:
System tests still fail for `testIamPermission`. We can close this for now until the service is fixed.

bug: 79524739

PR #255, opened 2018-06-25
Typescript conversion (initial)

jkwlui:
Takes care of initial chores of converting to typescript. `tsc` compilation currently fails - they will be addressed in future PRs, organized by category.

- [ ] Tests and linter pass

 Changes:

 0e0b110 (Jonathan Lui, 12 minutes ago)
    refactor(ts): temporarily disable noImplicitAny before augmenting types

 219ef2f (Jonathan Lui, 3 days ago)
    refactor(ts): module.exports => export { ... }

 c185356 (Jonathan Lui, 3 days ago)
    refactor(ts): use native `extends Error` for custom error classes instead of relying on package 'create-error-class'

 a4dd24a (Jonathan Lui, 3 days ago)
    refactor(ts): for...in => for...of Object.keys(...

 986ef7d (Jonathan Lui, 3 days ago)
    refactor(ts): function to arrow =>

 90ef7e9 (Jonathan Lui, 3 days ago)
    refactor(ts): gts fix

 371604e (Jonathan Lui, 3 days ago)
    refactor(ts): remove .js extension from imports

 5bbfd24 (Jonathan Lui, 3 days ago)
    refactor(ts): install available @types packages

 d79f184 (Jonathan Lui, 3 days ago)
    refactor(ts): setup typescript tools

 d09350d (Jonathan Lui, 4 days ago)
    refactor(ts): Rename *.js to .ts


package.json
```diff
@@ -108,6 +128,7 @@
     "proxyquire": "^2.0.0",
     "semistandard": "^12.0.0",
     "tmp": "^0.0.33",
+    "typescript": "~2.8.0",
```
JustinBeckwith: lets use 2.9.2

src/acl.ts
```diff
@@ -757,6 +757,7 @@ common.util.promisifyAll(Acl, {
   exclude: ['request'],
 });
 
-module.exports = Acl;
-
-module.exports.AclRoleAccessorMethods = AclRoleAccessorMethods;
+export {
```
JustinBeckwith: I would shy away from the changes to the exports in the initial PR.  Lets come back around and get these later.

JustinBeckwith (commented):
Looking good so far!

.clang-format
```diff
@@ -0,0 +1,3 @@
+Language:        JavaScript
```
JustinBeckwith: @ofrobots do we auto-generate these now?  Can we delete them, or should they stay here?

package.json
```diff
@@ -62,7 +62,14 @@
...
     "all-test": "npm test && npm run system-test && npm run samples-test",
-    "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json"
+    "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json",
+    "check": "gts check",
+    "clean": "gts clean",
+    "compile": "tsc -p .",
+    "fix": "gts fix",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "posttest": "npm run check"
```
JustinBeckwith: There is a very low chance this will work.  I would get rid of this line, and open an issue (if I haven't already) to turn `gts check` back on.

crwilcox:
If compilation fails should we be checking this into master?

.clang-format
ofrobots: We do autogenerate these. There is no harm if you delete them, as we fall back to the default, but I would suggest keeping them. The benefit is that with these files present the editors know what formatting they should use as you edit.

kinwa91:
Currently I have this PR merge to `typescript` branch, where more PRs to addresses the compilation errors will go into.

PR #260, opened 2018-06-25
Typescript: internalize class and object members into class

jkwlui:
Mostly just moving `this.x = ...` to the class header.

- [x] Tests and linter pass


JustinBeckwith (approved):
Assuming CI passes :)

PR #266, opened 2018-06-27
fix: remove trailing slashes from bucket name.

### LINKED ISSUE #170: Error message in Node.js console is not the orignial error message
#### Environment details

  - OS: Windows 10
  - Node.js version: v8.9.4
  - npm version: 5.8.0
  - `@google-cloud/storage` version:

#### Steps to reproduce

  1. Not quite sure since I seem to have done something wrong. Anyway this is more general. I am using Formidable to try to upload an audio file to a gcs bucket. For some reason @google-cloud tries to parse this as JSON. That is probably my fault and not the issue here.

The issue is the error message in the console from the local Firebase web server. I get the error

"ApiError: Error during request."

This is not very helpful, of course :-(
Fortunately you can get a trace with "console.log". The error is thrown in 

functions\node_modules\@google-cloud\storage\node_modules\@google-cloud\common\src\util.js:193:32

and this line looks like this:

      parsedHttpRespBody.err = new util.ApiError('Cannot parse JSON response');

I think it would be much easier for a developer if this error was made visible directly (instead of " ApiError: Error during request").

stephenplusplus:
Fixes #170 

This cleans up the fragments of a gs:// URL, should the user provide one to the Bucket constructor.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -302,7 +302,8 @@ class Bucket extends ServiceObject {
   constructor(storage, name, options?) {
     options = options || {};
 
-    name = name.replace(/^gs:\/\//, '');
+    // Allow for "gs://"-style input, and strip any trailing slashes.
+    name = name.replace(/^gs:\/\//, '').replace(/\/+$/, '');
 
     const methods = {
       /**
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
@stephenplusplus @callmehiphop what's the latest on this one?

stephenplusplus:
I've modernized the PR, and it should be good for a review.

PR #267, opened 2018-06-27
Typescript: use ts-style exports instead of module.exports

jkwlui:


src/acl.ts
```diff
@@ -760,6 +760,7 @@ common.util.promisifyAll(Acl, {
   exclude: ['request'],
 });
 
-module.exports = Acl;
-
-module.exports.AclRoleAccessorMethods = AclRoleAccessorMethods;
+export {
+  Acl,
+  AclRoleAccessorMethods,
```
kinwa91: `AclRoleAccessorMethods` is only used in tests, will have to fix the import later.

JustinBeckwith (approved):
LGTM after CI passes

PR #268, opened 2018-06-28
Typescript: misc. fixes to make tsc happy

kinwa91:
`tsc` compiles successfully following this PR.

This PR addresses the following (and more):
 - optional callback?
 - type cast to `any` if incompatible types can't easily be fixed
 - fixes imports/exports
 - comparing `Date`'s using `.valueOf()`

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -27,7 +27,6 @@ import path from 'path';
 import snakeize from 'snakeize';
 import request from 'request';
 
-import Storage from './index';
 import {Acl} from './acl';
 import {File} from './file';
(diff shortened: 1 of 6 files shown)


src/bucket.ts
```diff
@@ -71,7 +70,7 @@ class Bucket extends common.ServiceObject {
    * @name Bucket#storage
    * @type {string}
    */
-  storage: Storage;
+  storage: any;
```
kinwa91: Can't import `Storage` as a type since we use `Proxy` hack

PR #289, opened 2018-07-11
build: test-no-cover should compile ts before running test

kinwa91:


package.json
```diff
@@ -57,7 +57,7 @@
     "docs": "jsdoc -c .jsdoc.js",
     "lint": "eslint samples/ system-test/ && gts check",
     "cover": "nyc --reporter=lcov mocha --require intelli-espower-loader build/test/*.js && nyc report",
-    "test-no-cover": "mocha build/test/",
+    "test-no-cover": "npm run pretest && mocha build/test/",
```
JustinBeckwith: So what I'd rather do here is something like this:
https://github.com/googleapis/nodejs-firestore/pull/252

Check out the `package.json` and how I have it set up there.  For any npm script where you want to run something before it, you can create a `pre<NAME OF SOME RUN SCRIPT>` script, and it will automatically run before :)

JustinBeckwith: It's also worth looking at auth for TypeScript best practice:
https://github.com/google/google-auth-library-nodejs

kinwa91: Thanks! that's much cleaner

PR #291, opened 2018-07-11
TypeScript: make Bucket test pass

kinwa91:


kinwa91:
codecov issues fixed via #293

PR #294, opened 2018-07-11
TypeScript: make test/file.ts pass

kinwa91:
Should be last of the test fixes, after this is merge all tests should pass.


kinwa91:
@JustinBeckwith just one review will be fine :) don't know how I managed to assign the same person twice for review 🤷🏻‍♂️

PR #307, opened 2018-07-19
Typescript: workaround explicit 'any' cast in tests

jkwlui:


test/bucket.ts
```diff
@@ -44,10 +44,12 @@ function FakeFile(bucket, name, options?) {
   this.createWriteStream = options => {
     self.metadata = options.metadata;
     const ws = new stream.Writable();
-    (ws as any).write = () => {
-      ws.emit('complete');
-      ws.end();
-    };
+    extend(ws, {
```
JustinBeckwith: This doesn't feel right.  Instead of this, could we create an interface that has the `write` method and use that?

test/bucket.ts
```diff
@@ -62,24 +64,32 @@ let requestOverride;
...
-  // call.
-  return fakeRequest;
-};
-(fakeRequest as any).get = (...args) => {
-  return (requestOverride.get || fakeRequest).apply(null, args);
-};
-(fakeRequest as any).head = (...args) => {
-  return (requestOverride.head || fakeRequest).apply(null, args);
-};
+extend(fakeRequest, {
```
JustinBeckwith: same feedback as above.  Using the lack of type safety on extend feels like a hack, and I'd rather define an interface.

test/bucket.ts
kinwa91: That's a good point. `WriteStream#write()` is actually a thing, we just need to replace `util.noop` to a function that returns a `boolean`. Addressed in bd2911b

test/bucket.ts
kinwa91: https://github.com/googleapis/nodejs-storage/pull/307#discussion_r203872303

JustinBeckwith (approved):

PR #308, opened 2018-07-19
test: use assert.deepStrictEqual in tests

jkwlui:


samples/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "author": "Google Inc.",
   "engines": {
-    "node": ">=4"
+    "node": ">=7.6"
```
JustinBeckwith: That's a weird number 😆 I'd say either we support 6+ or 8+.

JustinBeckwith (approved):
LGTM w/ nit

samples/package.json
kinwa91: I targeted lowest version that supports async/await hmm.. I'll make it 8+ then.

PR #320, opened 2018-08-02
Support Bucket/Object lock operations

stephenplusplus:
#### To Dos
- Tests
  - [x] System
  - [ ] Unit

#### References
- https://docs.google.com/document/d/1wRpiAxTHZoypR1NLVP97Cj0WCGVKx43Tivso0xXhyEA/edit?ts=5b60f543#
- https://www.googleapis.com/discovery/v1/apis/storage/v1/rest

This introduces new behavior:

# Buckets

- **bucket#lock()**
Lock a previously-defined retention policy. This will prevent changes to the policy.

- **bucket#removeRetentionPeriod()**
Remove an already-existing retention policy from this bucket.

- **bucket#setRetentionPeriod(durationInSeconds)**
Lock all objects contained in the bucket, based on their creation time.

# Files

- **file#getExpirationDate()**
Get a Date object representing the earliest time this file will expire.

---

Additionally, `bucket#getMetadata()`, `bucket#setMetadata()`, `file#getMetadata()`, and `file#setMetadata()` are available if users wish to interact with the raw resource schemas as defined by the API.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1614,15 +1614,14 @@ class Bucket extends ServiceObject {
       }
 
       this.request(
-        {
-          method: 'POST',
-          uri: '/lockRetentionPolicy',
-          qs: {
-            ifMetagenerationMatch: metadata.metageneration,
+          {
+            method: 'POST',
+            uri: '/lockRetentionPolicy',
+            qs: {
+              ifMetagenerationMatch: metadata.metageneration,
+            },
           },
-        },
-        callback
-      );
+          callback);
     });
   }
 
@@ -1920,11 +1919,10 @@ class Bucket extends ServiceObject {
    */
   removeRetentionPeriod(callback) {
     this.setMetadata(
-      {
-        retentionPolicy: null,
-      },
-      callback
-    );
+        {
+          retentionPolicy: null,
+        },
+        callback);
   }
 
   /**
@@ -2135,13 +2133,12 @@ class Bucket extends ServiceObject {
    */
   setRetentionPeriod(duration, callback) {
     this.setMetadata(
-      {
-        retentionPolicy: {
-          retentionPeriod: duration,
+        {
+          retentionPolicy: {
+            retentionPeriod: duration,
+          },
         },
-      },
-      callback
-    );
+        callback);
   }
 
   /**
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1896,7 +1896,8 @@ class Bucket extends ServiceObject {
   }
 
   /**
-   * Remove an already-existing retention policy from this bucket.
+   * Remove an already-existing retention policy from this bucket, if it is not
+   * locked.
    *
    * @param {SetBucketMetadataCallback} [callback] Callback function.
    * @returns {Promise<SetBucketMetadataResponse>}
@@ -2103,7 +2104,8 @@ class Bucket extends ServiceObject {
    * {@link File#removeRetentionPeriod} and {@link File#setRetentionPeriod}. A
    * locked retention policy cannot be removed or shortened in duration for the
    * lifetime of the bucket. Attempting to remove or decrease period of a locked
-   * retention policy will result in a `PERMISSION_DENIED` error.
+   * retention policy will result in a `PERMISSION_DENIED` error. You can still
+   * increase the policy.
    *
    * @param {*} duration In seconds, the minimum retention time for all objects
    *     contained in this bucket.
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1601,12 +1601,14 @@ class Bucket extends ServiceObject {
    * const storage = require('@google-cloud/storage')();
    * const bucket = storage.bucket('albums');
    *
-   * bucket.lock(function(err, apiResponse) {});
+   * const metageneration = 2;
+   *
+   * bucket.lock(metageneration, function(err, apiResponse) {});
    *
    * //-
    * // If the callback is omitted, we'll return a Promise.
    * //-
-   * bucket.lock().then(function(data) {
+   * bucket.lock(metageneration).then(function(data) {
    *   const apiResponse = data[0];
    * });
    */
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1455,7 +1455,7 @@ class Bucket extends ServiceObject {
   /**
    * @callback GetBucketMetadataCallback
    * @param {?Error} err Request error, if any.
-   * @param {object} files The bucket metadata.
+   * @param {object} metadata The bucket metadata.
    * @param {object} apiResponse The full API response.
    */
   /**
@@ -1590,6 +1590,9 @@ class Bucket extends ServiceObject {
    * Lock a previously-defined retention policy. This will prevent changes to
    * the policy.
    *
+   * @param {SetBucketMetadataCallback} [callback] Callback function.
+   * @returns {Promise<SetBucketMetadataResponse>}
+   *
    * @example
    * const storage = require('@google-cloud/storage')();
    * const bucket = storage.bucket('albums');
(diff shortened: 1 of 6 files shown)


src/bucket.js
```diff
@@ -1799,6 +1836,34 @@ class Bucket extends common.ServiceObject {
     return new Notification(this, id);
   }
 
+  /**
+   * Remove an already-existing retention policy from this bucket.
```
frankyn: Only possible if policy is not locked.

src/bucket.js
```diff
@@ -1960,6 +2032,51 @@ class Bucket extends common.ServiceObject {
...
+  /**
+   * Lock all objects contained in the bucket, based on their creation time. Any
+   * attempt to overwrite or delete objects younger than the retention period
+   * will result in a `PERMISSION_DENIED` error.
+   *
+   * An unlocked retention policy can be modified or removed from the bucket via
+   * {@link File#removeRetentionPeriod} and {@link File#setRetentionPeriod}. A
+   * locked retention policy cannot be removed or shortened in duration for the
+   * lifetime of the bucket. Attempting to remove or decrease period of a locked
+   * retention policy will result in a `PERMISSION_DENIED` error.
```
frankyn: Retention policy can be increased when the retention policy is locked.

src/file.js
```diff
@@ -1827,6 +1871,41 @@ class File extends common.ServiceObject {
...
+   * (i.e. this value is set to false (see @{link File#release})). After being
+   * released, this object will be subject to bucket-level retention (if any).
+   *
+   * Alternatively, you may set a temporary hold. This will follow the same
+   * behavior as an event-based hold, with the exception that the bucket's
+   * retention policy will not renew for this file from the time the hold is
+   * released.
+   *
+   * @param {object} [options] Configuration object.
+   * @param {boolean} [options.temporary=false] - Set a temporary hold.
```
frankyn: Should be set to true in documentation, I think?

src/file.js
```diff
@@ -1827,6 +1871,41 @@ class File extends common.ServiceObject {
...
+   * @param {SetFileMetadataCallback} [callback] Callback function.
+   * @returns {Promise<SetFileMetadataResponse>}
+   */
+  hold(options, callback) {
```
frankyn: This is subtle to me, is this considered idiomatic for Node.js?

src/bucket.js
stephenplusplus: This is actually a copy/paste of what is in the REST docs: https://www.googleapis.com/discovery/v1/apis/storage/v1/rest

I updated ours to say "You can still increase the policy." at the end, although I think it is implied by the previous mentions of "...cannot be removed or shortened..." and "Attempting to remove or decrease...".

Just mentioning in case you want to make the option to increase the policy more pronounced in the REST docs as well.

src/file.js
stephenplusplus: By default, it is set to false in our code:

```js
if (options.temporary) {
  metadata.temporaryHold = true;
} else {
  metadata.eventBasedHold = true;
}
```

Should we flip it, so that by default, a "file.hold()" creates a temporary hold, as opposed to an event-based hold?

src/file.js
stephenplusplus: Would you mind elaborating?

frankyn: Sorry, I meant, given that each hold is different, could the surface API have separate methods?

src/bucket.js
frankyn: Thanks for the feedback, I'll forward this!

src/file.js
frankyn: My mistake please ignore*

src/file.js
frankyn: I like the idea of having these convenience methods, but it might be overdoing it as a user can update these holds using a setMetadata. What are your thoughts? I'm unsure honestly given the rest of the surface. Looking for input :)

cc: @mohammadnasre

frankyn: So disregard the mention of removing convenience methods. It makes sense here. 

Pending additional feedback from Mohammad.
I think, unifying methods is hard to understand, because at first glance I don't understand what kind of hold the method apply's. 

As a developer, I'd appreciate getting a hint in the name as to what the hold does given bucket lock surfaces two holds.

stephenplusplus: I started fleshing this out into multiple convenient methods. I realized the methods were such simple calls to `setMetadata`, e.g.:

```js
file.setTemporaryHold = function () { file.setMetadata({ temporaryHold: true }) }
file.setEventBasedHold = function () { file.setMetadata({ eventBasedHold: true }) }

file.releaseTemporaryHold = function () { file.setMetadata({ temporaryHold: false }) }
file.releaseEventBasedHold = function () { file.setMetadata({ eventBasedHold: false }) }
```

Seeing that, I started to lean a bit towards skipping the convenience methods, as we're not really adding that much convenience. We could just show an example in the docs for `setMetadata`?

frankyn: I think it add obfuscation, and I appreciate your input. Move forward with documentation instead. This is something we shouldn't over complicate if it doesn't help at all. It's more of an alias that could also help introduce developers to manipulating other metadata.

frankyn: One additional note, system tests should verify how the client library behaves when holds are enabled and attempting to perform an overwrite and delete.

I have additional tests if you'd like to reference tests in Java [PR#3574](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3574).

frankyn: Friendly ping, @stephenplusplus.

src/bucket.js
```diff
@@ -1525,6 +1525,43 @@ class Bucket extends common.ServiceObject {
...
 
+  /**
+   * Lock a previously-defined retention policy. This will prevent changes to
+   * the policy.
+   *
+   * @example
+   * const storage = require('@google-cloud/storage')();
+   * const bucket = storage.bucket('albums');
+   *
+   * bucket.lock(function(err, apiResponse) {});
```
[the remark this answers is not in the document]
stephenplusplus: As far as I can tell, and in my own experience, yes. But, I can't confirm definitively.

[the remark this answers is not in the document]
frankyn: Sorry, I'll rephrase, `storage.bucket('albums')` always returns bucket metadata unless something on the API changes or user does something different.

The issue with lock is it's a one time thing for every bucket. Once locked it can't be unlocked. A requirement is that it has the metageneration for the bucket to pass as a precondition.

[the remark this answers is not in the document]
stephenplusplus: Woops, sorry for the misunderstanding.

`storage.bucket('albums')` won't actually return any metadata through our API. The user can call, `storage.bucket('albums').getMetadata()` which will update the `metadata` property on that Bucket's instance.

In the code within `lock()`, we don't check for `bucket.metadata.metageneration` and instead just always do a metadata refresh to get the latest value. Is there a better way to handle that?

stephenplusplus:
@frankyn I've added the system tests to verify the hold behavior will block overwrite and delete operations on File objects. I've also removed `file.hold()` and `file.release()` and updated the docs for `file.setMetadata()`.

src/bucket.js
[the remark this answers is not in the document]
frankyn: Thanks, could you make metageneration a parameter instead for retentionLock() and getMetadata() before calling this method to pass in the generation? I'd rather this operation not do a metadata get as well. It defeats the purpose of optimistic concurrency in this case.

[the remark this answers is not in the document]
stephenplusplus: Done!

src/bucket.ts
```diff
@@ -1586,6 +1586,46 @@ class Bucket extends ServiceObject {
...
+   * @example
+   * const storage = require('@google-cloud/storage')();
+   * const bucket = storage.bucket('albums');
+   *
+   * bucket.lock(function(err, apiResponse) {});
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * bucket.lock().then(function(data) {
```
frankyn: Last nit to update sample. Surface LGTM, thanks @stephenplusplus!

stephenplusplus: 🤦‍♂️ Updated!

PR #334, opened 2018-08-10
refactor(ts): merge initial TypeScript conversion

kinwa91:
BREAKING CHANGE:

Storage will need to be imported as a named import:
```js
const { Storage } = require('@google-cloud/storage');
```

Storage cannot be instantiated without new:
```js
𝘅 const storage = Storage();  // deprecated
✔ const storage = new Storage();
```


kinwa91:
[Do not merge yet] - some system test failures after merging from master.

JustinBeckwith:
👋 How's this going?

kinwa91:
I am still tracking down those places within our docs that uses the old initialization style.

kinwa91:
@JustinBeckwith i think you might have to force merge this since lint is still failing.

sabrehagen:
When will this be released?

PR #335, opened 2018-08-10
refactor: use node-fetch instead of request

fhinkel:
Use `node-fetch` instead of `request` for calls that
do not require authentication. For those we use
`request` from `nodejs-common`.

Move `request` to devDependencies. System tests still
use `request`.


src/bucket.js
```diff
@@ -2297,20 +2297,21 @@ class Bucket extends common.ServiceObject {
...
-
-        const contentLength = resp.headers['content-length'];
-
-        if (is.number(contentLength)) {
-          options.resumable = contentLength > RESUMABLE_THRESHOLD;
-        }
+      const url = requestOptions.url;
+      const method = 'HEAD';
+      const headers = requestOptions.headers;
+      const redirect = requestOptions.followAllRedirects ? 'follow' : 'manual';
```
JustinBeckwith: I think `request` follows all redirects by default.  Would it make sense to always use `follow` here?

src/bucket.js
```diff
@@ -2297,20 +2297,21 @@ class Bucket extends common.ServiceObject {
...
+      fetch(url, {method, headers, redirect})
```
JustinBeckwith: I do not believe fetch will follow proxies by default :(  `request` does this for us, and it's one of the reasons `axios` just isn't going to work out.  I am starting to think we should create a `google-http-client` npm module that wraps our retry logic, backoff, proxy support, and redirect settings in a single place.

src/bucket.js
```diff
@@ -2326,15 +2327,32 @@ class Bucket extends common.ServiceObject {
...
     function upload() {
-      let sourceStream;
-
       if (isURL) {
-        sourceStream = request.get(requestOptions);
-      } else {
-        sourceStream = fs.createReadStream(pathString);
+        const url = requestOptions.url;
+        const method = 'GET';
+        const headers = requestOptions.headers;
```
JustinBeckwith: This feels like a lot of boilerplate.  Instead of extracting url, method, and headers from the request, why not just pass the requestOptions directly?

src/bucket.js
```diff
@@ -2326,15 +2327,32 @@ class Bucket extends common.ServiceObject {
...
+              callback(err);
+            });
+            dest.on('finish', () => {
+              callback(null, newFile, newFile.metadata);
+            });
+            dest.on('error', err => {
+              callback(err);
+            });
+          })
+          .catch(err => callback(err));
```
JustinBeckwith: I think this could just be `.catch(callback)`

test/bucket.js
```diff
@@ -60,22 +60,11 @@ function FakeNotification(bucket, id) {
   this.id = id;
 }
 
-const requestCached = request;
-let requestOverride;
-function fakeRequest() {
-  return (requestOverride || requestCached).apply(null, arguments);
+const fetchCached = fetch;
```
JustinBeckwith: If we're going to rewrite this code, I am personally *much* more fond of using `nock` if mocking the http transport library.  That way we're protected from whatever future changes we choose to make here.  Really interested to hear others thoughts :)

test/file.js
```diff
@@ -61,19 +60,6 @@ const fakeUtil = extend({}, util, {
 const fsCached = extend(true, {}, fs);
 const fakeFs = extend(true, {}, fsCached);
 
-let REQUEST_DEFAULT_CONF; // eslint-disable-line no-unused-vars
```
JustinBeckwith: It's awesome that this could just be deleted 🤣

src/bucket.js
fhinkel: Yes to putting it in a library, but this one doesn't do retry or backoff. Let's keep it simple. It's fetching images form somewhere not GCP related, no need to add a bunch of logic.

src/bucket.js
fhinkel: Good catch, I'll use it as the default.

src/bucket.js
fhinkel: Being verbose to show which options are understood by `fetch`. Except for the most basic ones they have different names.

test/bucket.js
fhinkel: Makes sense. Let's introduce a different mocking library in a separate PR.

src/bucket.js
JustinBeckwith: For retry/backoff in this case, that may be fine.  The problem is proxies :(

src/bucket.js
```diff
@@ -2326,15 +2327,32 @@ class Bucket extends common.ServiceObject {
     }
 
     function upload() {
-      let sourceStream;
-
       if (isURL) {
-        sourceStream = request.get(requestOptions);
-      } else {
-        sourceStream = fs.createReadStream(pathString);
+        const url = requestOptions.url;
```
stephenplusplus: Line 2108 needs to be updated-- it shows that we accept the options that `request` does. Considering that, this will be a breaking change unless we convert all `request` options to match `node-fetch`.

src/bucket.js
stephenplusplus: I would vote for being simple here. If it's possible to just pass the user's options directly through to node-fetch, that would be the best for all of us:

- Easy docs (just link to node-fetch's docs)
- Simple code (no mapping)

fhinkel:
Closing in favor of https://github.com/googleapis/nodejs-storage/pull/337

PR #337, opened 2018-08-15
semver: do not support upload() from url

fhinkel:
Remove the upload() from url functionality. Instead, use
File.createWriteStream.

For performance reasons, nodejs-storage should not require the `request`
module. As we are currently exposing `request` as part of the public
API for `upload()`, not using request is a breaking change. Instead of
using a different dependency for the http requests, we leave this decision
to the users.

If you were using `upload()` from a url, replace it with the following:

```
const request = require('request');
let readStream = request(url).createReadStream();
const file = bucket.file(name)
const writeStream = file.createWriteStream();
readStream.pipe(writeStream)
```

Use `node-fetch` instead of `request` in system tests.


fhinkel:
Sorry, didn't include the reasoning, see updated commit message.

stephenplusplus:
Heads up @Kiricon-- we're going to be removing the feature you added for us. Do you think it's a mistake? The dilemma is that it forces us to keep a dependency on `request`, which we are trying to avoid (it's big, and it has introduced some security concerns).

We have a couple of options:

1) Get rid of the feature. The user can handle fetching the data themselves, and pipe it into a `file.createWriteStream()`. This is definitely not as easy as "put this remote file in my bucket". The user will have to create a File object, and familiarize themselves with the streaming APIs that both our library and their possibly-new HTTP-fetching library.
2) Use a different request library, but somehow remain library-agnostic when accepting common HTTP options, like `headers`.

Thoughts and ideas welcome!

fhinkel:
It's been a week with no objections, can we merge this?

nnmrts:
I consider this as a mistake. There are no "perfomance reasons" to do this. It's now harder to upload files, without any benefits performance-wise. Welp.

PR #340, opened 2018-08-22
fix(deps): update dependency @google-cloud/common to ^0.23.0

fhinkel:
This Pull Request updates dependency @google-cloud/common (source) from ^0.21.0 to ^0.23.0.

Fixes https://github.com/googleapis/nodejs-storage/pull/340


src/bucket.ts
```diff
@@ -340,6 +340,7 @@ class Bucket extends ServiceObject {
       id: name,
       createMethod: storage.createBucket.bind(storage),
       methods,
+      requestModule: request,
```
JustinBeckwith: Is this strictly required?  For each library, does that mean we need to go pull in request now?

kinwa91: @fhinkel

kinwa91:
Feel free to ignore kokoro failures for now. We haven't set it up for storage just yet.

src/bucket.ts
fhinkel: Yes. For now, instead of `common`, each client library needs to provide an HTTP dependency. Of course, if not needed, you can provide an empty object here.

PR #349, opened 2018-09-04
build: run npm install when publishing

jkwlui:
Fixes problems publishing - i manually published 2.0.0, but we should fix it for next releases.


JustinBeckwith:
We should be using a synth script here, and the synth script should make sure this "just works" ™️.  I submitted a PR for that over in #351

PR #357, opened 2018-09-04
build: fix multi-line command in circle.yml

kinwa91:
Sorry! 😅


kinwa91:
Will fix lint on Kokoro in a later PR (due to running as `root`)

PR #380, opened 2018-09-11
Make some parameters optional

ThomasdenH:
- [ ] Tests and linter pass

Marks some optional parameters as such. This PR is by no means exhaustive but at least an incremental improvement.


ThomasdenH:
I signed it!

PR #389, opened 2018-09-13
fix: copy(): Use correct destination file name in URI

### LINKED ISSUE #355: Copy treats leading slash differently than Upload
If I upload to `/js/test.js` then it winds up in <em>gs://bucket-name/js/test.js</em>. However, if I then perform a copy to `/js/foo.js`, it winds up in <em>gs://bucket-name/<strong>/</strong>js/foo.js</em>

![screen shot 2018-09-04 at 3 32 09 pm](https://user-images.githubusercontent.com/798836/45053350-bdeefe80-b057-11e8-81ae-a8b33043f2b3.png)

I ran into this while batch moving over a hundred files. What was particularly odd was that some files would still be served through the LB correctly, while others would reply with `NoSuchKey`. 

#### Environment details

  - OS: macOS 10.13.6 (17G65)
  - Node.js version: v8.11.3
  - npm version: 5.6.0
  - `@google-cloud/storage` version: 1.7.0

#### Steps to reproduce

```javascript
const Storage = require('@google-cloud/storage');                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                
var config = {                                                                                                                                                                                                                                                  
  projectId: 'projectid',                                                                                                                                                                                                                                
  keyFilename: 'key-file.json'                                                                                                                                                                                 
};                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                
// Creates a client                                                                                                                                                                                                                                             
const storage = new Storage(config);                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                
const testObjectName = '/js/test.js';                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                
storage                                                                                                                                                                                                                                                         
  .bucket(bucketName)                                                                                                                                                                                                                                           
  .upload('.' + testObjectName, {                                                                                                                                                                                                                               
    destination: testObjectName                                                                                                                                                                                                                                 
  })                                                                                                                                                                                                                                                            
  .then(data => {                                                                                                                                                                                                                                               
    console.log('upload success:', data);                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                
//  without copy, file is written to gs://bucket-name/js/test.js                                
//    return Promise.all([                                                                                                                                                                                                                                      
//      storage                                                                                                                                                                                                                                                 
//        .bucket(bucketName)                                                                                                                                                                                                                                   
//        .file(testObjectName)                                                                                                                                                                                                                                 
//        .move('/js/foo.js')                                                                                                                                                                                                                                   
//    ]);                                                                                                                                                                                                                                                       
//  after copy, file is in gs://bucket-name//js/foo.js instead of /js/foo.js                    
  })                                                                                                                                                                                                                                                            
  .catch(console.error);      
```

stephenplusplus:
Fixes #355

cc @Alfredo-Delgado

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -530,7 +530,7 @@ class File extends ServiceObject {
         {
           method: 'POST',
           uri: `/rewriteTo/b/${destBucket.name}/o/${
-              encodeURIComponent(destName)}`,
+              encodeURIComponent(newFile.name)}`,
           qs: query,
           json: options,
           headers,
--- a/test/file.ts
+++ b/test/file.ts
@@ -453,9 +453,10 @@ describe('File', () => {
       }
 
       it('should allow a string', done => {
-        const newFileName = 'new-file-name.png';
+        const newFileName = '/new-file-name.png';
+        const newFile = new File(BUCKET, newFileName);
         const expectedPath =
-            `/rewriteTo/b/${file.bucket.name}/o/${newFileName}`;
+            `/rewriteTo/b/${file.bucket.name}/o/${newFile.name}`;
         assertPathEquals(file, expectedPath, done);
         file.copy(newFileName);
       });


src/file.ts
```diff
@@ -530,7 +530,7 @@ class File extends ServiceObject {
         {
           method: 'POST',
           uri: `/rewriteTo/b/${destBucket.name}/o/${
-              encodeURIComponent(destName)}`,
+              encodeURIComponent(newFile.name)}`,
```
stephenplusplus: The `File` constructor already parses out extra slashes from a given object name, so we can just re-use that here.

PR #391, opened 2018-09-14
Annotate types #3

kinwa91:


src/bucket.ts
```diff
@@ -337,6 +337,115 @@ export interface BucketExistsCallback {
...
+export interface GetBucketRequest extends GetConfig {
+  userProject?: string;
+}
+
+/**
+ * @typedef {array} GetBucketResponse
+ * @property {Bucket} 0 The {@link Bucket}.
+ * @property {object} 1 The full API response.
+ */
+export type GetBucketResponse = [Bucket, object];
```
JustinBeckwith: Why `object`?  Can we just use `request.Response` here since that's what we're actually passing?

src/bucket.ts
```diff
@@ -337,6 +337,115 @@ export interface BucketExistsCallback {
...
+
+/**
+ * @callback GetBucketCallback
+ * @param {?Error} err Request error, if any.
+ * @param {Bucket} bucket The {@link Bucket}.
+ * @param {object} apiResponse The full API response.
+ */
+export interface GetBucketCallback extends InstanceResponseCallback {
+  (err: Error, bucket: null, apiResponse: object);
```
JustinBeckwith: No need to have two lines here (I think).  You should be able to say:

```js
(err: Error|null, bucket: Bucket|null, apiResponse?: r.Request);
```

src/bucket.ts
```diff
@@ -337,6 +337,115 @@ export interface BucketExistsCallback {
...
+ */
+export interface GetLabelsRequest {
+  userProject?: string;
+}
+
+/**
+ * @typedef {array} GetLabelsResponse
+ * @property {object} 0 Object of labels currently set on this bucket.
+ */
+export type GetLabelsResponse = [object];
```
JustinBeckwith: Everywhere there's object, lets use r.Response

src/bucket.ts
```diff
@@ -1393,38 +1487,42 @@ class Bucket extends ServiceObject {
...
-    if (is.fn(options)) {
+  get(options?: GetBucketRequest): Promise<GetBucketResponse>;
+  get(callback: GetBucketCallback);
+  get(options: GetBucketRequest, callback: GetBucketCallback);
+  get(options?: GetBucketRequest|GetBucketCallback,
+      callback?: GetBucketCallback): Promise<GetBucketResponse>|void {
+    if (typeof options === 'function') {
       callback = options;
-      options = {};
+      options = {} as GetBucketRequest;
```
JustinBeckwith: Why's that cast needed? Seems like it shouldn't be.

src/bucket.ts
```diff
@@ -1393,38 +1487,42 @@ class Bucket extends ServiceObject {
    *   const apiResponse = data[1];
    * });
    */
-  get(options, callback?) {
-    if (is.fn(options)) {
+  get(options?: GetBucketRequest): Promise<GetBucketResponse>;
+  get(callback: GetBucketCallback);
+  get(options: GetBucketRequest, callback: GetBucketCallback);
+  get(options?: GetBucketRequest|GetBucketCallback,
```
JustinBeckwith: Ah I see - Just to make this crystal clear, please rename `options` to `optionsOrCallback`.  Then create a new variable for `options` in the method.

src/bucket.ts
```diff
@@ -1611,43 +1700,38 @@ class Bucket extends ServiceObject {
    *   const labels = data[0];
    * });
    */
-  getLabels(options, callback?) {
-    if (is.fn(options)) {
+  getLabels(options: GetLabelsRequest): Promise<GetLabelsResponse>;
+  getLabels(callback: GetLabelsCallback);
+  getLabels(options: GetLabelsRequest, callback: GetLabelsCallback);
+  getLabels(
```
JustinBeckwith: Same feedback as above for `optionsOrCallback`

kinwa91:
Fixed the following via the last few commits:
 - [x] Remove overloads in Callbacks
 - [x] Rename options args to `optionsOrCallback` to match with the actual overloaded type
 - [x] Rename -Request interfaces to -Options to match with name of argument (`options`)
 - [x] `apiResponse` typed `request.Response` instead of `object`

PR #396, opened 2018-09-16
samples: don't use USA formatted dates for expiry

broady:
Found via https://twitter.com/iaincollins/status/1040247045246930946
"I ran into the Google Cloud SDK documentation (and code examples!) using MM-DD-YYYY yesterday"

Docs state:
"A timestamp when this policy will expire. Any value given is passed to new Date()."

### DIFF
--- a/samples/files.js
+++ b/samples/files.js
@@ -339,7 +339,7 @@ function generateSignedUrl(bucketName, filename) {
   // These options will allow temporary read access to the file
   const options = {
     action: 'read',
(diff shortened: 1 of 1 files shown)


broady:
- [ ] Tests and linter pass

broady:
Wow, screw you, linter:

```
/home/node/project/samples/files.js
  342:27  error  Replace `(1000·*·60·*·60)` with `1000·*·60·*·60`  prettier/prettier
```

Fine, I'll replace the readable code with less readable code.

PR #401, opened 2018-09-19
optionsOrCallback could be undefined if not given, check before assign

kinwa91:
Hoping this fixes system-test

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -435,6 +435,7 @@ class File extends ServiceObject {
    * Another example:
    */
   copy(destination: string|Bucket|File, callback: FileCallback);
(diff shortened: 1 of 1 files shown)


src/file.ts
```diff
@@ -435,6 +435,7 @@ class File extends ServiceObject {
    * Another example:
    */
   copy(destination: string|Bucket|File, callback: FileCallback);
+  copy(destination: string|Bucket|File, callback: FileCallback);
```
JustinBeckwith: I am confused.  This looks like the same declaration twice?

kinwa91: My bad. That wasn't suppose to be there.

kinwa91:
Look's like it didn't fix the system-test.

PR #402, opened 2018-09-19
Improve TypeScript types (part 4)

JustinBeckwith:


src/bucket.ts
```diff
@@ -1644,7 +1653,8 @@ class Bucket extends ServiceObject {
         },
         (err, resp) => {
           if (err) {
-            callback(err, null, null, resp);
+            // tslint:disable-next-line:no-any
+            (callback as any)(err, null, null, resp);
```
kinwa91: This is a sad hack ☹️I suppose we can fix that later with how we call upon the paginator

JustinBeckwith: Lol yeah, it ain't pretty.  First pass, I want to get all the types we know are missing in.  We can come back and clean up the weird `any` stuff laster.  I don't mind as much here since this is internal.

JustinBeckwith:
Side note - my system tests appear to be passing?

kinwa91:
Samples test has always been flaky on that one test where it complains about unlink `downloaded.txt`..

src/bucket.ts
```diff
@@ -1624,12 +1629,16 @@ class Bucket extends ServiceObject {
...
    * Example of listing files, filtered by a prefix:
    */
-  getFiles(query: GetFilesOptions, callback?) {
+  getFiles(query?: GetFilesOptions): Promise<[File[]]>;
+  getFiles(query: GetFilesOptions, callback: GetFilesCallback): void;
+  getFiles(callback: GetFilesCallback): void;
+  getFiles(
+      queryOrCallback?: GetFilesOptions|GetFilesCallback,
+      callback?: GetFilesCallback): void|Promise<[File[]]> {
+    let query = typeof queryOrCallback === 'object' ? queryOrCallback : {};
```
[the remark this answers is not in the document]
kinwa91: oh it never expects `null`. Then all is good!

PR #409, opened 2018-09-24
Add typing for File#download()

ThomasdenH:
This adds Typescript types for File#download().

- [x] Tests and linter pass


ThomasdenH:
Running `gts fix` solved all CI issues.

ThomasdenH:
Is this under consideration to be merged?

src/file.ts
```diff
@@ -1306,15 +1319,6 @@ class File extends ServiceObject {
     (this.parent as ServiceObject).delete.call(this, options, callback);
   }
 
-  /**
```
JustinBeckwith: Can we keep the jsdoc comments, and move them up with the type definitions?

ThomasdenH:
I believe I have moved all JSDoc now. Do you want me to duplicate their types, considering they are already defined by Typescript?

PR #416, opened 2018-09-28
feat: use small HTTP dependency

fhinkel:


JustinBeckwith:
Nice!  The system and sample tests are currently down for this one, and we would need to get those passing first though :)

package.json
```diff
@@ -94,6 +94,7 @@
     "request": "^2.88.0",
```
JustinBeckwith: Wouldn't we want to move request over to dev dependencies?

fhinkel:
I was wondering how that tests ran so fast. No problem with waiting. How working on fixing the system tests?

package.json
fhinkel: Ideally 🙈. I wrote down that `gcs-resumable-upload` depends on it. Let me check.

JustinBeckwith: Should we maybe do that library first?

JustinBeckwith:
@fhinkel I think we have the tests worked out over here.  Can you resolve conflicts and give this another go?  Thanks!

PR #418, opened 2018-09-29
Bucket lock samples and tests

frankyn:
## Bucket Lock Samples
In collaboration with @billyjacobson.

- [x]  Code coverage does not decrease (@stephenplusplus -- what are the expectations for code coverage w.r.t to samples?)
- [x]  Appropriate docs were updated (@stephenplusplus -- not sure what needs to be done here). These samples will be part of documentation for GCS but not sure if I need to do additional work.
- [x]  Tests
- [x]  linter pass
- [x]  Define a retention policy
- [x]  Remove a retention policy if not locked
- [x]  Lock a retention policy
- [x]  Set temporary hold
- [x]  Release temporary hold
- [x]  Set event-based hold
- [x]  Release event-based hold
- [x]  Set default event-based hold
- [x]  Release default event-based hold
- [x]  View default event-based hold
- [x]  View bucket retention policy metadata
- [x]  View metadata for object
- [x]  Updated CLI
- [x]  Add CLI examples
- [x]  Updated README.md


stephenplusplus:
> - [ ] Appropriate docs were updated (@stephenplusplus -- not sure what needs to be done here). This samples will be part of documentation for GCS but not sure if I need to do additional work.

/cc @jmdobry about @frankyn's question above.

-

RE code coverage, samples aren't included in the coverage report, and aren't required AFAIK, so no worries.

samples/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@google-cloud/pubsub": "*",
-    "@google-cloud/storage": "^2.0.3",
+    "@google-cloud/storage": "+git",
```
frankyn: @stephenplusplus, not sure what to do here. Should I revert or change this before merging?

stephenplusplus: You can just keep this the same as it was. When we cut a release, the samples/package.json dependency version is updated automatically.

To run the sample tests locally, `npm run samples-test` from the root directory of the repo should handle the linking automatically!

frankyn:
Synced with @JustinBeckwith about the open question @stephenplusplus and it's a no-op.

PR #422, opened 2018-10-01
chore(samples): convert samples to async/await

JustinBeckwith:


kinwa91:
We got rid of catching when the promise rejects - I'm concerned it might breaks some test that assert that the operation failed or resulting in `UnhandledPromiseRejectionWarning`.

kinwa91:
Samples tests are passing on master with the recent fixes, and I've updated this branch. So we can start tackling some of the test failures as a result of this PR. :)

samples/files.js
```diff
@@ -343,22 +297,16 @@ function generateSignedUrl(bucketName, filename) {
   };
 
   // Get a signed URL for the file
-  storage
+  const [url] = storage
```
kinwa91: `const [url] = await storage`

samples/iam.js
```diff
@@ -125,44 +112,36 @@ function removeBucketIamMember(bucketName, roleName, members) {
...
-        `Removed the following member(s) with role ${roleName} from ${bucketName}:`
-      );
-      members.forEach(member => {
-        console.log(`  ${member}`);
-      });
-    })
-    .catch(err => {
-      console.error('ERROR:', err);
-    });
+  const [policy] = bucket.iam.getPolicy();
```
kinwa91: `const [policy] = await bucket.iam.getPolicy();`

PR #431, opened 2018-10-05
fix(tests): fix system tests

kinwa91:
This seems to have fixed the system test failures running CircleCI.
Let's get this in so we can debug the Kokoro system test failures.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,7 @@ jobs:
       - run: *npm_install_and_link
       - run:
           name: Run system tests.
-          command: npm run system-test
+          command: mkdir $HOME/.config && npm run system-test
           environment:
             GCLOUD_PROJECT: long-door-651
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - run: *npm_install_and_link
       - run:
           name: Run system tests.
-          command: npm run system-test
+          command: mkdir $HOME/.config && npm run system-test
```
JustinBeckwith: That's weird.  Why do we need this?

kinwa91: We're starting to see CI failures due to
```
Uncaught ResumableUploadError: A resumable upload could not be performed. The directory, /home/node/.config, is not writable. You may try another upload, this time setting `options.resumable` to `false`.
```

What's weirder is I've went back to previous commits where we didn't see this error and restarted the runs in CircleCI, and they all failed with this error.

I couldn't reproduce this error locally, and it doesn't happen on Kokoro.

PR #444, opened 2018-10-12
Revert "feat: use small HTTP dependency (#416)"

fhinkel:
System-tests broke with that commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,9 +94,9 @@
     "mime-types": "^2.0.8",
     "once": "^1.3.1",
     "pumpify": "^1.5.1",
+    "request": "^2.88.0",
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
So I have good news, and I have bad news.  The good news is that I re-ran the system tests, and all is green on master!  
http://long-door-651.appspot.com/

The bad news is that our tests are flaky. I filed #445 to track this, but I am thinking we should let this PR stay in.  Thoughts?

fhinkel:
Oh lol, thanks for re-running them. 

The system tests mostly up- and download files from GCS. That's a very deterministic process. I assume we have some race conditions between tests interfering with other tests. That should be easy to fix.

PR #461, opened 2018-10-17
ts: Annotate Iam types

kinwa91:


src/iam.ts
```diff
@@ -257,42 +316,55 @@ class Iam {
...
         (err, resp) => {
           if (err) {
-            callback(err, null, resp);
+            cb!(err, null, resp);
             return;
           }
 
           const availablePermissions = arrify(resp.permissions);
 
-          const permissionsHash = permissions.reduce((acc, permission) => {
```
kinwa91: This was a bug.. permissions could've been `string|string[]`.

PR #465, opened 2018-10-18
Minor docs correction

ellord:
The value in the example five lines up is 'world', so this line should match that.

- [x] Tests and linter pass


ellord:
I signed it!

PR #469, opened 2018-10-18
Use Object.assign instead of "extend" module

### LINKED ISSUE #394: Use EcmaScript Object.assign instead of extend
If possible, we should use `Object.assign` instead of relying in `extend`. 

So, instead of
```JavaScript
options = extend({metadata: {}}, options);
```
let's use standard ES6:
```JS
options = Object.assign({metadata: {}}, options);
```

zbjornson:
Fixes #394

From that issue:

> extend does a deep clone by default (I believe)

It's shallow by default; it's only deep if the first argument is `true`, so it's safe and easy to see where the replacement needs to be deep. None of the existing usages appear to *need* to be deep. Nonetheless I did this in two commits so it's easier to review.

**Edit** the fun stuff going on in the tests breaks if I use `Object.assign()` there, so for now this is at least just a devDependency instead of a prod dependency.

- [x] Tests and linter pass


JustinBeckwith:
Thanks for the nudge here!  I went ahead and submitted #510, which should get us much, much closer to where you wanna go :)

PR #548, opened 2018-12-10
fix: `expires` can be a Date, string, or number

Crazometer:
Brought types inline with documentation for `getSignedUrl` and `getSignedPolicy`

https://cloud.google.com/nodejs/docs/reference/storage/2.3.x/File#getSignedUrl

The documentation states `options.expires` can be `any` type.

Internally `new Date()` accepts `string | number | Date` so I have used that.

https://github.com/Microsoft/TypeScript/blob/c2898db9dd58c9b0681d6131ed6076d71d89e6b5/lib/lib.es2015.core.d.ts#L90

You may want to update the web documentation from `any` to `string | number | Date` but I'm not sure of your conventions for that.

- [ ] Tests and linter pass


src/file.ts
```diff
@@ -55,7 +55,7 @@ export interface GetSignedUrlConfig {
   cname?: string;
   contentMd5?: string;
   contentType?: string;
-  expires: number;
+  expires: string | number | Date;
```
Crazometer:
```suggestion
  expires: string|number|Date;
```

src/file.ts
```diff
@@ -112,7 +112,7 @@ export interface GetSignedPolicyCallback {
 
 export interface GetSignedPolicyOptions {
   equals?: string[]|string[][];
-  expires: Date;
+  expires: string | number | Date;
```
Crazometer:
```suggestion
  expires: string|number|Date;
```

JustinBeckwith:
Thank you so much for the fix!

PR #557, opened 2018-12-14
Bucket Policy Only Samples

frankyn:
These are samples to support Bucket Policy Only Samples.

- [x] Tests and linter pass

I can't assign you @chestercun, PTAL. Thank you.


samples/buckets.js
```diff
@@ -49,7 +49,7 @@ async function createBucket(bucketName) {
 async function listBuckets() {
   // [START storage_list_buckets]
   // Imports the Google Cloud client library
-  const {Storage} = require('@google-cloud/storage');
+  const { Storage } = require('@google-cloud/storage');
```
JustinBeckwith: Looks like you need to run `npm run fix` :)

samples/system-test/buckets.test.js
```diff
@@ -86,6 +86,68 @@ it(`should set a bucket's default KMS key`, async () => {
...
+  );
+
+  assert.strictEqual(
+    (results.stdout + results.stderr).includes(
+      `Bucket Policy Only is enabled for ${bucketName}.`
+    ),
+    true
+  );
+  const [metadata] = await bucket.getMetadata();
+  assert.strictEqual(metadata.iamConfiguration.bucketPolicyOnly.enabled, true);
```
kinwa91: (style)
`assert.ok(metadata.iamConfiguration.bucketPolicyOnly.enabled);`

frankyn: Nice! I'll fix that, thanks @kinwa91!

kinwa91:
@frankyn looks like we've never return `Date` object for `datetime` strings in nodejs-storage.

@stephenplusplus have we ever discussed whether to convert `datetime` strings to `Date` object in handwritten libraries before?

stephenplusplus:
Sorry I missed this question. I think our rule has been to simplify a handwritten response format as best we can, and leave the API's raw response format as it came.

Considering a made-up method:

```js
file.getUsefulInfo((err, info, apiResponse) => {
  info.createdAt === Date('...')
  info.size === Number(8800)

  apiResponse.createdAt === 'long-string-of-numbers'
  apiResponse.size === '8800'
})
```

So, if a method exists and we can be more helpful by returning a handwritten response, we should be. For the `apiResponse`, I think we should leave it be, as opposed to iterating over it with best effort type conversion, as we might run into some unintentional side effects.

frankyn:
Thanks @stephenplusplus and @kinwa91! Let's keep as-is. I'll resolve conflicts and merge **after** the workaround in this PR is no longer needed, e.g. Tomorrow EOD (Wednesday).

frankyn:
@kinwa91 are you okay with I merge?

PR #564, opened 2018-12-22
fix: return the apiResponse in the right position

JustinBeckwith:
This looked to be pretty broken all over.  Really could use someone to look closely :)

### DIFF
--- a/src/acl.ts
+++ b/src/acl.ts
@@ -478,13 +478,12 @@ class Acl extends AclRoleAccessorMethods {
             role: options.role.toUpperCase(),
           },
         },
-        (err, resp) => {
+        (err, body, res) => {
           if (err) {
-            callback!(err, null, resp);
+            callback!(err, null, res);
             return;
           }
-
-          callback!(null, this.makeAclObject_(resp), resp);
+          callback!(null, this.makeAclObject_(body), res);
         });
   }
 
@@ -568,8 +567,8 @@ class Acl extends AclRoleAccessorMethods {
           uri: '/' + encodeURIComponent(options.entity),
           qs: query,
         },
-        (err, resp) => {
-          callback!(err, resp);
+        (err, body, res) => {
+          callback!(err, res);
         });
   }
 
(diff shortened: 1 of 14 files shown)


kinwa91:
👀

kinwa91:
I'm a bit confused, is this correct?

Looking at the `request` docs, the callback is in the form:
```
(error, response, body) => void
```

Are we transforming it somewhere that I'm not aware of?

JustinBeckwith:
Yeah :(   This line of code has been the source of much pain:
https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L305

That method takes the arguments `(Error, Response, Body)` and calls back with `(Error, Body, Response)`.  Take a look through some of the other changes I made here - in many cases, we were treating the `body` and the `response` as the same object.  It worked (ish?) because `Body` is of type `any`, but if anyone is actually looking at Response objects in the callback - I would guess they were wrong.

kinwa91:
The changes looks okay, but I'm not sure about how we should treat this when releasing what with the potential breaking changes to users who might rely on reading the `apiResponse` object in the callback.

JustinBeckwith:
It's a good point, and I don't know that this is clear cut.  See, the issue is that we *weren't* returning the apiResponse in any of these places.  We were handing the body as the last parameter of the callback, which means the previous behavior was very much a bug.  Based on the all of the jsdoc documentation, and hell even the examples - this was a bug before, and we're fixing it.  That having been said...  it is a breaking change.  *scratches head*.

@stephenplusplus @callmehiphop I think we need a few opinions from y'all here :)

JustinBeckwith:
*gentle ping* @stephenplusplus

stephenplusplus:
The `apiResponse` that we returned to users was always meant to be the JSON-parsed body of the message. We used the `response` argument that `request()` provides to [pluck out any errors](https://github.com/googleapis/nodejs-common/blob/a75b2f4b1f8f27b69ffe6cd8b3b874e877a6baa9/src/util.ts#L316), then we [pass all the arguments back](https://github.com/googleapis/nodejs-common/blob/a75b2f4b1f8f27b69ffe6cd8b3b874e877a6baa9/src/util.ts#L305), but from the API clients, we never really used that raw `response`.

JustinBeckwith:
After much discussion, this was hella bad :)

PR #582, opened 2019-01-17
Upgrade nodejs-common

stephenplusplus:
## Blocked On
- ~~https://github.com/googleapis/nodejs-common/pull/349~~
- ~~https://github.com/googleapis/gcs-resumable-upload/pull/167~~
- ~~https://github.com/googleapis/gcs-resumable-upload/pull/171~~
- ~~https://github.com/googleapis/nodejs-common/pull/353~~
- ~~https://github.com/googleapis/nodejs-common/pull/356~~

I took the code from @JustinBeckwith out of #556.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1036,14 +1036,14 @@ class Bucket extends ServiceObject {
 
     // The default behavior appends the previously-defined lifecycle rules with
     // the new ones just passed in by the user.
-    this.getMetadata((err: ApiError|null, metadata: MetadataResponse) => {
+    this.getMetadata((err: ApiError|null, metadata: Metadata) => {
       if (err) {
         callback!(err);
         return;
       }
 
       const currentLifecycleRules = arrify(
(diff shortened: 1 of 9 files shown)


src/bucket.ts
```diff
@@ -522,6 +522,13 @@ class Bucket extends ServiceObject {
     // Allow for "gs://"-style input, and strip any trailing slashes.
     name = name.replace(/^gs:\/\//, '').replace(/\/+$/, '');
 
+    const requestQueryObject: {userProject?: string} = {};
```
callmehiphop: Should we just declare an external interface for this? Not sure if it will ever be useful for the user or not.

stephenplusplus: While `serviceObject(config.methods[method].reqOpts` follows the request library's signature, I would still consider it private behavior. I wouldn't document it, since we might come up with something different in the future, and I'm not sure the user would understand exactly what it does / what side effects it has.

PR #595, opened 2019-02-05
test: skip public bucket system tests running under VPCSC

jkwlui:
When running under VPCSC, the process will have no egress access to
public buckets. Skipping these system tests when
GOOGLE_CLOUD_TESTS_IN_VPCSC=true environment variable is set.

This is the same variable used in the Python system-test:
https://github.com/googleapis/google-cloud-python/blob/master/storage/tests/system.py#L37

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -137,7 +137,7 @@ describe('storage', () => {
         bucket = storageWithoutAuth.bucket('gcp-public-data-landsat');
       });
 
-      it('should list and download a file', (done) => {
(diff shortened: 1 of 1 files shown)


callmehiphop (approved):
LGTM - side note I think you can programmatically skip tests in `before` hooks by calling `this.skip()` (it will skip the entire parent describe block);

```js
before(function() {
  if (RUNNING_IN_VPCSC) return this.skip();
});
```

I think this has the added bonus of showing that the tests were intentionally skipped as opposed to them just being absent.

PR #614, opened 2019-02-22
fix: getSigned(Policy|Url) throws if expiration is invalid Date

### LINKED ISSUE #613: The request signature we calculated does not match the signature you provided - Incorrect format for expirationDate
I'm receiving this message when using getSignedUrl.

The issue is that the returned URL has an empty expiration URL parameter and after debugging on my side I found this is caused because I was **passing a date in "DD-MM-YYYY" format** which i**s not accepted by new Date()**. 

When trying to do so, nodejs launches an "Invalid Date" error when passing a date in that format (e.g. "31-12-2019").

However, the API returns an URL and resolves the promise which clearly is not the expected behaviour. 

**The API should throw an error by rejecting the promise** with an according error message from the new Date invocation.

stephenplusplus:
Fixes #613 

If the user provides an invalid Date to `file.getSignedPolicy()` or `file.getSignedUrl()`, we will now throw.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2045,6 +2045,10 @@ class File extends ServiceObject<File> {
     const callback = args.callback;
     const expires = new Date((options as GetSignedPolicyOptions).expires);
 
+    if (isNaN(expires.getTime())) {
+      throw new Error('The expiration date provided was invalid.');
+    }
+
     if (expires.valueOf() < Date.now()) {
       throw new Error('An expiration date cannot be in the past.');
     }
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
@stephenplusplus I get that it's likely unrelated, but can you investigate the grpc-js system test failures?  It's likely a race condition with running the system tests twice at the same time.  Using this as a good opportunity to squash those bugs out.

buthig:
WOW!!! Hats off to you guys. This was really quick. I used to work for Google Customer Service a few years ago and the response time was significantly bigger than this.

What a change towards customers!!!

PR #615, opened 2019-02-26
fix: handle errors from file#createReadStream

### LINKED ISSUE #602: Uncaught exception in Zlib when reading part of compressed file via File.createReadStream
In my application, I'm trying to read part of a file in GCS — e.g., read the header from a large CSV. To do this, I create readStream with the `end` property specified — `file.createReadStream({ end: 1e6 })` — and then listen on the `data`, `error`, and `end` events. 

However, this causes Zlib to [throw an uncaught error](https://stackoverflow.com/questions/33926969/how-can-i-avoid-zlib-unexpected-end-of-file-when-gunzipping-partial-files), crashing the node process — even though I am listening on the `error` event of the readStream. Ideally, this error would either be handled internally or emit to the error event on the readStream.

#### Environment details
  - Node.js version: 10.14
  - `@google-cloud/storage` version: 2.4.2

#### Steps to reproduce

  1. Pipe `file.createReadStream({ end: # })` where `end < file.metadata.size`

Thank you!

stephenplusplus:
Fixes #602

#### To Dos

- [ ] Confirm it works
  - `npm install stephenplusplus/nodejs-storage#spp--602`
- [x] Tests

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1245,13 +1245,16 @@ class File extends ServiceObject<File> {
       // This is hooked to the `complete` event from the request stream. This is
       // our chance to validate the data and let the user know if anything went
       // wrong.
+      let onCompleteCalled = false;
       const onComplete = (err: Error|null) => {
         if (err) {
+          onCompleteCalled = true;
           throughStream.destroy(err);
           return;
         }
 
         if (rangeRequest) {
+          onCompleteCalled = true;
           throughStream.end();
           return;
         }
@@ -1261,6 +1264,12 @@ class File extends ServiceObject<File> {
           this.getMetadata({userProject: options.userProject}, onComplete);
           return;
         }
+
+        if (onCompleteCalled) {
+          return;
+        }
+
+        onCompleteCalled = true;
 
         const hashes = {
           crc32c: this.metadata.crc32c,
--- a/src/file.ts
+++ b/src/file.ts
@@ -1212,6 +1212,8 @@ class File extends ServiceObject<File> {
               return;
             }
 
+            rawResponseStream.on('error', onComplete);
+
             const headers = rawResponseStream.toJSON().headers;
             const isCompressed = headers['content-encoding'] === 'gzip';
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -1235,9 +1237,9 @@ class File extends ServiceObject<File> {
                   rawResponseStream.pipe(pumpify.obj(throughStreams));
             }
 
-            rawResponseStream.on('end', onComplete).pipe(throughStream, {
-              end: false
-            });
+            rawResponseStream.on('error', onComplete)
```
callmehiphop: Is having 2 error listeners with the same callback intentional?

stephenplusplus: The code above this actually re-assigns what `rawResponseStream` is. When we did that, it was to avoid duplicating the event handler assignments.

callmehiphop: Oh ok, sorry missed that part.

src/file.ts
```diff
@@ -1235,9 +1237,9 @@ class File extends ServiceObject<File> {
...
+                .on('end', onComplete)
```
callmehiphop: Are there any scenarios where `onComplete` could fire multiple times? I sort of wonder if we should be using something like `once` to ensure that doesn't happen somehow.

stephenplusplus: I 'spose anything is possible! I added new code for that, please take a look!

src/file.ts
```diff
@@ -1235,21 +1237,24 @@ class File extends ServiceObject<File> {
...
           };
 
       // This is hooked to the `complete` event from the request stream. This is
       // our chance to validate the data and let the user know if anything went
       // wrong.
+      let onCompleteCalled = false;
       const onComplete = (err: Error|null) => {
         if (err) {
+          onCompleteCalled = true;
           throughStream.destroy(err);
```
stephenplusplus: `throughStream.destroy()` seems to be once-ified automatically.

PR #633, opened 2019-03-15
docs: document destination option

bcoe:
It's often the case that a user wants to change the destination name of an object they're uploading, but it's pretty buried in the docs.

- [x] Tests and linter pass

### DIFF
--- a/samples/files.js
+++ b/samples/files.js
@@ -116,6 +116,8 @@ async function uploadFile(bucketName, filename) {
   await storage.bucket(bucketName).upload(filename, {
     // Support for HTTP requests made with `Accept-Encoding: gzip`
     gzip: true,
+    // By setting the option `destination`, you can change the name of the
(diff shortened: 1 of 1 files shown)


samples/files.js
```diff
@@ -116,6 +116,8 @@ async function uploadFile(bucketName, filename) {
   await storage.bucket(bucketName).upload(filename, {
     // Support for HTTP requests made with `Accept-Encoding: gzip`
     gzip: true,
+    // By setting the option `destination`, you can change the name of the
```
JustinBeckwith: It's a little weird for me to read the comment in this context.  I'd expect to see something like this above a `destination` property in the obj, which could be commented out, but there for illustration.

bcoe: so perhaps:

`// destination: destinationFilename (defaults to filename)`

PR #637, opened 2019-03-22
feat: introduce v4 signed url

jkwlui:
~~requires https://github.com/googleapis/google-auth-library-nodejs/pull/649 for system-tests to pass~~


src/file.ts
```diff
@@ -2281,25 +2285,76 @@ class File extends ServiceObject<File> {
       throw new Error('The expiration date provided was invalid.');
     }
 
-    if (expiresInMSeconds < Date.now()) {
+    if (expiresInMSeconds < this.getDate().valueOf()) {
```
JustinBeckwith: What's the difference?

src/file.ts
```diff
@@ -2281,25 +2285,76 @@ class File extends ServiceObject<File> {
...
 
-    const config: GetSignedUrlConfigInternal = Object.assign({}, cfg);
-
-    config.action = ({
+    const verb = ({
       read: 'GET',
       write: 'PUT',
       delete: 'DELETE',
       resumable: 'POST',
-    } as {[index: string]: string})[config.action];
```
JustinBeckwith: I think I saw a conformance test looking for this type somewhere in another PR lol

src/file.ts
```diff
@@ -2281,25 +2285,76 @@ class File extends ServiceObject<File> {
...
+    } as {[index: string]: string})[cfg.action];
 
     const name = encodeURIComponent(this.name);
-    config.resource = '/' + this.bucket.name + '/' + name;
+    const resource = '/' + this.bucket.name + '/' + name;
```
JustinBeckwith: string interpolation please

src/util.ts
```diff
@@ -22,3 +22,24 @@ export function normalize<T = {}, U = Function>(
...
+             .replace(/[:-]/g, '')  // strip :'s and -'s
+             .split('.')[0] +
+      'Z';  // remove milliseconds part
+}
+
+/**
+ * Flatten an object into an Array of arrays, [[key, value], ..].
+ * @internal
+ */
+export function flattenObject<T>(obj: {[key: string]: T}): Array<[string, T]> {
```
JustinBeckwith: isn't this pretty much the same as `Object.entries`?

system-test/storage.ts
```diff
@@ -2809,6 +2809,7 @@ describe('storage', () => {
     it('should create a signed read url', done => {
       file.getSignedUrl(
           {
+            version: 'v2',
```
JustinBeckwith: Is `v2` still the default if none is passed?

test/file.ts
```diff
@@ -2496,14 +2503,16 @@ describe('File', () => {
   describe('getSignedUrl', () => {
     const CONFIG = {
       action: 'read',
-      expires: Date.now() + 2000,
-    };
+      expires: Date.now() + 2000,  // now + 2 seconds
+    } as {action: string, expires: number, version: string};
```
JustinBeckwith: Should be a better interface for us to use here somewhere :)

test/file.ts
```diff
@@ -2532,79 +2530,236 @@ describe('File', () => {
...
-    it('should return an error if signBlob errors', done => {
-      const error = new Error('Error.');
+        const config = Object.assign({}, CONFIG, {
+          responseType: type,
+        });
 
-      BUCKET.storage.authClient.sign = () => {
-        return Promise.reject(error);
-      };
+        directoryFile.getSignedUrl(config, (_: Error, signedUrl: string) => {
```
JustinBeckwith: Why do you need the types on the callback?  If the types on the method are good, you should be able to infer this.

test/file.ts
```diff
@@ -2532,79 +2530,236 @@ describe('File', () => {
...
 
-      file.getSignedUrl(CONFIG, (err: Error) => {
-        assert.strictEqual(err.name, 'SigningError');
-        assert.strictEqual(err.message, error.message);
-        done();
+      it('should add generation parameter', done => {
+        const generation = 10003320000;
+        const file = new File(BUCKET, 'name', {generation});
+
+        file.getSignedUrl(CONFIG, (err: Error, signedUrl: string) => {
```
JustinBeckwith: Drop the types on the callback if possible

src/util.ts
jkwlui: `Object.entries` is ES6+ as I've learned.. Can I use `lodash.entries` here?

test/file.ts
jkwlui: fixed!

system-test/storage.ts
jkwlui: Yes, with the expectation `v4` will be made the default in the future.

src/file.ts
```diff
@@ -2319,51 +2374,110 @@ class File extends ServiceObject<File> {
...
+
+          const blobToSign = [
+            'GOOG4-RSA-SHA256',
+            dateToISOString(now),
+            credentialScope,
+            canonicalRequestHash,
+          ].join('\n');
+
+          return this.storage.authClient.sign(blobToSign).catch((err) => {
+            throw new SigningError(err.message);
```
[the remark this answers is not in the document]
jkwlui: fixed

src/util.ts
JustinBeckwith: aghghghg - use whatever you like, then file an issue in the repo to clean it up once node 6 is EOL

jkwlui: okay. I'll leave it as-is and add the issue after this is merged.

src/file.ts
```diff
@@ -287,12 +290,7 @@ interface FileQuery {
 }
 
 interface SignedUrlQuery {
-  GoogleAccessId: string;
-  Expires: number;
-  Signature: string;
-  generation: number;
-  'response-content-type': string;
-  'response-content-disposition': string;
+  [key: string]: string;
```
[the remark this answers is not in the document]
jkwlui: We will be able to support additional query param in the future - one examples is `subresource`

src/file.ts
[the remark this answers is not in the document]
jkwlui: This is a good point. How should we go about doing this if we want it to throw (rejects with) `SigningError` while preserving. I couldn't find an existing pattern that I can look to. @bcoe @JustinBeckwith

src/file.ts
```diff
@@ -3013,6 +3127,47 @@ class File extends ServiceObject<File> {
...
+    return new Date();
+  }
+
+  /**
+   * Create canonical headers for signing v4 url.
+   * @param headers
+   * @private
+   */
+  private getCanonicalHeaders(headers: http.OutgoingHttpHeaders) {
+    // Sort headers by their lowercased names
```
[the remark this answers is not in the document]
jkwlui: Per v4 spec, canonical headers are sorted lexically after lowercasing the header names.

src/file.ts
```diff
@@ -3013,6 +3127,47 @@ class File extends ServiceObject<File> {
...
+            .map<[string, string | string[] | number | undefined]>(
+                ([headerName, value]) => [headerName.toLowerCase(), value],
+                )
+            .sort((a, b) => a[0].localeCompare(b[0]));
+
+    return sortedHeaders
+        .map(([headerName, value]) => {
+          if (!value) return;
+
+          const values = Array.isArray(value) ? value : [value.toString(10)];
```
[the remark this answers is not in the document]
jkwlui: headers can potentially have single/multiple values:

for example:
`My-Header: value1` and `My-Header: value1, value2` are both valid.

User supply either a `string` (single-valued) or an `array` (multi-valued), so we convert the string to an `Array<string>` here, so we can perform the trimming/stripping (below) using non-duplicate code.

src/file.ts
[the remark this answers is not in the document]
JustinBeckwith: In the future, could we then add `subresource`?  I really want to avoid open ended property bags unless it truly accepts random key value pairs.

src/file.ts
```diff
@@ -2288,18 +2293,69 @@ class File extends ServiceObject<File> {
...
-    const config: GetSignedUrlConfigInternal = Object.assign({}, cfg);
-
-    config.action = ({
+    const verb = ({
       read: 'GET',
       write: 'PUT',
       delete: 'DELETE',
       resumable: 'POST',
-    } as {[index: string]: string})[config.action];
+    } as {[index: string]: string})[cfg.action];
```
JustinBeckwith: this cast feels weird - is there an interface for what you're doing?

src/file.ts
```diff
@@ -2288,18 +2293,69 @@ class File extends ServiceObject<File> {
...
+          if (typeof config.responseDisposition === 'string') {
+            query['response-content-disposition'] = config.responseDisposition!;
+          }
+
+          if (this.generation) {
+            query.generation = this.generation.toString();
+          }
+
+          const parsedHost =
+              url.parse(config.cname || STORAGE_DOWNLOAD_BASE_URL);
```
JustinBeckwith: I believe `url.parse` is actually deprecated:
https://nodejs.org/docs/latest/api/url.html#url_legacy_url_api

src/file.ts
```diff
@@ -2319,51 +2375,110 @@ class File extends ServiceObject<File> {
...
+        .then(
+            signature => authClient.getCredentials().then(
+                credentials => ({
+                  GoogleAccessId: credentials.client_email!,
+                  Expires: config.expiration.toString(),
+                  Signature: signature,
+                })),
+            )
+        .catch((err) => {
+          throw new SigningError(err.message);
```
JustinBeckwith: I am worried about the loss of stack trace here.

src/file.ts
```diff
@@ -2319,51 +2375,110 @@ class File extends ServiceObject<File> {
...
-                  'attachment; filename="' + config.promptSaveAs + '"';
-            }
-            if (typeof config.responseDisposition === 'string') {
-              query['response-content-disposition'] =
-                  config.responseDisposition!;
-            }
+  private getSignedUrlV4(config: GetSignedUrlConfigInternal):
+      Promise<SignedUrlQuery> {
+    const now = new Date();
+    const nowInSeconds = Math.floor(now.valueOf() / 1000);
```
JustinBeckwith: Can we prefer `Math.floor(Date.now() * 1000)`?

src/file.ts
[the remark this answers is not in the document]
JustinBeckwith: That's a really great question 😆  Curious on @ofrobots thoughts here as well.

src/file.ts
jkwlui: fixed!

src/file.ts
```diff
@@ -3013,6 +3128,38 @@ class File extends ServiceObject<File> {
...
+  /**
+   * Create canonical headers for signing v4 url.
+   * @param headers
+   * @private
+   */
+  private getCanonicalHeaders(headers: http.OutgoingHttpHeaders) {
+    // Sort headers by their lowercased names
+    const sortedHeaders =
+        flattenObject(headers)
+            .map<[string, string | string[] | number | undefined]>(
```
JustinBeckwith: needs more generic types

src/file.ts
```diff
@@ -2319,51 +2374,110 @@ class File extends ServiceObject<File> {
...
+      extensionHeaders['x-goog-resumable'] = 'start';
+    }
+    if (config.contentMd5) {
+      extensionHeaders['content-md5'] = config.contentMd5;
+    }
+    if (config.contentType) {
+      extensionHeaders['content-type'] = config.contentType;
+    }
+
+    let signedHeaders = new Array<string>();  // should have host at least
```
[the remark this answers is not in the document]
jkwlui: fixed!

src/util.ts
```diff
@@ -22,3 +22,11 @@ export function normalize<T = {}, U = Function>(
...
   return {options, callback};
 }
+
+/**
+ * Flatten an object into an Array of arrays, [[key, value], ..].
+ * @internal
+ */
+export function flattenObject<T>(obj: {[key: string]: T}): Array<[string, T]> {
+  return Object.keys(obj).map((key) => [key, obj[key]] as [string, T]);
+}
```
JustinBeckwith: please add a newline

src/file.ts
```diff
@@ -2319,51 +2374,110 @@ class File extends ServiceObject<File> {
...
+    signedHeaders = Object.keys(extensionHeaders);
+    signedHeaders = signedHeaders.map(header => header.toLowerCase());
+    signedHeaders.sort();
+
+    const extensionHeadersString = this.getCanonicalHeaders(extensionHeaders);
+
+    const datestamp = dateToISOString(now).split('T')[0];
```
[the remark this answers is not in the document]
jkwlui: fixed with `dateformat` library

system-test/storage.ts
```diff
@@ -2844,6 +2846,59 @@ describe('storage', () => {
...
+    before(done => {
+      file = bucket.file('LogoToSign.jpg');
+      fs.createReadStream(FILES.logo.path)
+          .pipe(file.createWriteStream())
+          .on('error', done)
+          .on('finish', done.bind(null, null));
+    });
+
+    it('should create a signed read url', done => {
+      file.getSignedUrl(
```
JustinBeckwith: for new tests please use `async/await`

src/file.ts
```diff
@@ -2319,51 +2374,110 @@ class File extends ServiceObject<File> {
...
+          queryParams = {
+            'X-Goog-Algorithm': 'GOOG4-RSA-SHA256',
+            'X-Goog-Credential': credential,
+            'X-Goog-Date': dateToISOString(now),
+            'X-Goog-Expires': expiresPeriodInSeconds.toString(),
+            'X-Goog-SignedHeaders': signedHeaders.join(';'),
+          };
 
-            callback!(null, signedUrl);
+          const canonicalQueryParams = [];
```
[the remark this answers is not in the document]
jkwlui: fixed!

test/file.ts
```diff
@@ -2512,99 +2526,246 @@ describe('File', () => {
...
+    afterEach(() => {
+      timekeeper.reset();
+    });
 
+    it('should default to v2 if version is not given', done => {
       file.getSignedUrl(CONFIG, (err: Error, signedUrl: string) => {
         assert.ifError(err);
         assert.strictEqual(typeof signedUrl, 'string');
-        const expires = Math.round(CONFIG.expires / 1000);
+        const expires = Math.round((CONFIG.expires as number) / 1000);
```
JustinBeckwith: Prefer `Number(CONFIG.expires)`

test/file.ts
jkwlui: `let directoryFile: any`, probably for easier monkey-patching methods :(

We should really refactor these "stubs" with `sinon`

test/file.ts
jkwlui: same as above :(

src/file.ts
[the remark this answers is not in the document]
ofrobots: I'm not aware of the motivation behind using a custom error class; so I will refrain from commenting on that.

1. Can you copy over the `err.stack` property?
1. Is it possible for `authClient.sign` to throw the right 'type' of exception in the first place?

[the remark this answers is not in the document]
jkwlui: (I was refactoring existing code so I'm just speculating here) I believe a `SigningError` is used to differentiate errors that arise because of some kind of missing/problems with signing, e.g. no email in credentials, rather than some issue with the code (which would be an `Error`). 

I think I'll go with 2. here changing the `sign` method, although I'm still curious if there's a way to correctly do this. Copying over `err.stack` feels too hacky for my taste.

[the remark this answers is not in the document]
bcoe: @jkwlui @ofrobots I've definitely used this approach before, bubbling a custom error type, as a way of offering a more contextual handling of certain types of errors; as an example in yargs, it allows us to print known error types as pretty text to the user, while still bubbling application level errors.

I believe that instead of copying `err.stack` there's a method available for subclasses of Error: `captureStackTrace`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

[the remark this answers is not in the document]
ofrobots: @bcoe the idea is to _not_ capture the stack trace, because we want the context to remain the same as the original.

@jkwlui I don't necessarily consider copying over the stack to be 'hacky'.

src/file.ts
jkwlui: would like to be able to re-use the same Date object, hence the same datetime throughout the method.

src/file.ts
```diff
@@ -2179,6 +2182,8 @@ class File extends ServiceObject<File> {
    * @param {object} config Configuration object.
    * @param {string} config.action "read" (HTTP: GET), "write" (HTTP: PUT), or
    *     "delete" (HTTP: DELETE), "resumable" (HTTP: POST).
+   * @param {string} [config.version] The signing version to use. Currently support 'v2' and 'v4' signing.
```
stephenplusplus: For noting the default version, you can do:

```js
@param {string} [config.version='v2']
```

Also, I believe we're sticking to 80 char line length limits.

Here's a small rewrite:

```js
     * @param {string} [config.version='v2'] The signing version to use, either
     *     'v2' or 'v4'.
```

src/file.ts
jkwlui: Switched to using the `query-string` package.

The built-in `URLSearchParams` incorrectly encodes `;` character in `response-content-disposition` header as separate query params.

src/file.ts
jkwlui: Added an Enum type.

src/file.ts
jkwlui: Thanks Stephen

src/file.ts
```diff
@@ -33,6 +34,7 @@ import {Duplex, Writable, Readable} from 'stream';
 import * as streamEvents from 'stream-events';
 import * as through from 'through2';
 import * as xdgBasedir from 'xdg-basedir';
+import * as qs from 'query-string';
```
JustinBeckwith: Is there a reason `querystring.stringify` isn't good enough?
https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

If there isn't - we should document the why :)

package.json
```diff
@@ -109,10 +114,11 @@
     "power-assert": "^1.4.4",
     "prettier": "^1.7.0",
     "proxyquire": "^2.0.0",
+    "query-string": "^6.4.0",
```
JustinBeckwith: This is used in `file.ts`, but looks to only be added as a dev dependency.  Also - I wonder out loud if we really need a third party module for this :)

src/file.ts
```diff
@@ -287,12 +299,18 @@ interface FileQuery {
 }
 
 interface SignedUrlQuery {
-  GoogleAccessId: string;
-  Expires: number;
-  Signature: string;
-  generation: number;
-  'response-content-type': string;
-  'response-content-disposition': string;
+  GoogleAccessId?: string;
```
JustinBeckwith: Are all of these really optional?

src/file.ts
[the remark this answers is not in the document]
JustinBeckwith: Awesome!  Can you leave a comment saying that?

src/file.ts
jkwlui: I forgot this build-in module exists 😅

jkwlui: refactored to use `querystring`. Thanks for catching that!

src/file.ts
jkwlui: Good question!

v2 requires
```
  GoogleAccessId?: string;
  Expires?: number;
  Signature?: string;
```

while v4 requires
```
  'X-Goog-Algorithm'?: string;
  'X-Goog-Credential'?: string;
  'X-Goog-Date'?: string;
  'X-Goog-Expires'?: number;
  'X-Goog-SignedHeaders'?: string;
  'X-Goog-Signature'?: string;
```

jkwlui: Refactored `SignedUrlQuery` to contain the shared query params.

Create interfaces `{V4|V2}SignedUrlQuery` to contain params specific to each version.

src/file.ts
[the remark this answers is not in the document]
jkwlui: Talked to @frankyn about this.

Turns out in v2 we've been converting array header values using `${headerValue}` which turns the array into a string, delimited by `,` with no spaces. It is actually a valid construction of a multi-valued header which the GCS server accepts.

Refactored the v4 code to just convert the array to string.

src/file.ts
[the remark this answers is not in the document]
jkwlui: Yup!

test/file.ts
jkwlui: I think we agreed it's better not to try to refactor too much in a feature PR - fixing this with `sinon` would be quite a big change given the size of this test file.

test/file.ts
```diff
@@ -2617,6 +2617,21 @@ describe('File', () => {
         );
       });
 
+      it('should set correct settings if resumable', done => {
```
JustinBeckwith: Please use `async` for new tests :)

test/file.ts
```diff
@@ -2617,6 +2617,21 @@ describe('File', () => {
...
+        const config = Object.assign({}, CONFIG, {
+          action: 'resumable',
+        });
+
+        const spy = sinon.spy(file, 'getCanonicalHeaders');
+
+        file.getSignedUrl(config, (err: Error) => {
+          assert.ifError(err);
+          assert(spy.returnValues[0].indexOf('x-goog-resumable:start') > -1);
```
JustinBeckwith: use `.includes`

test/file.ts
jkwlui: the unit tests mock out `promisify`.... I'll try!

JustinBeckwith: ah, no big deal if it's a pain, just a preference :)

JustinBeckwith:
@jkwlui thanks for hanging with me on the feedback on this one 🙃

jkwlui:
Not at all LOL I appreciate the feedback!

holding off merging until @frankyn had a chance to take a look :)

src/file.ts
```diff
@@ -337,6 +364,13 @@ class RequestError extends Error {
   errors?: Error[];
 }
 
+type ValueOf<T> = T[keyof T];
+type HeaderValue = ValueOf<http.OutgoingHttpHeaders>;
+
+const SEVEN_DAYS = 7 * 24 * 60 * 60;
```
frankyn: Could you replace the value with expected result?
604800

src/file.ts
```diff
@@ -337,6 +364,13 @@ class RequestError extends Error {
...
+
+const DEFAULT_SIGNING_VERSION = 'v2';
```
frankyn: Add a note that this will change at a later time to V4.

src/file.ts
```diff
@@ -2180,6 +2214,8 @@ class File extends ServiceObject<File> {
    * @param {object} config Configuration object.
    * @param {string} config.action "read" (HTTP: GET), "write" (HTTP: PUT), or
    *     "delete" (HTTP: DELETE), "resumable" (HTTP: POST).
+   * @param {string} [config.version='v2'] The signing version to use, either
+   *     'v2' or 'v4.
```
frankyn: Add a warning when enabling v2 that the default will change in the future:
 
> You have generated a signed URL using the default v2 signing implementation. In the future, this will default to v4. You may experience breaking changes if you use longer than 7 day expiration times with v4. To opt-in to the behavior specify `config.version='v4'`

src/file.ts
```diff
@@ -2294,21 +2330,61 @@ class File extends ServiceObject<File> {
...
+
+    const config: GetSignedUrlConfigInternal = Object.assign({}, cfg, {
+      method,
+      expiration: expiresInSeconds,
+      resource,
+      name,
+    });
+
+    let promise: Promise<SignedUrlQuery>;
+    if (version === 'v2') {
```
frankyn: Could you log the future change:
> You have generated a signed URL using the default v2 signing implementation. In the future, this will default to v4. You may experience breaking changes if you use longer than 7 day expiration times with v4. To opt-in to the behavior specify config.version='v4'

src/file.ts
```diff
@@ -2180,6 +2214,8 @@ class File extends ServiceObject<File> {
    * @param {object} config Configuration object.
    * @param {string} config.action "read" (HTTP: GET), "write" (HTTP: PUT), or
    *     "delete" (HTTP: DELETE), "resumable" (HTTP: POST).
+   * @param {string} [config.version='v2'] The signing version to use, either
```
frankyn: Add a note that when using resumable (POST). The signature includes `x-goog-resumable:start`.

src/file.ts
```diff
@@ -2294,21 +2330,61 @@ class File extends ServiceObject<File> {
...
-      read: 'GET',
-      write: 'PUT',
-      delete: 'DELETE',
-      resumable: 'POST',
-    } as {[index: string]: string})[config.action];
+    const method = ActionToHTTPMethod[cfg.action];
 
     const name = encodeURIComponent(this.name);
-    config.resource = '/' + this.bucket.name + '/' + name;
+    const resource = `/${this.bucket.name}/${name}`;
```
frankyn: The new conformance tests will expect the trailing slash to be removed when file name is not given. (List objects request).

src/file.ts
```diff
@@ -337,6 +364,13 @@ class RequestError extends Error {
...
+type ValueOf<T> = T[keyof T];
+type HeaderValue = ValueOf<http.OutgoingHttpHeaders>;
+
+const SEVEN_DAYS = 7 * 24 * 60 * 60;
+
+const DEFAULT_SIGNING_VERSION = 'v2';
+
 /**
  * A File object is created from your {@link Bucket} object using
  * {@link Bucket#file}.
```
frankyn: Is it possible to create a signed URL to list objects in a bucket e.g, GET request without an object name? Is that something to do in a follow-up PR?

src/file.ts
jkwlui: We don't really do these in Node. 

@JustinBeckwith do we have any best-practices/alternatives for warning users of impending change in API during runtime in addition to jsdoc notes?

src/file.ts
jkwlui: I believe we've discussed this and concluded that this is out of scope for this PR. Adding bucket level signed URLs will be a future endeavour.

src/file.ts
jkwlui: List objects would be a bucket level signed URL - see below :)

src/file.ts
JustinBeckwith: Yes we do!  Checkout what I did here for google-auth-library:
https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/messages.ts

You want to use `process.emitWarning`, but there is a lot of nuance.  Following gal is probably a good idea.

jkwlui: gal?

JustinBeckwith: Sorry, google-auth-library

jkwlui: @frankyn Do we only emit warning if the user didn't provide the `version` config (and defaults to 'v2'), and not when the user specified specifically 'v2'?

frankyn: Only default is fine in this case.

jkwlui: > Yes we do!

Heh I must've remembered a different conversion about emitting warnings in Node. 😛 
I am trying to add a similar emitWarning logic here and I feel like I'm duplicating most of the code you have on `google-auth-library`. Any hints here, or should I just go ahead with it?

src/file.ts
jkwlui: We've decided to tackle Bucket level SignedURL in a future PR.

src/file.ts
JustinBeckwith: It's entirely possible there's a library to do this - not sure!  @googleapis/node-team?

jkwlui: @JustinBeckwith are you okay with me keeping the (mostly duplicated) code? I've looked high and low and couldn't find anything :(

JustinBeckwith: I am ok with it

JustinBeckwith:
@jkwlui @frankyn this is out of SLO - can we get it landed?

frankyn:
@jkwlui please merge and release when issues are resolved.

PR #645, opened 2019-03-26
docs: new synthtool generated README

bcoe:
refs: https://github.com/googleapis/google-cloud-node/issues/2868 this PR regenerates our README using `synthtool` rather than node-repo-tools.

It seems to be most of the way there, one thing that does jump out at me is the description isn't nearly as fleshed out ... I wonder if this grew organically over time; any thoughts about how to auto-generate the README with up-to-date samples, while maintaining a more useful description?

@busunkim96 thanks for all your help on this!

- [x] Tests and linter pass

### DIFF
--- a/.readme-partials.yaml
+++ b/.readme-partials.yaml
@@ -1,0 +1,8 @@
+introduction: |-
+    > Node.js idiomatic client for [Cloud Storage][product-docs].
+
+    [Cloud Storage](https://cloud.google.com/storage/docs) allows world-wide
+    storage and retrieval of any amount of data at any time. You can use Google
+    Cloud Storage for a range of scenarios including serving website content,
+    storing data for archival and disaster recovery, or distributing large data
+    objects to users via direct download.
--- a/.readme-partials.yaml
+++ b/.readme-partials.yaml
@@ -6,3 +6,9 @@ introduction: |-
     Cloud Storage for a range of scenarios including serving website content,
     storing data for archival and disaster recovery, or distributing large data
     objects to users via direct download.
+quickstart_footer:
+    // process.env.GOOGLE_APPLICATION_CREDENTIALS points to service account JSON.
+
+    // process.env.GOOGLE_CLOUD_PROJECT points to your project in GCP.
+
+    quickstart(process.env.GOOGLE_CLOUD_PROJECT, "my-bucket")
--- a/.readme-partials.yaml
+++ b/.readme-partials.yaml
@@ -11,4 +11,4 @@ quickstart_footer:
 
     // process.env.GOOGLE_CLOUD_PROJECT points to your project in GCP.
 
-    quickstart(process.env.GOOGLE_CLOUD_PROJECT, "my-bucket")
+    quickstart(process.env.GOOGLE_CLOUD_PROJECT)
--- a/.readme-partials.yaml
+++ b/.readme-partials.yaml
@@ -6,9 +6,3 @@ introduction: |-
     Cloud Storage for a range of scenarios including serving website content,
     storing data for archival and disaster recovery, or distributing large data
     objects to users via direct download.
-quickstart_footer:
-    // process.env.GOOGLE_APPLICATION_CREDENTIALS points to service account JSON.
-
-    // process.env.GOOGLE_CLOUD_PROJECT points to your project in GCP.
-
-    quickstart(process.env.GOOGLE_CLOUD_PROJECT)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,93 +1,55 @@
 [//]: # "This README.md file is auto-generated, all changes to this file will be lost."
-[//]: # "To regenerate it, use `npm run generate-scaffolding`."
+[//]: # "To regenerate it, use `python -m synthtool`."
```
JustinBeckwith: I think it's supposed to just be `synthtool`, and it yells at me when I try to run it like that 😆

README.md
```diff
@@ -1,93 +1,55 @@
...
 
-[![release level](https://img.shields.io/badge/release%20level-general%20availability%20%28GA%29-brightgreen.svg?style&#x3D;flat)](https://cloud.google.com/terms/launch-stages)
+[![release level](https://img.shields.io/badge/release%20level-general%20availability%20%28GA%29-brightgreen.svg?style=flat)](https://cloud.google.com/terms/launch-stages)
 [![npm version](https://img.shields.io/npm/v/@google-cloud/storage.svg)](https://www.npmjs.org/package/@google-cloud/storage)
 [![codecov](https://img.shields.io/codecov/c/github/googleapis/nodejs-storage/master.svg?style=flat)](https://codecov.io/gh/googleapis/nodejs-storage)
 
-> Node.js idiomatic client for [Cloud Storage][product-docs].
+Cloud Storage Client Library for Node.js
 
-[Cloud Storage](https://cloud.google.com/storage/docs) allows world-wide storage and retrieval of any amount of data at any time. You can use Google Cloud Storage for a range of scenarios including serving website content, storing data for archival and disaster recovery, or distributing large data objects to users via direct download.
```
JustinBeckwith: Yeah, it would be really nice to hang onto this, I think.  Not sure if putting the long description in the metadata or embedding it in a `README.md.tpl` file is the right path.

README.md
```diff
@@ -1,93 +1,55 @@
...
-
-
-* [Cloud Storage Node.js Client API Reference][client-docs]
-* [github.com/googleapis/nodejs-storage](https://github.com/googleapis/nodejs-storage)
-* [Cloud Storage Documentation][product-docs]
-
-Read more about the client libraries for Cloud APIs, including the older
```
JustinBeckwith: What happened to these 3 lines?

README.md
busunkim96: It *should* work either way. :thinking: `python -m synthtool` is what the synthtool README says.

callmehiphop:
Something that has always bugged me about the README's is that the link for the client API docs is not really in a convenient spot. IMO it should be available right at the start and I shouldn't need to scan the document for it.

README.md
JustinBeckwith: ahhh, I could be thinking of another tool - apologies for the confusion!

README.md
bcoe: I propose the late-game addition of `readme-partials.yml`, with a section `introduction`. This would give us the flexibility to gradually hand-write some portions of the API documentation, which I think we're going to inevitably want for our libraries.

README.md
bcoe: It looks like at some point we just made the decision to move this content up CC: @callmehiphop, I'm fine with shifting this around.

README.md
busunkim96: In this case it looks like it's copied from https://cloud.google.com/storage/docs/ verbatim.

JustinBeckwith: @bcoe I really think this is a *great* idea.  We had a few issues come up in the past asking for this, even with nodejs-repo-tools.

.readme-partials.yaml
```diff
@@ -0,0 +1,14 @@
+introduction: |-
+    > Node.js idiomatic client for [Cloud Storage][product-docs].
+
+    [Cloud Storage](https://cloud.google.com/storage/docs) allows world-wide
+    storage and retrieval of any amount of data at any time. You can use Google
+    Cloud Storage for a range of scenarios including serving website content,
+    storing data for archival and disaster recovery, or distributing large data
+    objects to users via direct download.
+quickstart_footer:
```
bcoe: @JustinBeckwith @callmehiphop it's now possible to customize the README output with `.readme-partials.yaml`. Right now the supported keys are `introduction` (for a handwritten library intro), and `quickstart_footer`, allowing us to provide some shim code that actually makes the quickstart functional.

JustinBeckwith: With #647 landing, can we drop this?

PR #647, opened 2019-03-27
docs(samples): refactor the quickstart to match the new rubric

JustinBeckwith:
This PR updates the quickstart sample to follow the new Node.js sample rubric:
- The projectId isn't explicitly used, so it's reference was dropped
- The region tag was moved inside of the `main` function
- Added a note asking the developer uncomment the bucketName
- Used an inner async function with the actual code

@bcoe this will fix the issue you had where you needed to add the footer thingy to synthtool. 
@jmdobry is the rubric published anywhere?

### DIFF
--- a/samples/quickstart.js
+++ b/samples/quickstart.js
@@ -15,22 +15,29 @@
 
 'use strict';
 
-// [START storage_quickstart]
-async function quickstart(
-  projectId = 'YOUR_PROJECT_ID', // Your Google Cloud Platform project ID
+function main(
   bucketName = 'my-new-bucket' // The name for the new bucket
 ) {
+  // [START storage_quickstart]
   // Imports the Google Cloud client library
   const {Storage} = require('@google-cloud/storage');
 
   // Creates a client
-  const storage = new Storage({projectId});
+  const storage = new Storage();
 
-  // Creates the new bucket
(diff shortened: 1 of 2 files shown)


samples/quickstart.js
```diff
@@ -15,22 +15,29 @@
...
 
-  // Creates the new bucket
-  await storage.createBucket(bucketName);
-  console.log(`Bucket ${bucketName} created.`);
+  /**
+   * TODO(developer): Uncomment these variables before running the sample.
+   */
+  // const bucketName = 'bucket-name';
+
+  async function createBucket() {
```
bcoe: I think it might be worth adding a comment (similar to bucketName) that calls out the magic `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

---

longer term, I'd love to think a bit more about the default approach we ask folks to configure their Node.js library -- it's been my experience that it would be more common to pass in a JSON configuration object, rather than pointing to one on disk, for the types of applications folks would be written in Node.js.

I believe we support this, but it's not called out predominantly (no action items for this pull request, just a thought that's been on my mind).

JustinBeckwith: This is a hella complicated topic with lots of history :) Let's chat tomorrow about it!

PR #649, opened 2019-03-27
docs: regenerate the samples/README.md

bcoe:
Fixes: https://github.com/googleapis/google-cloud-node/issues/2868 we can now magically auto-generate both README.md and samples/README.md (this will become all the more powerful with discussions around auto-generating `.repo-metadata.json` (CC: @busunkim96, @jmdobry).

This pull request:

* switches to generated `samples/README.md` (this file is a slight simplification, no longer rendering the yargs CLI output, which is being removed from newer samples any ways).
* removes `generate-scaffolding` from the `package.json` (as it's now not used by anything).

Related: https://github.com/googleapis/synthtool/pull/214

- [x] Tests and linter pass


samples/README.md
```diff
@@ -2,324 +2,215 @@
 [//]: # "To regenerate it, use `npm run generate-scaffolding`."
 <img src="https://avatars2.githubusercontent.com/u/2810941?v=3&s=96" alt="Google Cloud Platform logo" title="Google Cloud Platform" align="right" height="96" width="96"/>
 
-# Google Cloud Storage: Node.js Samples
+# [Google Cloud Storage: Node.js Samples](https://github.com/googleapis/nodejs-storage)
 
 [![Open in Cloud Shell][shell_img]][shell_link]
```
JustinBeckwith: So I should have asked this elsewhere - why would I want to open the README in cloud shell?  I kinda think maybe that doesn't belong there

image

samples/README.md
```diff
@@ -2,324 +2,215 @@
...
 
-[Cloud Storage](https://cloud.google.com/storage/docs) allows world-wide storage and retrieval of any amount of data at any time. You can use Google Cloud Storage for a range of scenarios including serving website content, storing data for archival and disaster recovery, or distributing large data objects to users via direct download.
+> Node.js idiomatic client for [Cloud Storage][product-docs].
```
JustinBeckwith: I don't think this is rendering correctly:
image

samples/README.md
bcoe: I don't know? ported this over from the old file; doesn't seem that useful to me.

samples/README.md
bcoe: this should be fixed now :+1:

samples/README.md
JustinBeckwith: I wonder if it's supposed to open the *directory* in cloud shell?  Maybe something got lost in translation?  Because I could see opening `/samples` being useful. I say either make it do that, or we drop it

samples/README.md
```diff
@@ -1,325 +1,195 @@
...
   * [Notifications](#notifications)
+  * [Quickstart](#quickstart)
   * [Requester Pays](#requester-pays)
-  * [Bucket Lock](#bucket-lock)
 
 ## Before you begin
 
 Before running the samples, make sure you've followed the steps in the
-[Before you begin section](../README.md#before-you-begin) of the client
+[Using the client library](https://github.com/googleapis/nodejs-storage#using-the-client-library) of the client
```
stephenplusplus: This doesn't read very well. Maybe it should be quoted around the link, and we can probably remove all the "client library" phrasing. Maybe "...make sure you've followed [these steps](#link)."

JustinBeckwith:
@bcoe can we land this?

PR #651, opened 2019-03-28
[DO NOT MERGE] Add publish-docs job generated via synth.

busunkim96:
The kokoro job for publishing docs generated with synthtool + repo metadata. See https://github.com/googleapis/synthtool/pull/203

- [ ] Tests and linter pass


busunkim96:
@bcoe I think Autosynth would come back through with a PR to revert the changes the following day (since the templates aren't updated). :) Mostly created this to make sure synth creates the right files.

JustinBeckwith:
I think we have this covered 👍

PR #654, opened 2019-04-01
docs(samples): Samples for v4 Signed URL

jkwlui:


JustinBeckwith:
That's a fine lookin API!

samples/files.js
```diff
@@ -312,6 +313,70 @@ async function generateSignedUrl(bucketName, filename) {
...
+    expires: Date.now() + 1000 * 60 * 60, // one hour
+  };
+
+  // Get a v4 signed URL for reading the file
+  const [url] = await storage
+    .bucket(bucketName)
+    .file(filename)
+    .getSignedUrl(options);
+
+  console.log(`The v4 signed url for reading ${filename} is ${url}`);
```
frankyn: Could you update the output to follow Java's output:

https://github.com/googleapis/google-cloud-java/blob/v4support/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java#L1499-L1502

samples/files.js
```diff
@@ -312,6 +313,70 @@ async function generateSignedUrl(bucketName, filename) {
...
+    contentType: 'application/octet-stream',
+  };
+
+  // Get a v4 signed URL for uploading file
+  const [url] = await storage
+    .bucket(bucketName)
+    .file(filename)
+    .getSignedUrl(options);
+
+  console.log(`The v4 signed url for uploading ${filename} is ${url}`);
```
frankyn: Could you update output to follow Java's output?

https://github.com/googleapis/google-cloud-java/blob/v4support/google-cloud-examples/src/main/java/com/google/cloud/examples/storage/snippets/StorageSnippets.java#L1536-L1542

samples/files.js
jkwlui: Changed!

samples/files.js
```diff
@@ -312,6 +313,79 @@ async function generateSignedUrl(bucketName, filename) {
...
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // const filename = 'File to access, e.g. file.txt';
+
+  // These options will allow temporary read access to the file
+  const options = {
+    version: 'v4',
+    action: 'read',
+    expires: Date.now() + 1000 * 60 * 60, // one hour
```
frankyn: Missed this in the last review, please change expires time to 15 minutes to be consistent with other language samples. For Read and Upload samples.

samples/files.js
```diff
@@ -312,6 +313,79 @@ async function generateSignedUrl(bucketName, filename) {
...
+  };
+
+  // Get a v4 signed URL for reading the file
+  const [url] = await storage
+    .bucket(bucketName)
+    .file(filename)
+    .getSignedUrl(options);
+
+  console.log('Generated GET signed URL:');
+  console.log(url);
```
bcoe: this is a little wordy, I might be tempted to consolidate in a single template string comment:

```
console.info(`generated the signed URL ${url} which is valid for one hour`)
```

samples/system-test/files.test.js
```diff
@@ -137,13 +141,72 @@ it('should make a file public', async () => {
...
 });
 
+it('should generate a v4 signed URL and read a file', async () => {
+  const output = await exec(
+    `${cmd} generate-v4-read-signed-url ${bucketName} ${copiedFileName}`
+  );
+
+  const regExp = new RegExp(
+    'Generated GET signed URL:\n' +
+      '(.*)\n' +
```
bcoe: folks will inevitably change the `console.log` without updating this test, so I think it will be a bit fragile.

Perhaps just assert that some of the content was output from the sample.

samples/system-test/files.test.js
```diff
@@ -137,13 +141,72 @@ it('should make a file public', async () => {
...
+  const text = await res.text();
+  assert.strictEqual(text, fileContent);
+});
+
+it('should generate a v4 signed URL and upload a file', async () => {
+  const output = await exec(
+    `${cmd} generate-v4-upload-signed-url ${bucketName} ${signedFileName}`
+  );
+  const regExp = new RegExp(
+    'Generated PUT signed URL:\n' +
```
bcoe: Again I'd probably try to reduce the size of this regex, perhaps just:

```
output.match(/URL: ([^ ]+)/)
```

samples/system-test/files.test.js
```diff
@@ -137,13 +141,72 @@ it('should make a file public', async () => {
...
+  );
+  assert.match(output, regExp);
+
+  const match = output.match(regExp);
+  const req = {
+    method: 'PUT',
+    headers: {'Content-Type': 'application/octet-stream'},
+    body: fileContent,
+  };
+  await fetch(match[1], req);
```
bcoe: In the name of consistency, I wonder if we should use https://www.npmjs.com/package/teeny-request?

samples/files.js
jkwlui: Good call out, but as @frankyn mentioned, we have standardized the outputs of these samples across languages so we'd just have to follow the canonical :)

samples/system-test/files.test.js
jkwlui: good call, fixed!

samples/system-test/files.test.js
jkwlui: In that case we need to migrate the root system-test as well, they're using fetch right now.

I'll change it here, and open an issue to migrate that as well.

jkwlui: Having looked at the code, I think I'd prefer keeping `node-fetch` as we'll likely have more code to use `teeny-request` since it's callback-only.

PR #664, opened 2019-04-04
refactor: use execSync for tests

JustinBeckwith:


samples/system-test/acl.test.js
```diff
@@ -54,49 +52,47 @@ after(async () => {
...
+  const out = execSync(`${cmd} print-bucket-acl ${bucketName}`);
   assert.match(out, /OWNER: project-editors-/);
   assert.match(out, /OWNER: project-owners-/);
   assert.match(out, /READER: project-viewers-/);
 });
 
-it('should print acl for a file', async () => {
-  const out = await exec(`${cmd} print-file-acl ${bucketName} ${fileName}`);
+it('should print acl for a file', () => {
+  const out = execSync(`${cmd} print-file-acl ${bucketName} ${fileName}`);
```
bcoe: Is the problem with the integration tests potentially that `execSync` is returning a `Buffer` and not a String?

JustinBeckwith: This was addressed above!  Good callout.

PR #666, opened 2019-04-05
fix(ts): improve return types for response metadata

JustinBeckwith:
As a bonus, drops dependencies on `teeny-request` and `@types/request`

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -25,8 +25,6 @@ import * as mime from 'mime-types';
 import * as path from 'path';
 
 const snakeize = require('snakeize');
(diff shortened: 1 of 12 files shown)


src/bucket.ts
```diff
@@ -176,28 +174,28 @@ export interface GetBucketOptions extends GetConfig {
   userProject?: string;
 }
 
-export type GetBucketResponse = [Bucket, request.Response];
+export type GetBucketResponse = [Bucket, Metadata];
 
 export interface GetBucketCallback {
   (err: ApiError|null, bucket: Bucket|null,
```
bcoe: seems like the linting gods might be angry.

PR #684, opened 2019-04-30
fix: validate action of getSignedUrl() function

### LINKED ISSUE #671: getSignedUrl fails silently with invalid url if `action` is invalid/undefined
The `file.getSignedUrl` method will successfully return a `signedUrl` but when making a request via the URL will give a `SignatureDoesNotMatch` response (expected).

- [ ] Add validation for the `action` parameter.

AVaksman:
Fixes #671

- [x] Tests and linter pass
- [x] Code coverage does not decrease

add validation for `action` parameter.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2319,12 +2319,6 @@ class File extends ServiceObject<File> {
    */
   getSignedUrl(cfg: GetSignedUrlConfig, callback?: GetSignedUrlCallback):
       void|Promise<GetSignedUrlResponse> {
-    const validAction = ['read', 'write', 'delete', 'resumable'];
-
-    if (!cfg.action || !validAction.includes(cfg.action)) {
-      throw new Error('The action is not provided or invalid.');
-    }
-
     const expiresInMSeconds = new Date(cfg.expires).valueOf();
 
     if (isNaN(expiresInMSeconds)) {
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -2319,6 +2319,12 @@ class File extends ServiceObject<File> {
    */
   getSignedUrl(cfg: GetSignedUrlConfig, callback?: GetSignedUrlCallback):
       void|Promise<GetSignedUrlResponse> {
+    const validAction = ['read', 'write', 'delete', 'resumable'];
```
bcoe: This looks reasonable to me, might be worth considering using an `Enum` for `TypeScript`, does this translate cleanly to the JavaScript API?

stephenplusplus: There's actually already this enum: https://github.com/googleapis/nodejs-storage/blob/7b1ef2c376e353758879050107426a980c502774/src/file.ts#L2341

We could just validate that `method` is something.

AVaksman: Thanks, done

AVaksman:
> What was the behavior of the old API, if I were to pass in an invalid action? This looks great, my only concern is it might be a breaking change:
> 
> * if the old logic would have failed with a more cryptic error, this seems reasonable as a non-breaking change.
> * if prior to this we would have succeeded, this feels like a breaking change.
> 
> Assuming we would have previously been throwing an exception, I'm 👍 on this change.

Previously `getSignedUrl` with invalid `action` parameter would still successfully return an invalid URL (a URL with permissions for non existing action so to speak) and any request via this URL would result with `SignatureDoesNotMatch` response. The error is cryptic since it points user in completely different direction.

bcoe:
@AVaksman feels like a `fix` to me, I continue to be a :+1:

PR #702, opened 2019-05-07
refactor: remove usage of async module

AVaksman:
Towards googleapis/google-cloud-node/issues/2883

- [x] Tests and linter pass


system-test/storage.ts
```diff
@@ -15,11 +15,12 @@
  */
 
 import * as assert from 'assert';
-import * as async from 'async';
 import * as crypto from 'crypto';
 import * as fs from 'fs';
 import fetch from 'node-fetch';
 const normalizeNewline = require('normalize-newline');
+const pLimit = require('p-limit');
```
JustinBeckwith:
```suggestion
import pLimit = require('p-limit');
```

system-test/storage.ts
```diff
@@ -15,11 +15,12 @@
...
+const {promisify} = require('util');
```
JustinBeckwith:
```suggestion
import {promisify} from 'util';
```

system-test/storage.ts
```diff
@@ -946,14 +898,9 @@ describe('storage', () => {
...
-    after(done => {
-      async.series(
-        bucketsToCreate.map(bucket => {
-          return (done: DeleteBucketCallback) => {
-            storage.bucket(bucket).delete(done);
-          };
-        }),
-        done
+    after(async () => {
+      Promise.all(
```
JustinBeckwith: Missing the `await`

system-test/storage.ts
```diff
@@ -1179,44 +1103,29 @@ describe('storage', () => {
...
-          );
-          done();
-        }
+      await bucket.addLifecycleRule({
+        action: 'delete',
+        condition: {
+          age: 30,
+          isLive: true,
+        },
+      }),
```
JustinBeckwith: weird formatting - would expect a semicolon here

system-test/storage.ts
```diff
@@ -1235,118 +1144,72 @@ describe('storage', () => {
...
-            if (!err) {
-              done(new Error('Expected an error.'));
-              return;
-            }
-            assert.strictEqual((err as ApiError).code, 403);
-            done();
-          }
-        );
+        await bucket.lock(bucket.metadata.metageneration);
+        try {
```
JustinBeckwith: Please use an `assert.rejects` here so we're 100% sure the exception was thrown

system-test/storage.ts
```diff
@@ -1433,19 +1274,14 @@ describe('storage', () => {
...
-            });
-          },
-          callback
+      async function deleteFilesAsync() {
+        await new Promise(resolve =>
+          setTimeout(resolve, RETENTION_PERIOD_SECONDS * 1000)
+        );
+        return Promise.all(
+          FILES.map(file =>
+            file.setMetadata({temporaryHold: null}).then(() => file.delete())
```
JustinBeckwith: The mix of promises and async/await is weird.  Can we just use an async function here and awaits?

system-test/storage.ts
```diff
@@ -2758,19 +2502,11 @@ describe('storage', () => {
...
-          assert.ifError(err);
-          async.parallel(
-            [file!.delete.bind(file), copiedFile!.delete.bind(copiedFile)],
-            done
-          );
-        });
-      });
+      const [file] = await bucket.upload(FILES.logo.path, opts);
+      const [copiedFile] = await file!.copy('CloudLogoCopy');
+      await Promise.all([file!.delete, copiedFile!.delete()]);
```
JustinBeckwith: Should this be `file.delete()`?

system-test/storage.ts
AVaksman: Fixed, thanks

system-test/storage.ts
AVaksman: Fixed as well, thanks

system-test/storage.ts
AVaksman: added await, thanks

system-test/storage.ts
AVaksman: True, fixed it

system-test/storage.ts
AVaksman: `assert.fejects` doesn't accept `async` function untill node 10.
reverted to 
```ts
bucket.setRetentionPeriod(RETENTION_DURATION_SECONDS / 2)
      .then(util.noop, err => assert.strictEqual(err.code, 403));
```
What version of Node are we supporting in `system-test`?

system-test/storage.ts
AVaksman: Changed that, thanks

system-test/storage.ts
AVaksman: Yes it should, fixed, thanks

AVaksman:
> This is such a huge improvement. Thank you!

Thank you.

AVaksman:
@JustinBeckwith PTAL

PR #708, opened 2019-05-10
feat: add file.isPublic() function

### LINKED ISSUE #683: File method to check whether an Object is Public
Right now there exist no method on `File` object to explicitly check whether a given a file is Public or Not. 

There should be a way to check this.

AVaksman:
Fixes #683

- [x] Tests and linter pass


AVaksman:
> `gaxios` is not in `package.json` explicitly. If we're using `gaxios` to make requests, then we should add it to `package.json` under `dependencies`

Done, thank you

system-test/storage.ts
```diff
@@ -3140,26 +3138,6 @@ describe('storage', () => {
     );
   }
 
-  async function isFilePublicAsync(file: File) {
```
stephenplusplus: It might be good to keep this because the `makePublic()` function is accomplishing that a certain way, and this makes sure we did it that way.

AVaksman: Agree, thank you
Moved `isPublic()` test to a scenario without credentials.

system-test/storage.ts
```diff
@@ -209,17 +209,12 @@ describe('storage', () => {
...
-          (err, files) => {
-            assert.ifError(err);
-            const file = files![0];
-            file.download(done);
-          }
-        );
+      it('should list and download a file', async () => {
+        const [files] = await bucket.getFiles({autoPaginate: false});
+        const file = files[0];
+        const [isPblc] = await file.isPublic();
```
stephenplusplus: Would it be okay to use the full `isPublic` word? I could see other maintainers and contributors getting into a typo battle with this var name.

src/file.ts
```diff
@@ -2595,6 +2601,63 @@ class File extends ServiceObject<File> {
...
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * file.isPublic().then(function(data) {
+   *   const resp = data[0];
+   * });
+   */
+
+  isPublic(callback?: IsPublicCallback): Promise<IsPublicResponse> | void {
+    gaxiosRequest({
+      method: 'HEAD',
```
stephenplusplus: We should add test(s) to check that we're sending `HEAD` and the correctly formatted URL.

test/file.ts
```diff
@@ -3205,6 +3206,39 @@ describe('File', () => {
...
+    it('should execute callback with `false` in response', done => {
+      sandbox.stub(gaxios, 'request').rejects({code: '403'});
+      file.isPublic((err: gaxios.GaxiosError, resp: boolean) => {
+        assert.ifError(err);
+        assert.strictEqual(resp, false);
+        done();
+      });
+    });
+
+    it('should propagate different error to user', done => {
```
stephenplusplus: Changing the message to `should propagate non-403 errors to user` might help others understand how this works.

src/file.ts
```diff
@@ -2595,6 +2601,63 @@ class File extends ServiceObject<File> {
...
+      url: `http://${this.bucket.name}.storage.googleapis.com/${this.id}`,
```
stephenplusplus: I think we need to encode the name:

```js
url: `http://${this.bucket.name}.storage.googleapis.com/${encodeURIComponent(this.name)}`,
```

system-test/storage.ts
AVaksman: Sure, done

src/file.ts
AVaksman: Added tests, tanks

test/file.ts
AVaksman: Sure, done

src/file.ts
AVaksman: sure, done

It only matters in the case where `file.name` property is changed on an already existing `file` object, which would render all other file properties' values incorrect in respect to the new `file.name`.
Shouldn't the `file.name` property be made `readonly`?

test/file.ts
```diff
@@ -3205,6 +3206,60 @@ describe('File', () => {
...
+        assert.ifError(err);
+        assert.strictEqual(spy.calledWithMatch({method: 'HEAD'}), true);
+        done();
+      });
+    });
+
+    it('should correctly format URL in the request', done => {
+      const expecterURL = `http://${
+        BUCKET.name
+      }.storage.googleapis.com/${encodeURIComponent(file.name)}`;
```
stephenplusplus: Could you check that the name is encoded by throwing a special character in there?

AVaksman: good point, done

src/file.ts
stephenplusplus: > It only matters in the case where file.name property is changed

Do you mean the `name` property is already encoded? I didn't see that in the code.

> Shouldn't the file.name property be made readonly?

I suppose. `name` wasn't intended to be a public property at all.

src/file.ts
```diff
@@ -2595,6 +2601,65 @@ class File extends ServiceObject<File> {
...
+   * @callback IsPublicCallback
+   * @param {?Error} err Request error, if any.
+   * @param {boolean} resp Whether file is public or not.
+   */
+  /**
+   * @typedef {array} IsPublicResponse
+   * @property {boolean} 0 Whether file is public or not.
+   */
+  /**
+   * Check whether this file is public or not.
```
stephenplusplus: We should explain how we are doing this. It is possible the server has a hiccup and returns an error code, even though the file is public.

AVaksman: Done! 
Thanks

AVaksman:
> Thanks @AVaksman for putting this together. Apologies for the delay in my review, LGTM.

Thanks.

Issue #540, opened 2018-12-05
Add bucket.uploadDirectory method


crwilcox:
A request was made via documentation feedback to have support or a sample for uploading a directory in addition to just having `bucket.upload`

crwilcox:
@JCIV-SE the method `bucket.upload` takes files not directories so it is expected that currently this doesn't work.

Though, I am now thinking maybe overloading `bucket.upload` to also take a directory may be a good idea if we were to undertake this work.

PR #731, opened 2019-06-06
feat: implement bucket.uploadDirectory() method

AVaksman:
Towards #540

- [x] Tests and linter pass

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -30,6 +30,7 @@ import {
 import {paginator} from '@google-cloud/paginator';
 import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
+import pLimit from 'p-limit';
 import * as async from 'async';
 import * as extend from 'extend';
 import * as fs from 'fs';
@@ -322,6 +323,16 @@ export interface UploadOptions
   kmsKeyName?: string;
   resumable?: boolean;
 }
+
+export interface UploadDirectoryOptions extends CreateWriteStreamOptions {
+  recurse?: boolean;
+}
+
+export interface UploadDirectoryCallback {
+  (err: Error | null, resp?: Array<{[k: string]: string | Error}>): void;
+}
+
+export type UploadDirectoryResponse = [Array<{[k: string]: string | Error}>];
 
 export interface MakeAllFilesPublicPrivateOptions {
   force?: boolean;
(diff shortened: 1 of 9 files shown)


src/bucket.ts
```diff
@@ -323,6 +324,16 @@ export interface UploadOptions
   resumable?: boolean;
 }
 
+export interface UploadDirectoryOptions extends CreateWriteStreamOptions {
+  recurse?: boolean;
+}
+
+export interface UploadDirectoryCallback {
+  (err: Error | null, resp?: Array<{[k: string]: string | Error}>): void;
```
callmehiphop: Can we move the nested object to its own type?

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+    directoryPathString: string,
+    options: UploadDirectoryOptions,
+    callback: UploadDirectoryCallback
+  ): void;
+  uploadDirectory(
+    directoryPathString: string,
+    callback: UploadDirectoryCallback
+  ): void;
+  /**
+   * @typedef {object} UploadDirectoryOptions Configuration options for Bucket#uploadDirectory().
```
callmehiphop: Lets wrap this in a `{@link}` tag

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * @param {boolean} [options.recurse] Whether to recursevely upload all files from
```
callmehiphop: I *think* these should be `@property` tags

callmehiphop: Also looks like a typo here

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * @param {string|boolean} [options.validation] Possible values: `"md5"`,
+   *     `"crc32c"`, or `false`. By default, data integrity is validated with an
+   *     MD5 checksum for maximum reliability. CRC32c will provide better
+   *     performance with less reliability. You may also choose to skip
+   * validation completely, however this is **not recommended**.
+   */
+  /**
+   * @typedef {array} UploadDirectoryResponse
+   * @property {object} [response] Response for each of the uploaded files.
+   * @param {string} [filename] Absolute file path.
```
callmehiphop: I think these should all be `@property`

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   *     billed for the request.
+   * @param {string|boolean} [options.validation] Possible values: `"md5"`,
+   *     `"crc32c"`, or `false`. By default, data integrity is validated with an
+   *     MD5 checksum for maximum reliability. CRC32c will provide better
+   *     performance with less reliability. You may also choose to skip
+   * validation completely, however this is **not recommended**.
+   */
+  /**
+   * @typedef {array} UploadDirectoryResponse
+   * @property {object} [response] Response for each of the uploaded files.
```
callmehiphop: I think since we're describing an array these should be index values.

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   */
+  /**
+   * @callback UploadDirectoryCallback
+   * @param {?Error} err Request error, if any.
+   * @param {array} resp The list of object responses for every uploaded file.
+   */
+
+  /**
+   * Upload a directory to the bucket. This is a convenience method for multiple
+   * {@link Bucket#upload}.
```
callmehiphop: I think maybe this should say something like
> This is a convenience method for making multiple `{@link Bucket#upload}` calls.

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * @returns {Promise<UploadDirectoryResponse>}
+   *
+   * @example
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const bucket = storage.bucket('albums');
+   *
+   * //-
+   * // Upload a directory from a local path.
+   * //-
```
callmehiphop: The `//-` syntax was for an old version of the docs, we shouldn't be using it anymore. Instead lets use multiple `@example` tags and `<caption>` to add a descriptions to the examples.

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * bucket.uploadDirectory('/local/path/myfolder', function(err, resp) {
+   *   // Your bucket now contains:
+   *   // - "myfolder/" directory (with the contents of `/local/path/myfolder/')
+   *
+   *   // `file` is an instance of a File object that refers to your new file.
```
callmehiphop: Is this leftover from copy/paste? I don't see a `file` variable anywhere

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * });
+   *
+   *
+   * //-
+   * // You can also have files gzip'd on the fly.
+   * //-
+   * bucket.uploadDirectory(path.join(__dirname, 'index', { gzip: true }, function(err, resp) {
```
callmehiphop: We should either `import` path or not use it at all for the example.

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   *   // Your bucket now contains:
+   *   // - "index/" directory (automatically compressed with gzip)
+   *
+   *   // Downloading the file with `file.download` will automatically decode
+   * the
```
callmehiphop: looks like this comment needs some formatting love, also I'm a little confused about the context of `file.download`, are we saying any of the files you just uploaded will be automatically decoded?

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   *   // file.
+   * });
+   *
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * bucket.uploadDirectory('local-directory-path').then(function(data) {
```
callmehiphop: Could probably make this a little more modern with

```js
const [response] = await bucket.uploadDirectory('local-directory-path');
```

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+   * });
+   */
+
+  uploadDirectory(
+    directoryPathString: string,
+    optionsOrCallback?: UploadDirectoryOptions | UploadDirectoryCallback,
+    callback?: UploadDirectoryCallback
+  ): Promise<UploadDirectoryResponse> | void {
+    // tslint:disable-next-line:no-any
+    if ((global as any)['GCLOUD_SANDBOX_ENV']) {
```
callmehiphop: @stephenplusplus maybe we should move this to `process.env.GCLOUD_SANDBOX_ENV`?

callmehiphop: Actually do we even need this anymore? I almost think this was from a time where we were smoke testing the JSDoc snippets

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+    }
+
+    const options =
+      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    callback =
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : callback;
+
+    let stat: fs.Stats | Error;
+    try {
+      stat = fs.lstatSync(directoryPathString);
```
callmehiphop: I think we might want to use the async version of this? @bcoe @stephenplusplus thoughts?

src/bucket.ts
```diff
@@ -3174,6 +3185,201 @@ class Bucket extends ServiceObject {
...
+    Promise.all(
+      fileList.map(file =>
+        limit(() => {
+          let destination = path.relative(pathDirName, file);
+          if (process.platform === 'win32') {
+            destination = destination.replace(/\\/g, '/');
+          }
+          const uploadOptions = extend({}, {destination}, options);
+          return this.upload(file, uploadOptions).then(
+            () => ({fileName: uploadOptions.destination, status: 'success'}),
```
callmehiphop: I feel like we're definitely going to want to pass back the entire response body somehow

src/bucket.ts
stephenplusplus: Yep, we don't use this anymore.

src/bucket.ts
AVaksman: Done. Left the types as is till we finalize what this method should return.

src/bucket.ts
AVaksman: Done, thanks

src/bucket.ts
AVaksman: Done, thanks

src/bucket.ts
AVaksman: done

src/bucket.ts
AVaksman: done

src/bucket.ts
AVaksman: done

src/bucket.ts
AVaksman: yep, removed

src/bucket.ts
AVaksman: removed

src/bucket.ts
AVaksman: changed

src/bucket.ts
AVaksman: done

src/bucket.ts
AVaksman: using new style examples for docs

src/bucket.ts
AVaksman: Yes, the client library always requests data from the server "as is" via the [headers](https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts#L1238-L1241) ([docs](https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding)) and automatically decompresses if [`isCompressed`](https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts#L1306)

src/bucket.ts
AVaksman: Do you prefer something like
```ts
return this.upload(file, uploadOptions).then(
            (res) => ({fileName: uploadOptions.destination, status: res[0]}),
            (err: Error) => ({fileName: uploadOptions.destination, status: err})
          );
```
`File` object contains `metadata` property which is the `Metadata` part of the `bucket.upload` response. So essentially we will be returning entire `bucket.upload` response.
Please share you thoughts.

callmehiphop: Ah ok, that sounds good to me! Thanks for the break down

src/bucket.ts
```diff
@@ -3174,6 +3189,177 @@ class Bucket extends ServiceObject {
...
+      throw new Error(`${directoryPathString} is an invalid directory`);
+    }
+
+    const fileList: string[] = [];
+    const pathDirName = path.dirname(directoryPathString);
+
+    getFileList(directoryPathString);
+
+    function getFileList(directory: string) {
+      const items = fs.readdirSync(directory);
```
callmehiphop: I think we want to use the async versions of all the `fs` methods, I believe the sync versions are all blocking operations.

src/bucket.ts
```diff
@@ -3174,6 +3189,177 @@ class Bucket extends ServiceObject {
...
+        const stat = fs.lstatSync(fullPath);
+        if (stat.isFile()) {
+          fileList.push(fullPath);
+        } else if (options.recurse && stat.isDirectory()) {
+          getFileList(fullPath);
+        }
+      });
+    }
+
+    const limit = pLimit(10);
```
callmehiphop: I think we should make this number configurable by the user

Issue #746, opened 2019-06-14
refactor: remove dependency on async module


bcoe:
we still rely on `async` in a few places, which broke a recent release of the module (as it was a devDependency only).

AVaksman:
@bcoe is devDependency on `async` is OK or rather remove that as well?

bcoe:
@AVaksman @callmehiphop I think if it simplifies our refactor, keeping `async` as a dev dependency for the time being is reasonable.

Will leave it to you and @callmehiphop to decide.

PR #747, opened 2019-06-14
refactor: remove async usage from bucket

callmehiphop:
Relates to #746

- [x] Tests and linter pass

### DIFF
--- a/samples/bucketLock.js
+++ b/samples/bucketLock.js
@@ -34,7 +34,9 @@ async function setRetentionPolicy(bucketName, retentionPeriod) {
     .bucket(bucketName)
     .setRetentionPeriod(retentionPeriod);
   console.log(
-    `Bucket ${bucketName} retention period set for ${metadata.retentionPolicy.retentionPeriod} seconds.`
+    `Bucket ${bucketName} retention period set for ${
+      metadata.retentionPolicy.retentionPeriod
+    } seconds.`
   );
   // [END storage_set_retention_policy]
 }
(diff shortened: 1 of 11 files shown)


callmehiphop:
Working on tests, but early feedback would be lovely!

samples/bucketLock.js
```diff
@@ -34,7 +34,9 @@ async function setRetentionPolicy(bucketName, retentionPeriod) {
     .bucket(bucketName)
     .setRetentionPeriod(retentionPeriod);
   console.log(
-    `Bucket ${bucketName} retention period set for ${metadata.retentionPolicy.retentionPeriod} seconds.`
+    `Bucket ${bucketName} retention period set for ${
```
JustinBeckwith: Going to want to `rm -rf node_modules package-lock.json` && `npm run fix` here.  Prettier made an update in a patch release I think that's causing some grief.

callmehiphop: Whoops, well this is embarrassing 😱

src/bucket.ts
```diff
@@ -30,11 +30,12 @@ import {
 import {paginator} from '@google-cloud/paginator';
 import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
-import * as async from 'async';
 import * as extend from 'extend';
 import * as fs from 'fs';
 import * as mime from 'mime-types';
 import * as path from 'path';
+import pLimit from 'p-limit';
```
bcoe: have we opted to move away from `async` to `p-limit` for other libraries. Curious what the advantages are.

callmehiphop: Good question, this is the only one that I'm aware of. I saw that #731 planned on pulling it in as a regular dependency, so I'm sort of piggy backing off of that.

AFAIK most of the ways we used `async` here can be replaced by native Promises, but managing concurrency is not something that comes out of the box there. In the past I've usually turned to [p-queue](https://www.npmjs.com/package/p-queue), but `p-limit` works here too and we were already slated to use it.

AVaksman: Followed the suggestion [here](https://github.com/googleapis/nodejs-spanner/pull/449#issuecomment-442529440)

PR #751, opened 2019-06-20
feat: hmac service account

jkwlui:
Added the following `Storage` client methods:
 - [x] createHmacKey
 - [x] getHmacKeys
 - [x] hmacKey

HmacKey class and methods:
 - [x] get
 - [x] delete
 - [x] update


bcoe (approved):
this is looking pretty lovely to me, on a quick read.

bcoe:
looking fairly good to me, once tests are passing :smile:

frankyn (changes requested):
Overall, looks good. I have two nit questions.

src/hmacKey.ts
```diff
@@ -0,0 +1,355 @@
...
+       * @param {DeleteHmacKeyOptions} [options] Configuration options.
+       * @param {DeleteHmacKeyCallback} [callback] Callback function.
+       * @returns {Promise<DeleteHmacKeyResponse>}
+       *
+       * @example
+       * const {Storage} = require('@google-cloud/storage');
+       * const storage = new Storage();
+       *
+       * //-
+       * // Delete HMAC key after making the key inactive.
```
stephenplusplus: Is this behavior that we could do by default, or at least optionally?

src/hmacKey.ts
```diff
@@ -0,0 +1,355 @@
...
+    super({
+      parent: storage,
+      baseUrl: `/projects/${storage.projectId}/hmacKeys`,
+      id: accessId,
+      methods,
+    });
+
+    this.accessId = accessId;
+    this.metadata = {accessId};
+    this.parent = storage;
```
stephenplusplus: This should be done automatically by the ServiceObject class.

src/hmacKey.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    super({
+      parent: storage,
+      baseUrl: `/projects/${storage.projectId}/hmacKeys`,
+      id: accessId,
+      methods,
+    });
+
+    this.accessId = accessId;
+    this.metadata = {accessId};
```
stephenplusplus: We usually don't stub metadata, as that's meant to be reserved for only server-side representations. It is refreshed with every `getMetadata()` or `setMetadata()`, and set by `getHmacKeys()`.

src/hmacKey.ts
```diff
@@ -0,0 +1,355 @@
...
+    const opts = Object.assign({}, options);
+    // autoCreate is ignored - key must be created using Storage.createHmacKey.
+    delete opts.autoCreate;
+
+    const reqOpts = {
+      uri: '/',
+      qs: opts,
+    };
+
+    this.request(reqOpts, (err, metadata, res) => {
```
stephenplusplus: Did you try the default behavior of `get` here, that is set by the ServiceObject class?

src/hmacKey.ts
```diff
@@ -0,0 +1,355 @@
...
+        'Cannot update HmacKey with an undefined/empty options object.'
+      );
+    }
+
+    const {options, callback} = normalize<
+      UpdateHmacKeyOptions,
+      HmacKeyMetadataCallback
+    >(optionsOrCb, cb);
+
+    this.request(
```
stephenplusplus: `this.setMetadata()` should be able to do all of this, and we won't need the `update()` method.

src/storage.ts
```diff
@@ -550,6 +626,132 @@ export class Storage extends Service {
...
+    const {options, callback} = normalize<
+      CreateHmacKeyOptions,
+      CreateHmacKeyCallback
+    >(optionsOrCb, cb);
+    const query = Object.assign({}, options, {serviceAccountEmail});
+
+    this.request(
+      {
+        method: 'POST',
+        uri: `/projects/${this.projectId}/hmacKeys`,
```
stephenplusplus: I think we need to get the project ID from the auth client first. Or does this still do the placeholder find and replace?

src/storage.ts
```diff
@@ -550,6 +626,132 @@ export class Storage extends Service {
...
+
+        const hmacKey = new HmacKey(this, resp.metadata.accessId);
+        hmacKey.metadata = resp.metadata;
+
+        const resource = {
+          hmacKey,
+          secret: resp.secret,
+        };
+
+        callback!(null, resource, resp);
```
stephenplusplus: This signature would be the first to break from the convention across the libraries, where we strictly return the created resource instance. Can the secret be stored on the HmacKey instance?

src/storage.ts
```diff
@@ -664,6 +866,114 @@ export class Storage extends Service {
...
+   * }, callback);
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * storage.getHmacKeys().then(function(data) {
+   *   const hmacKeys = data[0];
+   * });
+   */
+  getHmacKeys(options?: GetHmacKeysOptions): Promise<GetHmacKeysResponse>;
```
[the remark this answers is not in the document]
jkwlui: Yes, although I don't have a strong opinion on `list` vs `get`.

Using `get` will give us consistency within this library while using `list` allows us to follow the API naming and be consistent with other languages. 😅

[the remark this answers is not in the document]
frankyn: That's fine. I'd rather it be consistent with the library at this point. If we could go back in time I'd suggest list over get per the API.

src/hmacKey.ts
jkwlui: Are you suggesting something like, check if the key is active, if it is then make it inactive then delete when calling `hmacKey.delete`?

stephenplusplus: Yep!

src/hmacKey.ts
jkwlui: Ah I see. `ServiceObject.get()` wraps `getMetadata()` and auto-creates the object if it doesn't exist.

Auto-create is not defined for HMAC key, so we should just expose the `getMetadata()` method inherited from `ServiceObject`. Does that make sense? @stephenplusplus

src/hmacKey.ts
jkwlui: ack.

src/hmacKey.ts
stephenplusplus: We can actually keep "get()", and if that autoCreate setting is passed, it would be ignored if we don't have a "create" method defined. But just curious, why can't we have the create method? If it is because it requires additional configuration, that is actually possible to be defined in the get(options) argument.

jkwlui: The `id` (accessId) is generated by the server, so `autoCreate` wouldn't make sense for an HMAC key.

jkwlui: I should explain more: an HMAC key is different from, say, a Bucket where a user defines the name (id) of the possibly non-existent resource.

stephenplusplus: Ah, gotcha. Thanks for explaining! So I think you can still pass 'methods.get = true' in the constructor and as long as methods.create isn't set, it'll work as expected. Let me know if I'm wrong on that :)

src/storage.ts
jkwlui: Unlike buckets or files where the API response is the resource itself, creating HMAC key returns a secret in addition to the resource. This makes it clearer that the secret is only returned on creation.

src/storage.ts
```diff
@@ -754,19 +1064,37 @@ export class Storage extends Service {
...
+   * retrieve and populate the metadata.
+   *
+   * @param {string} accessId The HMAC key's access ID.
+   * @returns {HmacKey}
+   * @see HmacKey
+   *
+   * @example
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const hmacKey = storage.hmacKey('ACCESS_ID');
```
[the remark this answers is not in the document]
jkwlui: Storage#hmacKey returns a reference to an HmacKey object.

src/hmacKey.ts
jkwlui: Yup, looks like that works. On second thought, I think we should expose `getMetadata` rather than `get`, since the get method doesn't make much sense for HMAC keys without `autoCreate`.

stephenplusplus: `get`'s purpose has rightly been [called into question before](https://github.com/googleapis/nodejs-common/issues/234), but it is still something we expose across the libraries wherever `getMetadata` exists, even if `autoCreate` isn't available. I would vote to keep that going for the sake of parity.

src/storage.ts
stephenplusplus: Could it be `callback(null, hmacKey, resp.secret, resp)`? Or maybe if we said in the docs where `secret` can be found within the `resp` argument? I think the consistent hierarchy is important to maintain, where `createThing` returns `Thing`, unboxed.

src/storage.ts
jkwlui: Yes, I expect the placeholder replacement would work here. Any concerns that it might not?

src/hmacKey.ts
jkwlui: hmacKey only provides `PUT` method to update the key, while `setMetadata` makes a `PATCH` request which it doesn't support.

src/hmacKey.ts
jkwlui: Added `get` and `getMetadata` methods (inherited from `ServiceObject`)

src/hmacKey.ts
jkwlui: removed `this.metadata = {accessId};`

src/hmacKey.ts
jkwlui: fixed.

src/hmacKey.ts
jkwlui: Decided against adding the force delete logic to prevent unintended deletion of key. Users need to explicitly make the key inactive first before deleting in case a key is still being used somewhere.

src/hmacKey.ts
stephenplusplus: You can actually configure this in the constructor:

setMetadata: {
  reqOpts: {
    method: 'PUT',
  },
}

src/storage.ts
stephenplusplus: Actually, looking at this again, what is the full URL you are posting to?

frankyn (approved):

src/hmacKey.ts
stephenplusplus: To elaborate on that, that object would go in the `const methods = {}` object in the constructor, as a sibling to `get: true` and the other properties. When you need further customization, you can use an object instead of `true` to define default request options.

src/storage.ts
jkwlui: `https://www.googleapis.com/storage/v1/projects/{{projectId}}/hmacKeys`

stephenplusplus: Gotcha, thanks. This should work as expected. Sorry for my confusion!

src/hmacKey.ts
jkwlui: It looks like setting `reqOpts` here will not override the `method` config:

https://github.com/googleapis/nodejs-common/blob/deea6b2c5581b0d9892b98c7b17506039dd1a5e4/src/service-object.ts#L455-L460

The override is overwritten by the last object.

stephenplusplus: [*forehead slap*](https://github.com/googleapis/nodejs-common/pull/353/files)

I can fix this soon, but if you wanted to jump ahead, it should be something like:

```js
const reqOpts = extend(
  true,
  {
    uri: '',
    method: 'PATCH',
  },
  methodConfig.reqOpts,
  {
    json: metadata,
    qs: options,
  }
});
```

jkwlui: Pending release: https://github.com/googleapis/nodejs-common/pull/452

src/hmacKey.ts
```diff
@@ -193,62 +246,69 @@ export class HmacKey extends ServiceObject<HmacKeyMetadata | undefined> {
...
        *   .then((data) => {
-       *     const hmacKey = data[0];
+       *     const hmacKeyMetadata = data[0];
+       *     console.log(hmacKeyMetadata);
        *   });
        */
-      get: true,
+      setMetadata: {
+        reqOpts: {
+          method: 'PUT',
```
stephenplusplus: ❤️

test/hmacKey.ts
```diff
@@ -198,21 +203,9 @@ describe('HmacKey', () => {
...
-
-      it('should throw with an empty metadata object', () => {
-        assert.throws(() => {
-          hmacKey.update({}, assert.ifError);
-        }, /Cannot update HmacKey/);
-      });
-
-      it('should accept an options object', () => {
-        hmacKey.update({state: 'INACTIVE'}, {}, assert.ifError);
+    describe('setMetadata', () => {
```
stephenplusplus: No tests needed! The common tests cover that for us, all we need to know is that we properly defined the configuration object in the constructor.

stephenplusplus: Same for all of these methods, actually.

jkwlui: Removed, thank you!

src/hmacKey.ts
jkwlui: Do we need to update the minimum `@google-cloud/common` version in package.json here?

stephenplusplus: Oh yes, good catch.

jkwlui:
@stephenplusplus thanks for your review - can I get another look + check off for this one? Thanks!

test/index.ts
```diff
@@ -189,6 +192,127 @@ describe('Storage', () => {
...
+      assert.throws(
+        () =>
+          storage.createHmacKey({
+            userProject: 'my-project',
+          }),
+        /must be a service account email/
+      );
+    });
+
+    it('should honor the userProject option', async () => {
```
stephenplusplus: Since we don't have any special lines in the code that handle this option, I don't think we should have a test for it. If you want to just make sure that arbitrary user-provided options are treated as a query object (the line `const query = Object.assign({}, options, {serviceAccountEmail});`), let's have a test for that. (You could also test that it doesn't modify the user's object as well)

test/index.ts
```diff
@@ -189,6 +192,127 @@ describe('Storage', () => {
...
+      };
+
+      storage.createHmacKey(
+        SERVICE_ACCOUNT_EMAIL,
+        (err: Error, hmacKey: HmacKey, secret: string) => {
+          assert.ifError(err);
+          assert.strictEqual(secret, response.secret);
+          assert(hmacKey instanceof HmacKey);
+          assert.strictEqual(hmacKey.metadata, metadataResponse);
+          assert.strictEqual(hmacKey.accessId, metadataResponse.accessId);
```
stephenplusplus: This checks behavior of the HmacKey class, which we shouldn't worry about here.

test/index.ts
```diff
@@ -189,6 +192,127 @@ describe('Storage', () => {
...
+          apiResponse: HmacKeyResourceResponse
+        ) => {
+          assert.ifError(err);
+          assert.strictEqual(apiResponse, response);
+          done();
+        }
+      );
+    });
+
+    it('should execute callback with request error', done => {
```
stephenplusplus: Please check that it also gets the apiResponse argument as well.

test/index.ts
```diff
@@ -508,6 +632,154 @@ describe('Storage', () => {
...
+        callback(null, {nextPageToken: token, items: []});
+      });
+
+      storage.getHmacKeys(
+        {maxResults: 5},
+        // tslint:disable-next-line: no-any
+        (err: Error, _hmacKeys: [], nextQuery: any) => {
+          assert.ifError(err);
+          assert.strictEqual(nextQuery.pageToken, token);
+          assert.strictEqual(nextQuery.maxResults, 5);
```
stephenplusplus: Using the real `maxResults` option makes it seem like that's actually doing something, when I think the real test is that the user-provided options are still in tact inside of `nextQuery`. Could you separate the query object and confirm that the properties still exist?

```js
const query = {a: 'b', c: 'd'}
const expectedNextQuery = Object.assign({}, query, {pageToken: token})
storage.getHmacKeys(query, (err, _hmacKeys, nextQuery) => {
  assert.strictEqual(nextQuery, expectedNextQuery)
})
```

test/index.ts
```diff
@@ -508,6 +632,154 @@ describe('Storage', () => {
...
+      );
+    });
+
+    it('should return null nextQuery if there are no more results', done => {
+      storageRequestStub.callsFake((_opts: {}, callback: Function) => {
+        callback(null, {items: []});
+      });
+
+      storage.getHmacKeys(
+        {maxResults: 5},
```
stephenplusplus: Maybe don't provide a query object here at all for that same reason above, about how it could make it confusing that `maxResults` has an effect.

test/index.ts
```diff
@@ -508,6 +632,154 @@ describe('Storage', () => {
...
+        (err: Error, _hmacKeys: [], nextQuery: {}) => {
+          assert.ifError(err);
+          assert.strictEqual(nextQuery, null);
+          done();
+        }
+      );
+    });
+
+    it('should return HmacKey objects', done => {
```
stephenplusplus: Can you combine this test with `should populate returned HmacKey object with accessId and metadata`? I think we can handle that all in one place.

src/storage.ts
```diff
@@ -754,19 +1052,37 @@ export class Storage extends Service {
...
+   */
+  hmacKey(accessId: string) {
```
stephenplusplus: I believe we need tests for this method!

src/storage.ts
```diff
@@ -754,19 +1052,37 @@ export class Storage extends Service {
...
+    return new HmacKey(this, accessId);
```
stephenplusplus: We should throw here if accessId is not provided, and remove the throw from the HmacKey constructor.

src/storage.ts
```diff
@@ -754,19 +1052,37 @@ export class Storage extends Service {
...
+  }
 }
 
 /*! Developer Documentation
  *
  * These methods can be auto-paginated.
  */
-paginator.extend(Storage, 'getBuckets');
+paginator.extend(Storage, ['getBuckets', 'getHmacKeys']);
```
stephenplusplus: I think this needs the test updated.

src/storage.ts
```diff
@@ -754,19 +1052,37 @@ export class Storage extends Service {
...
 
 /*! Developer Documentation
  *
  * All async methods (except for streams) will return a Promise in the event
  * that a callback is omitted.
  */
 promisifyAll(Storage, {
-  exclude: ['bucket', 'channel'],
+  exclude: ['bucket', 'channel', 'hmacKey'],
```
stephenplusplus: I think this needs the test updated.

stephenplusplus (changes requested):

src/hmacKey.ts
```diff
@@ -0,0 +1,330 @@
...
+    };
+
+    super({
+      parent: storage,
+      baseUrl: `/projects/${storage.projectId}/hmacKeys`,
+      id: accessId,
+      methods,
+    });
+
+    this.accessId = accessId;
```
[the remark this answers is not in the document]
jkwlui: Yes. I'm mirroring what we do with Bucket id (name):

`this.name = name;`

test/index.ts
jkwlui: 8933248

src/hmacKey.ts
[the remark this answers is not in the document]
stephenplusplus: Ah, ok. You can skip it here. We do that because we need that name at different places in the code, if I remember correctly. But we don't advertise it as being publicly accessible.

test/index.ts
jkwlui: 536c8aa

src/storage.ts
jkwlui: fixed

src/storage.ts
jkwlui: Added to `promisifyAll` test to make sure `hmacKey` is in the array passed to the function. Did I miss another place where tests should be updated?

src/hmacKey.ts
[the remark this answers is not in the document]
jkwlui: Ah gotcha!

test/index.ts
jkwlui: fixed!

src/storage.ts
stephenplusplus: Nope, that was all. Thanks!

src/hmacKey.ts
```diff
@@ -0,0 +1,323 @@
...
+    });
+  }
+}
+
+/*! Developer Documentation
+ *
+ * All async methods (except for streams) will return a Promise in the event
+ * that a callback is omitted.
+ */
+promisifyAll(HmacKey);
```
stephenplusplus: You could check, but I don't believe we need this if we're just using the inherited methods.

test/hmacKey.ts
```diff
@@ -0,0 +1,96 @@
...
+      };
+
+      hmacKey = new HmacKey(STORAGE, ACCESS_ID);
+    });
+
+    it('should promisify all the things', () => {
+      assert(promisifyAllStub.calledOnce);
+    });
+
+    it('should assign Storage instance', () => {
```
stephenplusplus: We can remove this test, since this is behavior done in the parent.

test/index.ts
```diff
@@ -189,6 +192,154 @@ describe('Storage', () => {
     });
   });
 
+  describe('hmacKey', () => {
+    it('should throw if accessId is not provided', () => {
+      assert.throws(() => {
+        storage.hmacKey();
+      }, /access ID is needed/);
```
stephenplusplus: Could you check the full error message?: `An access ID is needed to create an HmacKey object.`

test/index.ts
```diff
@@ -189,6 +192,154 @@ describe('Storage', () => {
...
+        callback(null, response);
+      };
+
+      storage.createHmacKey(SERVICE_ACCOUNT_EMAIL, done);
+    });
+
+    it('should throw without a serviceAccountEmail', () => {
+      assert.throws(
+        () => storage.createHmacKey(),
+        /must be a service account email/
```
stephenplusplus: Could you check the full error message (here and the other spots): `The first argument must be a service account email to create an HMAC key.`

src/storage.ts
```diff
@@ -247,6 +289,35 @@ export class Storage extends Service {
...
+   * //-
+   * // If you anticipate many results, you can end a stream early to prevent
+   * // unnecessary processing and API requests.
+   * //-
+   * storage.getHmacKeysStream()
+   *   .on('data', function(bucket) {
+   *     this.end();
+   *   });
+   */
+  getHmacKeysStream: () => Readable;
```
stephenplusplus: Could you add this to https://github.com/googleapis/nodejs-storage/blob/0b8cbb11c9dcddff3b315eff5357655cac988909/test/index.ts#L114

stephenplusplus (approved):
Sorry that I missed some stuff last time. I'll pre-approve, since these should be pretty trivial. Thank you for your patience with my comments!

src/storage.ts
jkwlui: fixed!

src/storage.ts
```diff
@@ -754,19 +1052,41 @@ export class Storage extends Service {
       }
     );
   }
+
+  /**
+   * Get a reference to an HmacKey object.
+   * Note: this does not fetch the HMAC key's metadata. Use HmacKey#get() to
+   * retrieve and populate the metadata.
+   *
+   * @param {string} accessId The HMAC key's access ID.
```
frankyn: @jkwlui could you add an optional parameter for projectId? This would help in the case the user doesn't want to use the project defined by credentials or GOOGLE_CLOUD_PROJECT.

frankyn:
LGTM, thanks @jkwlui!

stephenplusplus (approved):

stephenplusplus:
@jkwlui a new linty error has appeared, seemingly after [I updated this PR to use the new common](3f00d92074db88cea9dbd01dd090125dbdec315b):

```
src/hmacKey.ts:323:7 - error TS2322: Type '{ delete: boolean; get: boolean; getMetadata: boolean; setMetadata: { reqOpts: { method: string; }; }; }' is not assignable to type 'Methods'.
  Property 'setMetadata' is incompatible with index signature.
    Type '{ reqOpts: { method: string; }; }' is not assignable to type 'boolean | { reqOpts?: CoreOptions | undefined; }'.
      Type '{ reqOpts: { method: string; }; }' is not assignable to type '{ reqOpts?: CoreOptions | undefined; }'.
        Types of property 'reqOpts' are incompatible.
          Type '{ method: string; }' is not assignable to type 'CoreOptions'.
            Types of property 'method' are incompatible.
              Type 'string' is not assignable to type '"GET" | "HEAD" | "POST" | "DELETE" | "PUT" | "CONNECT" | "OPTIONS" | "TRACE" | "PATCH" | undefined'.
323       methods,
          ~~~~~~~
  node_modules/@google-cloud/common/build/src/service-object.d.ts:54:5
    54     methods?: Methods;
           ~~~~~~~
    The expected type comes from property 'methods' which is declared here on type 'ServiceObjectConfig'
```

tenor

jkwlui:
The tests are blocked by #818

PR #752, opened 2019-06-24
docs: detailed clarification of 'getSignedUrl' method docs

### LINKED ISSUE #735: Update incorrect documentation for getSignedUrl
image

The documentation for `File.getSignedUrl` https://cloud.google.com/nodejs/docs/reference/storage/2.5.x/File#getSignedUrl

incorrectly says contentType is optional. It is not, after an hour of debugging I realized from stackoverflow, that you must set the content type when making the url, or the upload fails with a 403 every time.

AVaksman:
Fixes #735 

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2295,11 +2295,14 @@ class File extends ServiceObject<File> {
    *     'v2' or 'v4.
    * @param {string} [config.cname] The cname for this bucket, i.e.,
    *     "https://cdn.example.com".
-   * @param {string} [config.contentMd5] The MD5 digest value in base64. If you
-   *     provide this, the client must provide this HTTP header with this same
-   *     value in its request.
-   * @param {string} [config.contentType] If you provide this value, the client
-   *     must provide this HTTP header set to the same value.
+   * @param {string} [config.contentMd5] The MD5 digest value in base64. Just like
+   *     if you provide this, the client must provide this HTTP header with this same
+   *     value in its request, so to if this parameter is not provided here,
+   *     the client must not provide any value for this HTTP header in its request.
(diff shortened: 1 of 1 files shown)


src/file.ts
```diff
@@ -2295,11 +2295,14 @@ class File extends ServiceObject<File> {
...
-   * @param {string} [config.contentMd5] The MD5 digest value in base64. If you
-   *     provide this, the client must provide this HTTP header with this same
-   *     value in its request.
-   * @param {string} [config.contentType] If you provide this value, the client
-   *     must provide this HTTP header set to the same value.
+   * @param {string} [config.contentMd5] The MD5 digest value in base64. Just like
+   *     if you provide this, the client must provide this HTTP header with this same
+   *     value in its request, so to if this parameter is not provided here,
+   *     the client must not provide any value for this HTTP header in its request.
+   * @param {string} [config.contentType] Just like if you provide this, the client
```
bcoe: Should this read:

> If you provide this, the client must provided the corresponding HTTP header in its request. If no header is provided, the client must similarly not provide a header in the request.

AVaksman: I am trying to clarify the docs and in the same time to not exclude options.
The issue here is signature calculation. If these params are not provided, there values, for the sake of signature, default to `''`. Having this in mind, not providing this param in the method and providing same header with `undefined` or `''` value in the request will work, that's why I wrote 

> if this parameter is not provided here,
> the client must not provide any value for this HTTP header in its request

Hinting that providing an "empty" header is OK. (Since user may use third party modules which may work differently).

How does below sounds?

> If you provide this value, the client must provide a corresponding HTTP header set to the same value, if this value is not provided here, the client must similarly not provide this header in the request.

Issue #733, opened 2019-06-07
Timeout should be configurable as StorageOption


ildella:
There is no way to configure the timeout for a storage call. 
In 2.5 the [StorageOptions](https://cloud.google.com/nodejs/docs/reference/storage/2.5.x/global#StorageOptions) object that is passed as parameter on a new Storage() does not have any field for that. 

In the [python documentation](https://cloud.google.com/appengine/docs/standard/python/googlecloudstorageclient/retryparams_class) there is `urlfetch_timeout` for this. 

It would be great to have the same option in the nodejs library as well.

stephenplusplus:
@AVaksman this appears to have been addressed in https://github.com/googleapis/nodejs-common/pull/455. Is more work required, or is this feature live for new Storage downloads?

stephenplusplus:
Ping @AVaksman for clarity!

AVaksman:
Work in progress to expose timeouts to user on a high level.

moshekrup:
Any updates for this issue?
Does any way to define the timeout for the storage request?
Thanks

PR #756, opened 2019-06-27
feat: add request timeout configuration to storage object and its methods

AVaksman:
Towards #733

- [x] Tests and linter pass

This would be a first step to expose `request.timeout` configuration to storage functions.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -55,6 +55,11 @@ export interface StorageOptions extends GoogleAuthOptions {
    * Defaults to `www.googleapis.com`.
    */
   apiEndpoint?: string;
+  /**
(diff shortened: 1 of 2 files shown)


AVaksman:
Will it make seance to enable `timeout` on the `[service](https://github.com/googleapis/nodejs-common/blob/master/src/service.ts#L103)` and make it available to other client APIs?

jkwlui:
I feel like we should expose it at the @google-cloud/common level, as @AVaksman suggested. @stephenplusplus thought?

stephenplusplus:
Yes, the lower level, the better!

AVaksman:
Thank you, closing this PR and preparing changes for @google-cloud/common level.

PR #759, opened 2019-06-28
fix(docs): fix sample code in docs

bcoe:
addresses an open customer issue related to "headers" being pushed outside of the comment in sample code.


src/file.ts
```diff
@@ -1154,8 +1154,7 @@ class File extends ServiceObject<File> {
    * remoteFile.createReadStream()
    *   .on('error', function(err) {})
    *   .on('response', function(response) {
-   *     // Server connected and responded with the specified status and
-   * headers.
```
bcoe: the tiny fix.

PR #766, opened 2019-07-12
fix(typescript): make SetLabelOptions optional

### LINKED ISSUE #732: Storage: Support bucket.location_type
Location type is a new read-only field for buckets. This corresponds to internal bug 130187277.

callmehiphop:
test(bucket): location type system tests
docs(bucket): add verbiage about deprecated storage classes

---

Closes #732 

- [x] Tests and linter pass

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -2752,9 +2752,10 @@ class Bucket extends ServiceObject {
    *
    * @see [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
    *
-   * @param {string} storageClass The new storage class. (`multi_regional`,
-   *     `regional`, `standard`, `nearline`, `coldline`, or
-   *     `durable_reduced_availability`)
+   * @param {string} storageClass The new storage class. (`standard`,
+   *     `nearline`, `coldline`, or `durable_reduced_availability`).
+   *     **Note:** The legacy storage classes `multi_regional` and `regional`
+   *     have been deprecated.
    * @param {object} [options] Configuration options.
    * @param {string} [options.userProject] - The ID of the project which will be
    *     billed for the request.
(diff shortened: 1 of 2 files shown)


callmehiphop:
@JesseLovelace as mentioned previously, no code change was required. So I've added some e2e tests to make sure the field was populated correctly.

src/bucket.ts
```diff
@@ -2599,7 +2599,7 @@ class Bucket extends ServiceObject {
 
   setLabels(
     labels: Labels,
-    options: SetLabelsOptions
+    options?: SetLabelsOptions
```
JustinBeckwith: Looks like this is a `fix` if we're making actual product changes, no?

callmehiphop: Ahh, I forgot I tried to sneak that in 😱

src/bucket.ts
```diff
@@ -2752,9 +2752,10 @@ class Bucket extends ServiceObject {
    *
    * @see [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
    *
-   * @param {string} storageClass The new storage class. (`multi_regional`,
-   *     `regional`, `standard`, `nearline`, `coldline`, or
-   *     `durable_reduced_availability`)
+   * @param {string} storageClass The new storage class. (`standard`,
+   *     `nearline`, `coldline`, or `durable_reduced_availability`).
+   *     **Note:** The legacy storage classes `multi_regional` and `regional`
```
callmehiphop: @JesseLovelace I thought about remove the available storage classes in favor of just linking to the upstream docs. I could see the list going stale and getting overlooked. WDYT?

JesseLovelace: That's a valid concern, I'm good with either approach

PR #773, opened 2019-07-19
feat: allow removal of resumable upload cache

### LINKED ISSUE #217: Spurious precondition failures caused by ifGenerationMatch preserved across save() calls
## Blocked

- [ ] https://github.com/googleapis/gcs-resumable-upload/pull/243
- [ ] https://github.com/googleapis/gcs-resumable-upload/pull/244

---

I'm using `file('example', { generation: 0 }).save()` and discovered a strange behavior. I've managed to get resumable saves stuck in my config store, and as a result, when I subsequently invoke `file('example').save()`, the `ifGenerationMatch` header is still present on the response.

This is caused by the resumable upload code, which is turned on by default. I'm not sure whether the bug is here or in that library.

Steps to reproduce:

```js
const bucket = new Storage({ projectId }).bucket(bucketName)

await bucket.file('example').save('hello', { resumable: false })

try {
  await bucket.file('example', { generation: 0 }).save('hello')
} catch (e) {
  // Catch a precondition failure
}

// Saves a new version as expected.
await bucket.file('example').save('hello', { resumable: false })

// Expected to save a new version, but throws precondition failure
await bucket.file('example').save('hello')
```

jiren:
- bump gcs-resumable-upload to 2.2.3 to introduce resumable file cache cleanup
- `gcs-resumable-upload` resumable file cache cleanup method (`uploadInstance.deleteConfig()`) exposed for storage api.
- system test case for precondition failures

Fixes #217

- [x] Tests and linter pass


AVaksman:
Seems like unrelated lint issue in samples.
Resolves in PR #784

system-test/storage.ts
```diff
@@ -2824,6 +2826,33 @@ describe('storage', () => {
         files![1].metadata.generation
       );
     });
+
+    it('should throw an error Precondition Failed on overwrite with version 0, then save file with and withour resumable', done => {
```
jkwlui: nit: withour => without

system-test/storage.ts
```diff
@@ -2824,6 +2826,33 @@ describe('storage', () => {
...
+      const fileName = `test-${Date.now()}.txt`;
+      bucketWithVersioning
```
jkwlui: We can flatten this callback chain using Promise or async/await.

jiren: @jkwlui  Converted to async await

src/file.ts
```diff
@@ -3361,6 +3361,18 @@ class File extends ServiceObject<File> {
     });
   }
 
+  /**
```
stephenplusplus: Should we beef up this documentation, maybe explaining why this might be useful? Also, it could use an `@example` to match the other doc blocks.

src/file.ts
```diff
@@ -3361,6 +3361,18 @@ class File extends ServiceObject<File> {
...
+   * Delete failed resumable upload file cache.
+   */
+  deleteResumableCache() {
```
stephenplusplus: This is definitely nitpicky, but we've tried to keep the methods in these files alphabetically ordered, making it easier to jump in and find what you need when maintaining. (Private methods can stay at the bottom, defined after the public ones).

Likewise, the describe block in the test file for the method should match the order of the source file.

src/file.ts
jiren: @stephenplusplus Added example and updated doc

src/file.ts
jiren: Updated as per alphabetical order.

test/file.ts
```diff
@@ -2024,6 +2024,21 @@ describe('File', () => {
     });
   });
 
+  describe('deleteResumableCache', () => {
+    it('should delete resumable file upload cache', done => {
+      resumableUploadOverride = {
+        upload() {
+          return {
```
stephenplusplus: Could you add assertions here that the correct parameters were passed? The lines we'd be testing from the source code:

```js
bucket: this.bucket.name,
file: this.name,
generation: this.generation,
```

jiren: @stephenplusplus  added assertions.

stephenplusplus:
I'm pretty sure we don't need to keep this in limbo anymore. Merge time.

PR #774, opened 2019-07-22
fix(deps): drop dependency on through2

JustinBeckwith:


JustinBeckwith:
This may help shed some light on it:
https://github.com/googleapis/nodejs-firestore/pull/618#event-2337855128

> In particular, through2 overrides the default _destroy with an implementation that passes this through to the next stream.

bcoe:
@JustinBeckwith my vote would be that we close this issue, like the firestore issue; but that we start lobbying through the [tooling group](https://github.com/nodejs/tooling) to address some of the shortcomings in the streams API.

JustinBeckwith:
From what I can tell - it looks like this should magically start to work with Node 10 and up, so I'm totally supportive of just waiting this one out.  I don't like dependencies 🙃

PR #782, opened 2019-07-26
feat:  [Storage] Support UniformBucketLevelAccess

frankyn:
- [ ] Review from GCS eng team(@thanhld94).
- [x] Review from GCS cDPE team (@jkwlui).
- [x] Integration Tests.
- [x] Update Bucket Policy Only samples to use UBLA methods instead.
- [x] Sample Tests.
- [ ] Fix pending for backend bug: 134695162. Remove workarounds.


package.json
```diff
@@ -58,6 +58,7 @@
     "concat-stream": "^2.0.0",
     "date-and-time": "^0.7.0",
     "duplexify": "^3.5.0",
+    "express": "^4.17.1",
```
jkwlui: Don't think you intended to add this :)

samples/bucketLock.js
```diff
@@ -47,12 +47,12 @@ async function getRetentionPolicy(bucketName) {
   // Creates a client
   const storage = new Storage();
   const [metadata] = await storage.bucket(bucketName).getMetadata();
-  if (metadata.hasOwnProperty('retentionPolicy')) {
+  if ('retentionPolicy' in metadata) {
```
stephenplusplus: This isn't too conventional in JS. `metadata.hasOwnProperty('retentionPolicy')` is like the over-verbose, "I want to be 1000% sure" style, and ``retentionPolicy' in metadata` is like the cool kid, "I don't need semicolons" attitude. Just `if (metadata.retentionPolicy)` is the middle ground :)

frankyn: Thanks @stephenplusplus.

The reason why I stopped using metadata.hasOwnProprety was due to a linting issue.. that I can't recall at the moment. I'll revert.

AVaksman: The relevant PR #784 to make linter happy is merged.

bcoe:
@frankyn I think you should be able to rebase now and linting will work.

frankyn:
Thanks @AVaksman!

JesseLovelace:
Closing, see #903

PR #788, opened 2019-07-31
test: add benchmarking wrapper

jadekler:
This PR adds a gRPC server which wraps the storage client library. The methods specified in the proto are read and write, though the only implemented method in this PR is read.

The PR also:

- Adds an `apiScheme?` to `StorageOptions` (alongside `apiEndpoint?`)
- Causes validation: false to actually not do validation, instead of the current behavior in which validation: false always returns a validation error

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -1,0 +1,13 @@
+# benchwrapper
+
+benchwrapper is a lightweight gRPC server that wraps the storage library for
+bencharmking purposes.
+
+## Running
+
+```
+cd nodejs-storage
+npm install
+export STORAGE_EMULATOR_HOST=localhost:8080
+npm run benchwrapper -- --port 50051
+```
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "fix": "gts fix && eslint --fix '**/*.js'",
     "prepare": "npm run compile",
     "docs-test": "linkinator docs",
-    "predocs-test": "npm run docs"
+    "predocs-test": "npm run docs",
+    "benchwrapper": "node bin/benchwrapper.js"
   },
   "dependencies": {
     "@google-cloud/common": "^2.0.0",
@@ -74,6 +75,7 @@
   },
   "devDependencies": {
     "@google-cloud/pubsub": "^0.30.0",
+    "@grpc/proto-loader": "^0.5.1",
     "@types/compressible": "^2.0.0",
     "@types/concat-stream": "^1.6.0",
     "@types/configstore": "^4.0.0",
@@ -97,6 +99,7 @@
     "eslint-config-prettier": "^6.0.0",
     "eslint-plugin-node": "^9.0.0",
     "eslint-plugin-prettier": "^3.0.0",
+    "grpc": "^1.22.2",
     "gts": "^1.0.0",
     "intelli-espower-loader": "^1.0.1",
     "jsdoc": "^3.6.2",
@@ -114,6 +117,7 @@
     "source-map-support": "^0.5.6",
     "tmp": "^0.1.0",
     "typescript": "~3.5.0",
-    "uuid": "^3.1.0"
+    "uuid": "^3.1.0",
+    "yargs": "^13.3.0"
   }
 }
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -291,9 +291,12 @@ export class Storage extends Service {
    */
   constructor(options: StorageOptions = {}) {
     options.apiEndpoint = options.apiEndpoint || 'www.googleapis.com';
+    const url =
+      process.env.STORAGE_EMULATOR_HOST ||
+      `https://${options.apiEndpoint}/storage/v1`;
     const config = {
       apiEndpoint: options.apiEndpoint,
-      baseUrl: `https://${options.apiEndpoint}/storage/v1`,
+      baseUrl: url,
       projectIdRequired: false,
       scopes: [
         'https://www.googleapis.com/auth/iam',
--- a/bin/benchwrapper.js
+++ b/bin/benchwrapper.js
@@ -1,0 +1,55 @@
+const grpc = require('grpc');
+const protoLoader = require('@grpc/proto-loader');
+const {Storage} = require('../build/src');
+
+const argv = require('yargs')
+  .option('port', {
+    description: 'The port that the Node.js benchwrapper should run on.',
+    type: 'number',
+    demand: true,
+  })
+  .parse();
+
+const PROTO_PATH = __dirname + '/storage.proto';
+// Suggested options for similarity to existing grpc.load behavior.
+const packageDefinition = protoLoader.loadSync(PROTO_PATH, {
+  keepCase: true,
+  longs: String,
+  enums: String,
+  defaults: true,
+  oneofs: true,
+});
+const protoDescriptor = grpc.loadPackageDefinition(packageDefinition);
+const storageBenchWrapper = protoDescriptor.storage_bench;
+
+const storageClient = new Storage();
+
+function read(call, callback) {
+  const bucketName = call.request.bucketName;
+  const objectName = call.request.objectName;
+
+  storageClient
+    .bucket(bucketName)
+    .file(objectName)
+    .download({validation: false})
+    .then(function(data) {
+      // Do nothing with contents.
+    });
+
+  callback(null, null);
+}
+
+function write(call, callback) {
+  // TODO(deklerk)
+  callback(null, null);
+}
+
+const server = new grpc.Server();
+
+server.addService(storageBenchWrapper['StorageBenchWrapper']['service'], {
+  read: read,
+  write: write,
+});
+console.log('starting on localhost:' + argv.port);
+server.bind('0.0.0.0:' + argv.port, grpc.ServerCredentials.createInsecure());
+server.start();
--- a/bin/storage.proto
+++ b/bin/storage.proto
@@ -1,0 +1,45 @@
+// Copyright 2016 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicwable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package storage_bench;
+option java_multiple_files = true;
+option java_package = "com.benchwrapper";
+
+message ObjectRead{
+  // The bucket string identifier.
+  string bucketName = 1;
+  // The object/blob string identifier.
+  string objectName = 2;
+}
+
+message ObjectWrite{
+   // The bucket string identifier.
+  string bucketName = 1;
+  // The object/blob string identifiers.
+  string objectName = 2;
+  // The string containing the upload file path.
+  string destination = 3;
+}
+
+message EmptyResponse{
+}
+
+service StorageBenchWrapper{
+  // Performs an upload from a specific object.
+  rpc Write(ObjectWrite) returns (EmptyResponse) {}
+  // Read a specific object.
+  rpc Read(ObjectRead) returns (EmptyResponse){}
+}


jadekler:
cc @acmp1

package.json
```diff
@@ -92,6 +93,7 @@
     "@types/tmp": "0.1.0",
     "@types/uuid": "^3.4.4",
     "@types/xdg-basedir": "^2.0.0",
+    "@types/yargs": "^13.0.1",
```
JustinBeckwith: Trying hard to keep our dependencies down.  Do we strictly need yargs here for something simple only we're ever going to run?  @bcoe your call :)

src/bin/README.md
```diff
@@ -0,0 +1,13 @@
+# benchwrapper
```
JustinBeckwith: Thinking out loud - should we just move this to a `tools` directory and write it in JavaScript?

src/file.ts
```diff
@@ -1359,10 +1359,11 @@ class File extends ServiceObject<File> {
           md5: this.metadata.md5Hash,
         };
 
+        const doValidation = crc32c || md5;
```
JustinBeckwith: Interesting!  Can you explain this change?

src/storage.ts
```diff
@@ -55,6 +55,11 @@ export interface StorageOptions extends GoogleAuthOptions {
    * Defaults to `www.googleapis.com`.
    */
   apiEndpoint?: string;
+  /**
+   * The scheme of the service used to make requests.
+   * Defaults to `https`.
+   */
+  apiScheme?: string;
```
JustinBeckwith: So this is an issue.  We don't specifically expose this as a property in any other library, and I'd be worried to have a snowflake here.  

@broady what are your thoughts?

JustinBeckwith: Aside - if we *do* decide to do this, we would also need to update the jsdoc annotation

src/storage.ts
```diff
@@ -291,9 +296,10 @@ export class Storage extends Service {
    */
   constructor(options: StorageOptions = {}) {
     options.apiEndpoint = options.apiEndpoint || 'www.googleapis.com';
+    options.apiScheme = options.apiScheme || 'https';
     const config = {
       apiEndpoint: options.apiEndpoint,
-      baseUrl: `https://${options.apiEndpoint}/storage/v1`,
+      baseUrl: `${options.apiScheme}://${options.apiEndpoint}/storage/v1`,
```
JustinBeckwith: What is an example of a time I *wouldn't* want to use https?

tsconfig.json
```diff
@@ -2,10 +2,12 @@
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "noImplicitAny": false
```
JustinBeckwith: We can't turn off `noImplicitAny` - happy to help through whatever compiler issues it brings up, though honestly I think we should just stick this in a `tools` directory as JavaScript and call it a day

src/bin/README.md
jadekler: Sure - I don't mind either way. Would y'all discuss amongst yourself and let me know what's best? (I don't have a stake in the game)

src/file.ts
jadekler: It's briefly explained in the PR description, but to expand on it:

- If you have validation: crc32 or validation: md5, it will do the according validation
- If you have validation: false, it will return an error message saying "The downloaded data did not match the data from the server."

Result is there's no way to turn validation off, even though the options seem to suggest you can. So, this change makes false actually mean false by separating out "should i do validation" and "result of validation" into two separate vars.

src/storage.ts
jadekler: In other libraries - such as Go and Java - the host includes the scheme. In contrast, in Node, the host is everything after scheme up to port. Would you prefer me to make it conform in that way - by having scheme be part of `apiHost?`?

src/storage.ts
jadekler: Connecting to a local emulator.

src/file.ts
stephenplusplus: It seems like the code from master should work, so maybe it's something else happening? Here are the lines I'm referencing, in order inside of `createReadStream`:

```
let crc32c = true;
let md5 = false;

// ... later ...
} else if (options.validation === false) {
  crc32c = false;
}

// ... later ...
let failed = crc32c || md5;
// this should evaluate to `false`

if (crc32c /* (false) */ && hashes.crc32c) {
  // not run
}

if (md5 /* (false) */ && hashes.md5) {
  // not run
}

if (md5 /* (false) */ && !hashes.md5) {
  // not run
} else if (failed /* false */) {
  // not run
} else {
  // this runs:
  throughStream.end();
}
```

tsconfig.json
jadekler: Np! I get the following when I remove it:

```
$ npm run compile

> @google-cloud/storage@3.0.4 compile /Users/deklerk/workspace/nodejs-storage
> tsc -p . && cp src/bin/storage.proto build/src/bin/storage.proto

src/bin/benchwrapper.ts:44:15 - error TS7006: Parameter 'call' implicitly has an 'any' type.

44 function read(call, callback) {
                 ~~~~

src/bin/benchwrapper.ts:44:21 - error TS7006: Parameter 'callback' implicitly has an 'any' type.

44 function read(call, callback) {
                       ~~~~~~~~

src/bin/benchwrapper.ts:60:16 - error TS7006: Parameter 'call' implicitly has an 'any' type.

60 function write(call, callback) {
                  ~~~~

src/bin/benchwrapper.ts:60:22 - error TS7006: Parameter 'callback' implicitly has an 'any' type.

60 function write(call, callback) {
                        ~~~~~~~~

src/bin/benchwrapper.ts:67:19 - error TS7053: Element implicitly has an 'any' type because expression of type '"StorageBenchWrapper"' can't be used to index type 'GrpcObject | typeof Client | Message<object>'.
  Property 'StorageBenchWrapper' does not exist on type 'GrpcObject | typeof Client | Message<object>'.

67 server.addService(storageBenchWrapper['StorageBenchWrapper']['service'], {
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 10 errors.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @google-cloud/storage@3.0.4 compile: `tsc -p . && cp src/bin/storage.proto build/src/bin/storage.proto`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @google-cloud/storage@3.0.4 compile script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/deklerk/.npm/_logs/2019-07-31T22_25_00_451Z-debug.log
```

src/bin/README.md
JustinBeckwith: I would prefer to create the tools dir with pure JavaScript, but defer to @bcoe for the final  call

src/file.ts
jadekler: Womp womp, you guys are totally right. I had remembered that lost block as having an `if !failed` section. I suspect when I was running into it I must have not had verification: false.

src/storage.ts
JustinBeckwith: Eugh, gross.  Here's my thoughts:
- We should probably have the `apiEndpoint` accept an optional schema prefix
- We should have a utility function in `nodejs-common` that is responsible for parsing the apiEndpoint, determining if it has a schema (or port), then assembling the final thing
-  Roll this out more broadly to other nodejs libraries so it's consistent.

That having all be said - I would rather not do that.  If the scenario here is a local emulator - can we rely on an env var (which makes more sense IMO) to help us do . the right thing?  I think that's what we're doing in PubSub. 

https://github.com/googleapis/nodejs-pubsub/blob/master/src/pubsub.ts#L524

src/file.ts
jadekler: Reverted

src/storage.ts
JustinBeckwith: That makes a lot of sense.  I think we want to . rely on on an env var for that anyway (see above)

src/storage.ts
jadekler: Happy to go the env approach! We do the same in Go with Pub/Sub, Storage, Spanner, Firestore, and BigQuery IIRC. (`<SERVICE>_EMULATOR_HOST`) Will make the adjustment

src/bin/README.md
JustinBeckwith: Chatted with @bcoe - let's create a `bin` folder in the root of the repo, and put the `*.js` and `README` file inside of there.  We can just use JavaScript.

tsconfig.json
JustinBeckwith: Decided to not use TypeScript for this, so not a concern :) Should be able to just remove this

tsconfig.json
```diff
@@ -2,10 +2,12 @@
...
   },
   "include": [
     "src/*.ts",
+    "src/bin/*.ts",
```
JustinBeckwith: Can get rid of this now

package.json
bcoe: As long as it's a development dependency, I like that it creates a self documenting CLI; I think anything for the benchmarker should be installed with `--save-dev`.

src/bin/README.md
bcoe: add to that, each dependency should be saved with `--save-dev`, so that they're in the `devDependencies` stanza rather than the `dependencies` stanza.

I still think it would be good too add a script tag:

```
"script": {
  "test": "...",
   "benchmark": "./bin/benchmark.js" 
}
```

The default exit of `benchmark.js` could then print for `.usage`:

```
npm run benchmark -- --port=[port to connect to]
```

src/storage.ts
bcoe: any reason we can't just grab `URL.parse(apiEndpoint)` and fetch the scheme off that.

tsconfig.json
jadekler: Done

tsconfig.json
jadekler: Done

jadekler:
(still working on getting this ready - will ping for review)

src/storage.ts
jadekler: I personally prefer the env var: it's simpler for the benchwrapper (don't need --emulatorHost), it's less complicated than string parsing, and conforms with how Java and Go work.

jadekler: Done

jadekler:
I think I've addressed all review comments, but it's a bit hard to tell with github's UI heh. Would y'all mind taking another look?

src/storage.ts
```diff
@@ -290,10 +290,14 @@ export class Storage extends Service {
    * @param {StorageOptions} [options] Configuration options.
    */
   constructor(options: StorageOptions = {}) {
+    let url = `https://${options.apiEndpoint}/storage/v1`
+    if (process.env.STORAGE_EMULATOR_HOST != undefined) {
```
bcoe: I think we can pass this in to the constructor, like so:

```js
const storageClient = new Storage({
  baseUrl: process.env.STORAGE_EMULATOR_HOST
})
```

tsconfig.json
bcoe: I think we probably don't need any changes to `tsconfig.json` any longer.

bin/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+benchwrapper is a lightweight gRPC server that wraps the storage library for
+bencharmking purposes.
+
+## Running
+
+```
+cd nodejs-storage
+npm install
+npm run compile
```
JustinBeckwith: Don't need to `npm run compile` anymore :)

src/storage.ts
```diff
@@ -290,10 +290,14 @@ export class Storage extends Service {
    * @param {StorageOptions} [options] Configuration options.
    */
   constructor(options: StorageOptions = {}) {
+    let url = `https://${options.apiEndpoint}/storage/v1`
```
JustinBeckwith: This is using `options.apiEndpoint` before it's set.  We should move up this line first:
```
 options.apiEndpoint = options.apiEndpoint || 'www.googleapis.com';

File: src/storage.ts (hunk 8 above)
Comment (JustinBeckwith) [reply src/storage.ts]: I don't believe we expose `baseUrl` as a configurable property (I could be wrong)

Comment (JustinBeckwith) [inline src/storage.ts]: Instead of this:

```js
const url = process.env.STORAGE_EMULATOR_HOST || `https://${options.apiEndpoint}/storage/v1`;
```

tsconfig.json
JustinBeckwith: We can delete this now.

jadekler: Done

src/storage.ts
jadekler: Done

src/storage.ts
jadekler: Done

bin/README.md
jadekler: Done

JustinBeckwith:
Just need to run 'npm run fix' and you should be good here

jadekler:
> Just need to run 'npm run fix' and you should be good here

Uhhhh... so I got [this](https://gist.github.com/jadekler/63192582aa14dca4dfe08cde5726c6a3) when I ran that:

```
   1:22  error    "grpc" is extraneous                           node/no-extraneous-require
   2:29  error    "@grpc/proto-loader" is extraneous             node/no-extraneous-require
```

But... they're not extraneous. They're both used. What's the deal with that? o.O

```
  36:13  error    'contents' is assigned a value but never used  no-unused-vars
```

Done.

```
  44:3   warning  Unexpected 'todo' comment                      no-warning-comments
```

Why is this a warning? Can I ignore this?

```
   29:29  error  "@google-cloud/storage" is not found  node/no-missing-require
```

There were many of these, and I can't tell if they're meaningful or not.

-------------

It also seems to have done a bunch of linting, which I have committed and pushed.

JustinBeckwith:
See this is a good error!  `grpc` and `@grpc/proto-loader` are available as transitive dependencies, but we should be explicit about the version we want.  We can do that by running:

```
npm install --save-dev grpc @grpc/proto-loader
```

And then committing the resulting changes to the `package.json`.  `extraneous` is a *terrible* word for what it's trying to express, FWIW.

jadekler:
> See this is a good error! `grpc` and `@grpc/proto-loader` are available as transitive dependencies, but we should be explicit about the version we want. We can do that by running:
> 
> ```
> npm install --save-dev grpc @grpc/proto-loader
> ```

Done.

> And then committing the resulting changes to the `package.json`. `extraneous` is a _terrible_ word for what it's trying to express, FWIW.

Hahaha indeed.

JustinBeckwith:
@bcoe you ok to merge here?  Thinking the system test failure is unrelated/flaky

bcoe:
@JustinBeckwith I'm :+1: on landing this, I've created a `p0` tracking issue to figure out why system tests are failing on master.

PR #791, opened 2019-08-01
docs(samples): Add get bucket metadata sample

frankyn:
cc: @JesseLovelace 
- [x] Tests and linter pass


samples/buckets.js
```diff
@@ -67,6 +67,40 @@ async function listBuckets() {
   // [END storage_list_buckets]
 }
 
+async function getBucketMetadata(bucketName) {
```
JustinBeckwith: Could I trouble you to put this sample in its own file?  Then you can add metadata that gets picked up automatically.  See https://github.com/googleapis/nodejs-bigquery/blob/master/samples/getModel.js#L19 for an example!

samples/buckets.js
```diff
@@ -67,6 +67,40 @@ async function listBuckets() {
...
+
+  /**
+   * TODO(developer): Uncomment the following line before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+
+  // Get Bucket Metadata
+  const [metadata] = await storage.bucket(bucketName).getMetadata();
+
+  console.log(`BucketName: ${metadata.name}`);
```
JustinBeckwith: Could we really simplify this? 

```js
for (let [key, value] of Object.entries(metadata)) {
  console.log(`${key}: ${value}`);
}
```

samples/buckets.js
frankyn: Sure thing! Thanks for the example.

samples/buckets.js
frankyn: YES!! Thanks for the tip.

PR #799, opened 2019-08-06
chore: specify readable-stream@3 to support asyncIterator

### LINKED ISSUE #765: Support readable[Symbol.asyncIterator]() Readble returned by createReadStream()
`createReadStream()` is typed as creating a `Readable`, however it doesn't fulfill the interface as of  node 10.16, as there is no `readable[Symbol.asyncIterator]()`. 

https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_readable_symbol_asynciterator

AVaksman:
Fixes #765

- [x] Tests and linter pass

Specify `readable-stream@3` in `package.json`
`readable-stream@3` is needed for `through2` to support `readable[Symbol.asyncIterator]()`
Currently `throught2` is OK with ether one.

https://github.com/rvagg/through2/blob/70ad45f185f1c62b5d002f6e37a277501eb5da67/package.json#L21-L23

And among requirements of other modules gets  `deduped` to version 2.


package.json
```diff
@@ -68,6 +68,7 @@
     "onetime": "^5.1.0",
     "p-limit": "^2.2.0",
     "pumpify": "^2.0.0",
+    "readable-stream": "^3.4.0",
```
JustinBeckwith: I have been playing whack-a-mole trying to avoid bringing that package in, and instead rely on it being in node.js core.  What versions of node.js don't currently work with async iterators?

AVaksman: `readable[Symbol.asyncIterator]()` was introduced in Node 10.

Since it all originates from using `through2` ([here](https://github.com/AVaksman/nodejs-storage/blob/support_asyncIterator/src/file.ts#L1194) and [here](https://github.com/AVaksman/nodejs-storage/blob/support_asyncIterator/src/file.ts#L1722)) I am looking in to replacing `through2` with `stream.PassThrough` and dropping `through2` all together.

stephenplusplus:
Sorry if I missed something, but why not keep through2 around? Seems safer than writing our own implementation, which will need testing and need to be maintained, etc.

package.json
AVaksman: Removed `readable-stream` extra dependency.
Adopt `PassThrough` (with override for `destroy` for Node 8 to mimic Node 10).
Drop `through2` dependency in favor of native `PassThrough`.
WDYT?

AVaksman:
> Sorry if I missed something, but why not keep through2 around? Seems safer than writing our own implementation, which will need testing and need to be maintained, etc.

In order to support `asyncIterator` `through2` needs to use `readable-sream@3`, however, since `through2` is OK with ether one ([link](https://github.com/rvagg/through2/blob/70ad45f185f1c62b5d002f6e37a277501eb5da67/package.json#L21-L23)), among requirements of other modules through2 ends up "deduped" to version 2.

I offered to specify `readable-stream@3` dependency here in `package.json`, this way `through2` will end up "deduped" with version 3.

@JustinBeckwith  would prefer to rely on node.js core.

PassThrough.destroy of Node 8 does not emit 'close' event, where as Node 10 and `through2` do. Just for that purpose I override `.destroy` method mimicking Node 10 logic.

stephenplusplus:
Should we see if through2 wants to use readable-stream@3 only?

@JustinBeckwith IMO, streams really aren't a good place to roll our own anything. There are quirks and compatibility issues everywhere, just look at any of the code of a popular stream-helper module. I would feel more comfortable using readable-stream, and frankly admitting defeat on our ridding efforts-- even if we ditch it now, eventually, we're gonna end up with it somewhere, somehow anyway.

AVaksman:
For reference

through2 `destroy` implementation

https://github.com/rvagg/through2/blob/70ad45f185f1c62b5d002f6e37a277501eb5da67/through2.js#L11-L21

Node 10 `destroy` implementation

https://github.com/nodejs/node/blob/b884ceb518d5a63c1d5a85ca9b7e87b04e02ab4b/lib/internal/streams/destroy.js#L4-L56

AVaksman:
> Should we see if through2 wants to use readable-stream@3 only?
👍 

Especially as it clearly says in their documentation

>  **As of `through2@3.x`, readable-stream@3 is being used and is not compatible with older versions > of Node.js.**

stephenplusplus:
PR sent to through2: https://github.com/rvagg/through2/pull/99/files

AVaksman:
> PR sent to through2: https://github.com/rvagg/through2/pull/99/files

Thank you

stephenplusplus:
Just a heads up we got a nice reply on the PR at through2: https://github.com/rvagg/through2/pull/99. Please share any thoughts over there!

AVaksman:
> Just a heads up we got a nice reply on the PR at through2: [rvagg/through2#99](https://github.com/rvagg/through2/pull/99). Please share any thoughts over there!

Unfortunately `stream.destroy()` starts emitting `close` only since Node 10.0.0 ([5e3f51648e](https://github.com/nodejs/node/commit/5e3f51648e#diff-0da8feb832c813299da638cf7fdc65c9R47) and corrected in [a7c25b7d42](https://github.com/nodejs/node/commit/a7c25b7d42b79878548b9d660af0195e7318fc33#diff-0da8feb832c813299da638cf7fdc65c9R34))

Options as I see:

1. Add `readable-stream@3` directly here - not really desirable
2. Accept some sort of custom stream for Node 8 usage
3. Wait until Node 8 support is discontinued and use native `PassThrough` without customization
If option 3 is desirable, than options 1 and 2 are temporary as it is.

WDYT?

stephenplusplus:
I don't mind 1 🤷‍♂

package.json
JustinBeckwith: I defer this one to @bcoe

bag-man:
Thanks guys!

PR #812, opened 2019-08-14
test: add test for x-goog-api-client header

bcoe:
adds a test confirming that `x-goog-api-client` header is being populated for metrics.


bin/benchwrapper.js
```diff
@@ -32,7 +48,7 @@ function read(call, callback) {
     .bucket(bucketName)
     .file(objectName)
     .download({validation: false})
-    .then(function(data) {
+    .then(function() {
```
JustinBeckwith: ... but why have the `.then` at all?

package.json
```diff
@@ -106,7 +106,7 @@
     "jsdoc-fresh": "^1.0.1",
     "linkinator": "^1.5.0",
     "mocha": "^6.0.0",
-    "nock": "^10.0.1",
+    "nock": "^11.0.0-beta.31",
```
JustinBeckwith: `11.0.0` is out :)

test/headers.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('headers', () => {
+  before(() => {
+    nock.disableNetConnect();
+  });
+  it('populates x-goog-api-client header', async () => {
+    const storage = new Storage({
+      projectId: 'foo',
+    });
+    const bucket = storage.bucket('foo-bucket');
+    const metadata = nock('http://metadata.google.internal.')
```
JustinBeckwith: This is going to be a problem.  On my machine, I am logged in with the `Cloud SDK`.  I also have a `GOOGLE_APPLICATION_CREDENTIALS` var set.  With existing logins created ... the auth library will never get to the metadata service to try talking to it.  

This test would fail due to the `disableNetconnect` above, and I'd be super confused why my unit tests were failing.

Instead - I think we should use proxyquire, and stub out the `getClient` call, or the `request` call, or something like that.

I'm open to other ideas, but here's what I've done:
https://github.com/JustinBeckwith/gcbuild/blob/master/test/test.ts#L18

test/headers.ts
```diff
@@ -0,0 +1,50 @@
...
+            this.req.headers['x-goog-api-client'][0]
+          )
+        );
+        return {};
+      });
+    await bucket.create();
+    metadata.done();
+    req.done();
+  });
+  after(() => {
```
JustinBeckwith: Don't think this is strictly required

bcoe: I think in the name of cleanup it's the right call, otherwise we have net connections disabled on subsequent tests.

JustinBeckwith: (I don't feel strongly about this) the idea here is that we should *never* have unit tests that are making outbound HTTP requests.  We want to save that for the sample/system tests.

test/headers.ts
bcoe: I see no reason to not just use `proxyquire`.

PR #818, opened 2019-08-21
fix(ts): fix @types/nock to 10.0.0

jkwlui:
Starting from 10.0.3, @types/nock started using nock as default export,
even though the library uses module export. This broke our typescript
compilation.

Fixing the version of @types/nock to 10.0.0 fixed the problem.

# TODO
- [ ] Report issue at DefinitelyTyped.


jkwlui:
nope - this is not `@types/nock`

PR #834, opened 2019-08-28
docs(samples): add set file metadata sample

### LINKED ISSUE #222: Setting metadata seems really difficult to reason with.

praveenqlogic01:
Fixes #222  (it's a good idea to open an issue first for discussion)

Added a sample file which explain user how to add user metadata and context metadata to file

### DIFF
--- a/samples/fileSetMetadata.js
+++ b/samples/fileSetMetadata.js
@@ -35,22 +35,22 @@ async function main(bucketName = 'my-bucket', filename = 'file.txt') {
   // const bucketName = 'Name of a bucket, e.g. my-bucket';
   // const filename = 'File to access, e.g. file.txt';
 
-  // user define custom metadata
-  const userMetadata = {
-    my: 'custom',
-    properties: 'go here',
-  };
-
-  //metadata Properties
-  const contextMetadata = {
-    contentType: 'application/x-font-ttf',
-    metadata: userMetadata,
-  };
-
   const [metadata] = await storage
     .bucket(bucketName)
     .file(filename)
-    .setMetadata(contextMetadata);
+    .setMetadata({
+      // Predefinded metadata for server e.g. 'cacheControl', 'contentDisposition',
+      // 'contentEncoding', 'contentEncoding', 'contentLanguage', 'contentType'
+      contentDisposition: 'attachment; filename*=utf-8\'\'"anotherImage.jpg"',
+      contentType: 'image/jpeg',
+
+      // Note or actionable items for user e.g. uniqueId,
+      // object description or other useful information.
+      metadata: {
+        description: 'file description...',
+        modified: '1900-01-01',
+      },
+    });
 
   console.log(metadata);
 
--- a/samples/fileSetMetadata.js
+++ b/samples/fileSetMetadata.js
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+// sample-metadata:
+//   title: Storage Set File Metadata.
+//   description: Set file metadata.
+//   usage: node fileSetMetadata.js <BUCKET_NAME> <FILE_NAME>
+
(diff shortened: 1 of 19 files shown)


samples/fileSetMetadata.js
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // const filename = 'File to access, e.g. file.txt';
+
+  // user define custom metadata
+  const userMetadata = {
+    my: 'custom',
+    properties: 'go here',
```
bcoe: would properties tend to be an object?

praveenqlogic01: No, taken the example from [here](https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts#L744)

samples/system-test/files.test.js
```diff
@@ -250,6 +250,21 @@ it('should get metadata for a file', () => {
...
+it('should set metadata for a file', () => {
+  // set in sample
+  const userMetadata = {
+    my: 'custom',
+    properties: 'go here',
+  };
+  const output = execSync(
+    `node fileSetMetadata.js ${bucketName} ${copiedFileName}`
+  );
+  const metadata = eval(`(${output})`);
```
bcoe: I don't love the use of `eval` here, could we just use `JSON.parse()`, or `assert.match`?

samples/fileSetMetadata.js
bcoe: 👍 what sort of metadata might I want to set on a file; even though this is drawn from an existing sample, I wonder if we could make the sample demonstrate a concrete usage of what someone might do with metadata.

samples/system-test/files.test.js
praveenqlogic01: done

samples/fileSetMetadata.js
praveenqlogic01: @bcoe Hi, i refactor my code and elaborate the use of `custom metadata` and `context metadata`,
hope you like it.

samples/fileSetMetadata.js
```diff
@@ -0,0 +1,60 @@
...
+  // Creates a client
+  const storage = new Storage();
+
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // const filename = 'File to access, e.g. file.txt';
+
+  const [metadata] = await storage
```
JustinBeckwith: If a user copies and pastes the sample between the region tags here, their code won't run because of the top level `await`.  We should wrap this specific piece of code in side of an async function, then call it inside the region tag:

```js
async function setFileMetadata() {
  const [metadata] = await storage....
  ....
}
setFileMetadata();
```

I know this smells, but it's pretty frequent feedback and part of the style guide.  Adding @bcoe in case he has an idea of how to catch this with a lint rule or somethin'

bcoe: it's hard, because it's not _technically_ a top level `await` which I'm sure linting could catch, it's just a top-level await in the context of the region tags.

So, for what it's worth, I saw that top-level await just landed in V8, so I wouldn't be shocked if we have it upstreamed to Node 12/13 ... the wrapping function might become less necessary over the next year.

samples/fileSetMetadata.js
```diff
@@ -0,0 +1,62 @@
...
+        // object description or other useful information.
+        metadata: {
+          description: 'file description...',
+          modified: '1900-01-01',
+        },
+      });
+
+    console.log(metadata);
+  }
+  // [END storage_set_metadata]
```
AVaksman: Please move the `END` region tag down till after `setFileMetadata();` to ensure that copied sample code (between the region tags) will run as is.

praveenqlogic01: done

samples/system-test/files.test.js
```diff
@@ -250,6 +250,23 @@ it('should get metadata for a file', () => {
...
+  // used in sample
+  const userMetadata = {
+    description: 'file description...',
+    modified: '1900-01-01',
+  };
+  const output = execSync(
+    `node fileSetMetadata.js ${bucketName} ${copiedFileName}`
+  );
+
+  assert.match(
```
JustinBeckwith: We could simplify this to:

```js
assert.includes(output, `description: '${userMetadata.description}'`);
```

:)

PR #837, opened 2019-08-28
revert: remove unsupported keepAcl param from file#copy

### LINKED ISSUE #836: Doc outdated: option "keepAcl" in file#copy is not in action
According to https://github.com/googleapis/nodejs-storage/issues/21 that feature is not supported. But it appears in the documentation.

https://github.com/googleapis/nodejs-storage/blob/840eb6a2f639a108b85985fef5def5170aa172f3/src/file.ts#L862

AVaksman:
Fixes #836

- [x] Tests and linter pass

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -295,7 +295,6 @@ export interface FileOptions {
 
 export interface CopyOptions {
   destinationKmsKeyName?: string;
-  keepAcl?: string;
   predefinedAcl?: string;
   token?: string;
   userProject?: string;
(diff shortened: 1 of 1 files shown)


AVaksman:
> I might be wrong, but this looks like a TypeScript breaking change?

https://github.com/googleapis/nodejs-storage/pull/165#discussion_r318775074

callmehiphop:
I see, maybe the commit title should be revert instead of docs then.

stephenplusplus:
Sorry, Internet :(

PR #841, opened 2019-09-05
fix: add warning for unsupported keepAcl param in file#copy

AVaksman:
Related to #836

- [x] Tests and linter pass

As per https://github.com/googleapis/nodejs-storage/pull/839#issuecomment-528378478


src/file.ts
```diff
@@ -984,6 +987,13 @@ class File extends ServiceObject<File> {
       options = optionsOrCallback;
     }
 
+    if (options.hasOwnProperty('keepAcl')) {
+      // TODO: remove keepAcl from interface in next major.
+      console.warn(
```
stephenplusplus: This is scary to me, because we've always considered our chance to communicate dev-only messages is pre-installation via the changelog, but not at runtime. That could potentially interrupt their users, since we are commonly a base layer in more complex applications.

stephenplusplus: cc @bcoe @callmehiphop for thoughts!

callmehiphop: I might be misremembering, but I seem to recall using the console to warn users about deprecated methods in the auth client as well (paging @JustinBeckwith). I'm ok with this, but think we should take a hard stance of how we communicate this kind of change across the board.

bcoe: I've been known to use this approach in yargs:

https://github.com/yargs/yargs/blob/5d7ad989a851398587a0349cdd15344769b4cd79/yargs.js#L1035

And haven't received many complaints; my general thinking has been. 

* it's less disruptive to start by warning folks than throwing.
* it draws attention to a change that _will_ eventually be breaking.
* Node itself has done this for a lot of features, e.g., `Buffer`, so the community is accustomed to it.

In this case, my gut was that only a small percentage of folks would be explicitly setting `keepAcl` so it's probably fairly low impact.

AVaksman: Done, thanks
@JustinBeckwith could I trouble you to look over the switch to `process.emitWarning` once per process (last commit).

JustinBeckwith:
LGTM - only ask is that we land this as a `fix` since it has a material change to the behavior of the API

PR #848, opened 2019-09-10
chore: better instructions for running benchwrapper

jadekler:


bin/README.md
```diff
@@ -6,8 +6,9 @@ bencharmking purposes.
 ## Running
 
 ```
+npm install -g typescript
```
callmehiphop: Why is a global install of TypeScript needed to run the benchmarks?

jadekler: It's needed for `npm install` to work.

jkwlui: It shouldn't be needed.. What error did you receive?

jadekler: Tried to repro but couldn't. Closing.

Issue #709, opened 2019-05-13
CONTENT_DOWNLOAD_MISMATCH with successful file download


taiyokato:
#### Environment details
* OS: Windows 10 Pro 1809
* Node.js version: v10.15.3 LTS
* npm version: 6.4.1
* `@google-cloud/storage` version: 2.5.0

#### Steps to reproduce
0. Prepare a bucket with Cloud KMS Customer managed Key encryption. All files are private.
1. Upload an JPEG image file using: 
```javascript
function UploadFile() {
    storage.bucket("somebucket").upload("somefolder/someimage.jpg", {
        gzip: true,
        destination: "somefolder/someimage.jpg",
    }).then(results => {
        console.log("upload OK");
    })
    .catch(err => {
        console.error("Error: ", err);
    });
}
UploadFile()
// Upload works fine.
```
2. Try downloading the img file using: 
```javascript
function DownloadFile() {    
    const file = storage.bucket("somebucket").file("somefolder/someimage.jpg");
    
    file.download({
        destination: "someimage.jpg",
        // validation: false, // Why even disable validation?
    }).then(res => {
        console.log("DL OK");
    }).catch(err => {
        console.error(err); // Throws CONTENT_DOWNLOAD_MISMATCH
    })
}
DownloadFile()
```
3. ***Correct decompressed file is downloaded*** AND exception is thrown: `code=CONTENT_DOWNLOAD_MISMATCH message=The downloaded data did not match the data from the server. To be sure the content is the same, you should download the file again.`

I've read through [Issue 566](https://github.com/googleapis/nodejs-storage/issues/566), but seems like not a solution.

`storage.bucket("somebucket").file("somefolder/someimage.jpg").download({validation: false});` works, but there's no reason to or should disable validation.

To make sure if the hashes are actually a mismatch, I ran a local md5sum check on the downloaded and original image files.

```
$ md5sum downloaded.jpg original.jpg
a045a2e8e6b8d84aa8a319bcdba05419  downloaded.jpg
a045a2e8e6b8d84aa8a319bcdba05419  original.jpg
```

Downloaded file is a match to the original file.

BTW, This problem doesn't happen if the image is not gzipped on upload.

Thanks

jiren:
Validation error because of crc32c check failure. 
https://github.com/googleapis/nodejs-storage/blob/61eeb64a6bca194361aacf2a312d27d0e6d63b35/src/file.ts#L1372

[...]

PR #850, opened 2019-09-11
feat: add flag to allow disabling auto decompression by client

### LINKED ISSUE #827: Allow disabling automatic decompression when downloading files
When sending files downloaded from GCS back to the user, it would be nice if there existed an option to disable decompressing files that are compressed at rest. Currently we end up decompressing on download (well, this library does) and compressing on the way out to clients again.

A flag in `file.download()` that can disable `gzip` compression (I absolutely think decompressing should be the default).

AVaksman:
Fixes #827
- [x] Tests and linter pass


src/file.ts
```diff
@@ -359,6 +359,7 @@ export interface CreateReadStreamOptions {
   validation?: 'md5' | 'crc32c' | false | true;
   start?: number;
   end?: number;
+  returnCompressed?: boolean;
```
stephenplusplus: How about `decompress`, default `true`?

AVaksman: done

src/file.ts
```diff
@@ -1113,8 +1113,8 @@ class File extends ServiceObject<File> {
    *     NOTE: Byte ranges are inclusive; that is, `options.start = 0` and
    *     `options.end = 999` represent the first 1000 bytes in a file or object.
    *     NOTE: when specifying a byte range, data integrity is not available.
-   * @property {booblean} [returnCompressed] Disable auto decompression of the
-   *     received data. By default this option set to `false`.
+   * @property {booblean} [decompress] Disable auto decompression of the
```
stephenplusplus: I think if you put `[decompress=true]`, that'll tip off the docs site that the default value is `true`. Also, I think it's just called a `boolean` 😅.

AVaksman: Done! 😹

JustinBeckwith:
@frankyn @jkwlui mind taking a look?

src/file.ts
```diff
@@ -1120,6 +1121,10 @@ class File extends ServiceObject<File> {
    *     NOTE: Byte ranges are inclusive; that is, `options.start = 0` and
    *     `options.end = 999` represent the first 1000 bytes in a file or object.
    *     NOTE: when specifying a byte range, data integrity is not available.
+   * @property {boolean} [decompress=true] Disable auto decompression of the
+   *     received data. By default this option set to `true`.
```
frankyn: this option *is* set to `true`

src/file.ts
```diff
@@ -1120,6 +1121,10 @@ class File extends ServiceObject<File> {
    *     NOTE: Byte ranges are inclusive; that is, `options.start = 0` and
    *     `options.end = 999` represent the first 1000 bytes in a file or object.
    *     NOTE: when specifying a byte range, data integrity is not available.
+   * @property {boolean} [decompress=true] Disable auto decompression of the
```
frankyn: Is this from the perspective of the GCS service?

src/file.ts
AVaksman: Currently nodejs-storage client _does_ decompress by default if data received from the server is compressed 
i. e. response contains header `content-encoding: 'gzip'`

`decompress` property seemed a better sounding name vs originally offered `returnCompressed` as per suggestion [above](https://github.com/googleapis/nodejs-storage/pull/850#discussion_r323419992)

src/file.ts
AVaksman: No, nodejs-sotage API always requests that the data should be served as is (compressed if applicable) i. e. sending appropriate [headers](https://github.com/googleapis/nodejs-storage/blob/b2660d8c7d68058634a20ae923afe6a44abe6b4e/src/file.ts#L1239-L1242)
```ts
const headers = {
  'Accept-Encoding': 'gzip',
  'Cache-Control': 'no-store',
} as Headers;
```

This flag is only from the perspective of nodejs-storage client to allow this library to not transcode the data if received from the server in compressed form and return to user compressed..

src/file.ts
stephenplusplus: There is a missing "is" in the sentence: "By default this option set to `true`."

AVaksman: > There is a missing "is" in the sentence: "By default this option set to `true`."

fixed, thanks

stephenplusplus:
@AVaksman this also fixes #709, right?

src/file.ts
stephenplusplus: @frankyn ping!

test/file.ts
```diff
@@ -1203,6 +1203,17 @@ describe('File', () => {
           .resume();
       });
 
+      it('should not gunzip the response if "retunCompressed: true" is passed', done => {
```
jkwlui: nit: returnCompressed => decompress

AVaksman:
> @AVaksman this also fixes #709, right?

@stephenplusplus No, the bottom line of #709 is that server does not respect `cache-control: no transform` header from JSON API (only XML) and does not serve doubly compressed data (for example `jpg` + `gzip`), but rather automatically removes the top level compression (`gzip`) and serves the data singly compressed (just `.jpg`).

This PR cover the case when user is downloading `txt.gzip` file. Upon receiving `content-encoding: gzip`, the library would auto decompress the data and return to user (just `.txt`). Here we allowing the user to request from the library not to auto-decompress the data upon receipt but rather pass it to the user as is (`.txt.gzip`).

test/file.ts
AVaksman: Thanks, fixed

PR #874, opened 2019-10-03
fix: file#move do not delete origin file if same as destination

AVaksman:

Currently if user passes same `bucket` object, `file`object, or `fileName` string as the original file, the client will attempt to "copy" (essentially leaving the file in place) and subsequently delete the origin/destination file.

### DIFF
--- a/test/file.ts
+++ b/test/file.ts
@@ -3461,15 +3461,9 @@ describe('File', () => {
       });
 
       it('should not delete the destination is same as origin', done => {
-        // sinon.stub(file, 'request').callsFake((callback: Function) => {
-        //   callback(null, null);
-        // })
         file.request = (config: {}, callback: Function) => {
           callback(null, {});
         };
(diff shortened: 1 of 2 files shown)


jkwlui:
Sounds like what we really should do here is to error out when the user tries to copy/move to the same destination, unless I'm missing something here?

AVaksman:
IMHO I am not sure passing same destination on copy/move deserves an error (would be a breaking change). WDYT?

jkwlui:
Yeah I would try to avoid a breaking change here as well. The draft looks good, promote it to a PR?

test/file.ts
```diff
@@ -3459,11 +3460,40 @@ describe('File', () => {
         });
       });
 
+      it('should not delete the destination is same as origin', done => {
+        // sinon.stub(file, 'request').callsFake((callback: Function) => {
```
jkwlui: Remove commented code.

AVaksman: Thanks, done

PR #876, opened 2019-10-04
feat(bucket): add enableLogging method

### LINKED ISSUE #23: logging endpoint support
_From @chrishiestand on June 5, 2017 19:23_

I need to be able to programmatically use the storage logging endpoints: https://cloud.google.com/storage/docs/access-logs

(It would be so much nicer to be able to make an API call to get storage usage per bucket rather than create a new bucket, add logging, parse logfiles and so on - but that is a GCP issue, not a google-cloud-node issue.)

I see you are already working on custom endpoint support: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1630

I don't know how far away custom endpoints are, but in any case it would be very useful to me to add google-cloud-node storage logging endpoint support in one form or another.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2359_

stephenplusplus:
Fixes #23 

Hello! This adds the `bucket.enableLogging()` shortcut method as discussed in #23.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1833,9 +1833,7 @@ class Bucket extends ServiceObject {
    *   }
    * });
    *
-   * //-
-   * // Optionally, provide a destination bucket.
-   * //-
+   * @example <caption>Optionally, provide a destination bucket.</caption>
    * const config = {
    *   prefix: 'log',
    *   bucket: 'destination-bucket'
@@ -1843,9 +1841,8 @@ class Bucket extends ServiceObject {
    *
    * bucket.enableLogging(config, function(err, apiResponse) {});
    *
-   * //-
-   * // If the callback is omitted, we'll return a Promise.
-   * //-
+   * @example <caption>If the callback is omitted, we'll return a Promise.
+   *     </caption>
    * bucket.enableLogging(config).then(function(data) {
    *   const apiResponse = data[0];
    * });
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1870,22 +1870,25 @@ class Bucket extends ServiceObject {
     }
 
     (async () => {
+      let setMetadataResponse;
+
       try {
         await this.acl.add({
           entity: 'group-cloud-storage-analytics@google.com',
           role: 'WRITER',
         });
-        const [setMetadataResponse] = await this.setMetadata({
+        [setMetadataResponse] = await this.setMetadata({
           logging: {
             logBucket,
             logObjectPrefix: config.prefix,
           },
         });
-        callback!(null, setMetadataResponse);
       } catch (e) {
         callback!(e);
         return;
       }
+
+      callback!(null, setMetadataResponse);
     })();
   }
 
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -99,6 +99,15 @@ export interface LifecycleRule {
   storageClass?: string;
 }
 
+export interface EnableLoggingOptions {
+  bucket?: string | Bucket;
+  prefix: string;
+}
+export type EnableLoggingResponse = [];
+export interface EnableLoggingCallback {
+  (err: Error | null): void;
+}
+
 export interface GetFilesOptions {
   autoPaginate?: boolean;
   delimiter?: string;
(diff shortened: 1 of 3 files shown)


src/bucket.ts
```diff
@@ -1789,6 +1798,100 @@ class Bucket extends ServiceObject {
...
+   *
+   * bucket.enableLogging(config, function(err, apiResponse) {
+   *   if (!err) {
+   *     // Logging functionality enabled successfully.
+   *   }
+   * });
+   *
+   * //-
+   * // Optionally, provide a destination bucket.
+   * //-
```
callmehiphop: Probably want to use a new `@example` tag with a `<caption>`

src/bucket.ts
```diff
@@ -1789,6 +1798,100 @@ class Bucket extends ServiceObject {
...
+   * const config = {
+   *   prefix: 'log',
+   *   bucket: 'destination-bucket'
+   * };
+   *
+   * bucket.enableLogging(config, function(err, apiResponse) {});
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
```
callmehiphop: Same as above!

src/bucket.ts
```diff
@@ -1789,6 +1798,100 @@ class Bucket extends ServiceObject {
...
+      typeof config === 'function' ||
+      typeof config.prefix === 'undefined'
+    ) {
+      throw new Error('A configuration object with a prefix is required.');
+    }
+
+    let logBucket = this.id;
+
+    if (typeof config.bucket !== 'undefined') {
+      if (util.isCustomType(config.bucket, 'Bucket')) {
```
callmehiphop: Totally just a side note, but we might consider turning this method into a generic to avoid having to casting things later.

src/bucket.ts
```diff
@@ -1789,6 +1798,100 @@ class Bucket extends ServiceObject {
...
+          entity: 'group-cloud-storage-analytics@google.com',
+          role: 'WRITER',
+        });
+        const [setMetadataResponse] = await this.setMetadata({
+          logging: {
+            logBucket,
+            logObjectPrefix: config.prefix,
+          },
+        });
+        callback!(null, setMetadataResponse);
```
callmehiphop: I don't think we should execute the callback here, only because it is wrapped inside of the `try` block and if a user error would occur, we would end up calling their callback twice

src/bucket.ts
```diff
@@ -1789,6 +1798,100 @@ class Bucket extends ServiceObject {
     );
   }
 
+  enableLogging(
+    config: EnableLoggingOptions
+  ): Promise<SetBucketMetadataResponse>;
+  /**
```
AVaksman: I believe a `callback` overload is needed here.

src/bucket.ts
stephenplusplus: I’m sure you’re right, although I’m not following the scenario. Could you elaborate how that would happen?

src/bucket.ts
```diff
@@ -1789,6 +1798,97 @@ class Bucket extends ServiceObject {
...
+   *
+   * @example <caption>Optionally, provide a destination bucket.</caption>
+   * const config = {
+   *   prefix: 'log',
+   *   bucket: 'destination-bucket'
+   * };
+   *
+   * bucket.enableLogging(config, function(err, apiResponse) {});
+   *
+   * @example <caption>If the callback is omitted, we'll return a Promise.
```
stephenplusplus: @callmehiphop Ran into the end of the line. How's this looking?

src/bucket.ts
stephenplusplus: Sorry, would you mind showing me what I should do? 😊 Currently I have it written for when just required configuration is passed in and the user is in promise mode, and then when the configuration is passed in as well as the optional callback.

AVaksman: I believe the implementation version with optional `callback` (below) is not counted as an overload by TS. 
Then for TS users if callback is passed, the compiler will complain for `Expected 1 argument, but got 2`
image

adding this overload will make the compiler happy
```ts
enableLogging(
    config: EnableLoggingOptions,
    callback: SetBucketMetadataCallback
  ): void;
```

stephenplusplus: I haven't been able to reproduce that, but I'm sure it's something I'm doing wrong. I added it and pushed!

src/bucket.ts
callmehiphop: I actually think you have to run over the end in this case. Pretty sure our docs aren't smart enough to know that the closing tag is part of the caption still.

src/bucket.ts
AVaksman: You should be able to reproduce it by trying a `callback` mode system test.
```ts
    it('should enable logging on current bucket by default callback', done => {
      bucket.enableLogging({prefix: PREFIX}, (err, metadata) => {
        assert.ifError(err);
        assert.deepStrictEqual(metadata.logging, {
          logBucket: bucket.id,
          logObjectPrefix: PREFIX,
        });
        done();
      });
    });
```

src/bucket.ts
callmehiphop: Sure

```js
bucket.enableLogging({prefix: 'hithere'}, err => {
  if (err.code === 404) {
    console.log('oh no!');
  }
});
```

Using that example, if we pretend the user forgets to check that `err` actually exists when it doesn't, it might result in a `TypeError` being thrown. That would move us to the `catch` section where we would execute the callback again with that very same `TypeError`.

stephenplusplus: Gotcha, thanks!

src/bucket.ts
```diff
@@ -1789,6 +1794,104 @@ class Bucket extends ServiceObject {
...
+            logBucket,
+            logObjectPrefix: config.prefix,
+          },
+        });
+      } catch (e) {
+        callback!(e);
+        return;
+      }
+
+      callback!(null, setMetadataResponse);
```
stephenplusplus: @callmehiphop 🤷‍♂?

callmehiphop: LGTM! It is exactly what I would have done.

src/bucket.ts
```diff
@@ -1789,6 +1794,105 @@ class Bucket extends ServiceObject {
...
+        logBucket = config.bucket;
+      }
+    }
+
+    (async () => {
+      let setMetadataResponse;
+
+      try {
+        const [policy] = await this.iam.getPolicy();
+        policy.bindings.push({
```
stephenplusplus: @frankyn is this the correct configuration?

frankyn: +1, it's the role with the least number of permissions following the least privileges model.

src/bucket.ts
```diff
@@ -1789,6 +1794,105 @@ class Bucket extends ServiceObject {
...
+          members: ['cloud-storage-analytics@google.com'],
```
frankyn: Ah, I missed the system test failure, `cloud-storage-analytics@google.com` is a group.

You should do:
```
members: ['group:cloud-storage-analytics@google.com']
``

File: src/bucket.ts [hunk 10]
```diff
@@ -1789,6 +1794,105 @@ class Bucket extends ServiceObject {
...
+          role: 'roles/storage.objectAdmin',
```
frankyn: Use `roles/storage.objectCreator` instead. GCS shouldn't need to overwrite/delete data.

AVaksman:
Should we worry about making codecov happy?
adding a case scenario to disregard `config.bucket` value if it is neither instance of `Bucket` nor `string`?
Or error out in that case?

stephenplusplus:
@AVaksman that's a good point. I don't think we should error, but I just made the check less verbose to just take whatever they give us. PTAL

AVaksman:
LGTM

chrishiestand:
Thanks everyone!

PR #878, opened 2019-10-07
refactor: clarify the error message for file#move

AVaksman:

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -3052,6 +3052,7 @@ class File extends ServiceObject<File> {
           if (err) {
             err.message = 'file#delete failed with an error - ' + err.message;
             callback!(err, destinationFile, apiResponse);
+            return;
           }
           callback!(null, destinationFile, copyApiResponse);
         });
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -3047,10 +3049,14 @@ class File extends ServiceObject<File> {
         this.bucket.name !== destinationFile!.bucket.name
       ) {
         this.delete(options, (err, apiResponse) => {
-          callback!(err, destinationFile, apiResponse);
+          if (err) {
+            err.message = 'file#delete failed with an error - ' + err.message;
+            callback!(err, destinationFile, apiResponse);
```
stephenplusplus: This needs a return after. Alternatively, just modify the err.message property and continue to pass “err” into the callback on the line below. If it’s null, it will still be null on that line and nothing would have changed.

test/file.ts
```diff
@@ -3526,6 +3548,7 @@ describe('File', () => {
         };
         file.move('new-filename', (err: Error) => {
           assert.strictEqual(err, error);
+          assert.ok(err.message.includes('delete'));
```
stephenplusplus: Could you elaborate this check to include the whole prefix string and original error message?

AVaksman: Thanks, done here and by fail on copy above

src/file.ts
AVaksman: Thanks, added a `return` to be consistent with the same condition by `copy` above.

PR #887, opened 2019-10-17
build: packageName should be set for release configuration

bcoe:
we were not updating `samples/package.json`, because we didn't know the correct `packageName`.

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,1 +1,2 @@
 releaseType: node
+packageName: '@google-cloud/storage'


.github/release-please.yml
```diff
@@ -1 +1,2 @@
 releaseType: node
+packageName: '@google-cloud/storage'
```
JustinBeckwith: Grump.  Will we need to specify this anywhere?  Any reason we can't just read this from `package.json`?  All of the duplicate config makes me nervous.

bcoe: I thought about reading it from `package.json`, but it would mean that `release-please` needs to read in two config files, which is silly, and not every language has a `package.json`.

I think this is reasonable because it's a config setting specifically for `release-please`, and this is a config file for release-please.

bcoe:
landed this landed this setting still

PR #892, opened 2019-10-21
docs: explain how to use delimiter and prefix

### LINKED ISSUE #845: Storage bucket.getFiles() does not return prefixes
#### Environment details

  - OS: Linux 4a68f4538d68 4.14.131-linuxkit #1 SMP Fri Jul 19 12:31:17 UTC 2019 x86_64 Linux
  - Node.js version: v12.10.0
  - npm version: 6.11.3
  - `@google-cloud/storage` version: 3.2.1

#### Steps to reproduce

  1. call bucket.getFiles({prefix: 'some', delimiter: '/'})
  2. no way of getting the prefixes from the [response](https://cloud.google.com/storage/docs/json_api/v1/objects/list)

stephenplusplus:
Fixes #845

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -2092,20 +2092,13 @@ class Bucket extends ServiceObject {
    * }, callback);
    *
    * //-
-   * // With `autoPaginate: false`, it's possible to iterate over files which
-   * // incorporate a common structure using a delimiter.
-   * //
-   * // Consider the following remote objects:
-   * //
-   * //   1. "a"
-   * //   2. "a/b/c/d"
-   * //   3. "b/d/e"
-   * //
-   * // Using a delimiter of `/` will return a single file, "a".
-   * //
-   * // Check the `apiResponse` argument for the `prefixes` property to see what
-   * // "sub-directories" were found.
+   * // If the callback is omitted, we'll return a Promise.
    * //-
+   * bucket.getFiles().then(function(data) {
+   *   const files = data[0];
+   * });
+   *
+   * @example <caption><h6>Simulating a File System</h6><p>With `autoPaginate: false`, it's possible to iterate over files which incorporate a common structure using a delimiter.</p><p>Consider the following remote objects:</p><ol><li>"a"</li><li>"a/b/c/d"</li><li>"b/d/e"</li></ol><p>Using a delimiter of `/` will return a single file, "a".</p><p>`apiResponse.prefixes` will return the "sub-directories" that were found:</p><ol><li>"a/"</li><li>"b/"</li></ol></caption>
    * bucket.getFiles({
    *   autoPaginate: false,
    *   delimiter: '/'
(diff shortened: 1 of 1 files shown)


src/bucket.ts
```diff
@@ -548,6 +548,28 @@ class Bucket extends ServiceObject {
...
+   *     bucket.getFiles(nextQuery, callback);
+   *   } else {
+   *     // prefixes = The finished array of prefixes.
+   *   }
+   * }
+   *
+   * bucket.getFiles({
+   *   autoPaginate: false,
+   *   delimiter: '/'
+   * }, callback);
```
frankyn: Could we incorporate an example as well?
```
### Example
For example, I have a bucket with the following object names:
Remember each of the following names are separate names with the delimiter in the name.
Note: pre-chunking on the part of the name you might want to search on. 
 
- a
- a/b/c/d
- b/d/e
 
#### First Request
I call list_blobs with `delimiter="/" and prefix=None`, and GCS will respond with the following list of blobs:
 
- a
 
In `blobs.prefixes`, you'll now find additional information on available prefixes "or" subdirectories exist at this level:
 
- a/
- b/
 
#### Second Request
 
Now, I have a list of prefixes that I can continue opening similar to a filesystem. 
Next, I'll go into `prefix="a/" and delimiter="/"`.
List of Blobs:
- Empty (there are no objects in this directory)
 
List of Prefixes (sub-directories):
- a/b/
 
At this point I can continue making requests to simulate a file system using `prefix="a/b/" and delimiter="/"`. 
```

stephenplusplus: Sorry, forgot to tag you here. On it!

stephenplusplus: @frankyn I added this in this commit: https://github.com/googleapis/nodejs-storage/pull/892/commits/f415824d73509f9da17f179711d2a3e8b4da4dd9 -- PTAL!

Just a note about why there are two documentation areas. The documentation for the streaming method, `getFilesStream()`, makes a mention of how to use the non-streaming `getFiles()` with prefixes/delimiters. Then, `getFiles()` has the more elaborate documentation.

src/bucket.ts
```diff
@@ -548,6 +548,28 @@ class Bucket extends ServiceObject {
    *   .on('data', function(file) {
    *     this.end();
    *   });
+   *
+   * //-
```
callmehiphop: Should use a new `@example` tag with a `<caption>`

stephenplusplus:
@callmehiphop I'm not sure if we should do the caption thing:

- Everywhere else, even within the same method's docs, is the old fashioned way -- will they merge together on the docs site correctly? Should I update more than just my new ones?

- These new captions are more lengthy, where I use a `<ol>` style numbered list. For maintainability, stuffing that all on one line within the <caption> would be bad, but how would I make sure it ends up printed as a numbered list when rendered on the docs site?

callmehiphop:
@stephenplusplus I don't see why they wouldn't play well together. JSDoc doesn't recognize that token as anything other than a JS comment (as shown in [`getFiles`](https://googleapis.dev/nodejs/storage/latest/Bucket.html#getFiles)). 

I'm not really sure if you can nest html inside of a caption tag, but you should be able to test your changes locally to see what happens. Otherwise for lengthy captions, I just resort to adding JS comments in the code sample.

stephenplusplus:
@callmehiphop PTAL!

A preview:
image

PR #893, opened 2019-10-21
Revert "fix: use storage.googleapis.com for api endpoint"

frankyn:
Reverts googleapis/nodejs-storage#854. @broady and @JustinBeckwith, I'm reverting this change to align with the new rollout plan to reduce impact.

PR #899, opened 2019-10-24
fix: use storage.googleapis.com for api endpoint

frankyn:
Reverts googleapis/nodejs-storage#893

Turns out we want to no fix this and I'm reverting my revert based on discussion with @broady and @crwilcox.


broady:
LGTM

frankyn:
I'm pending Github approval to merge otherwise I can't.

PR #905, opened 2019-10-28
fix(signed-url): replace encodeURIComponent with custom encoding function

jkwlui:
Fixes #859  🦕

Made changes to how signed URLs (v2 and v4) are url-encoded:
1. `encodeURIComponent` does not encode the following characters: `! * ' ( )`, which should have been encoded.
2. Slashes `/` in the object's (file) name should not have been encoded. (caused issue #857)

PR #906, opened 2019-10-29
fix(tests): update conformance tests for signed urls

jkwlui:
Pulls the latest test data https://github.com/googleapis/conformance-tests.

There are 3 failing tests, each demonstrating a different bug (in Node):
1) POST for resumable uploads
 - will be fixed in #907
 - this is failing because it represents users passing a `X-goog-resumable: start` extension header in the request.
 - The nodejs-storage library applies this header for the user if the user specified `action: resumable` in `SignedUrlConfig`.
 - However, the library applies `x-goog-resumable: start` *lower-cased*, and when merged with the user provided `extensionHeaders`, both headers exist and result in duplicate signed header entries in the resulting URL.

2) Slashes in object name should not be URL encoded
 - will be fixed in #905  

3) Headers should be trimmed
 - We assumed sequential whitespaces should be replaced by a single space.
 - The correct behaviour here is to replace them with the last whitespace character:
 - `123\t\t\tabc => 123\tabc`

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -43,7 +43,7 @@ export function objectEntries<T>(obj: {[key: string]: T}): Array<[string, T]> {
  * Encode every byte except `A-Z a-Z 0-9 ~ - . _`.
  *
  * encodeURI patches encodeURIComponent() by:
- *  - additionally encoding `! * ' ( )` characters;
+ *  - additionally encoding `! * ' ( )` characters (@see [MDN: fixedEncodeURIComponent]{@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent}
  *  - conditionally encoding `/` if encodeSlash is `true`.
  * @param {string} uri The URI to encode.
  * @param [boolean=false] encodeSlash If `true`, the "/" character is not encoded.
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -35,3 +37,41 @@ export function normalize<T = {}, U = Function>(
...
+ */
+export function encodeURI(uri: string, encodeSlash: boolean): string {
+  // Encode using JavaScript's encodeURIComponent, excluding "/" if encodeSlash is
+  // `true`.
+  const encoded = uri
+    .split('/')
+    .map(encodeURIComponent)
+    .join(encodeSlash ? '%2F' : '/');
+
+  // Encode additional characters not encoded by encodeURIComponent.
```
alexander-fenster: Maybe just follow this [example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) from MDN:

```js
function fixedEncodeURIComponent(str) {
  return encodeURIComponent(str).replace(/[!'()*]/g, function(c) {
    return '%' + c.charCodeAt(0).toString(16);
  });
}
```
same but more readable IMHO.

jkwlui: fixed in #905

jkwlui:
Closing in favour of #912

PR #908, opened 2019-10-29
feat: add support for Archive storage class

### LINKED ISSUE #902: Storage: Add support for archive storage class
GCS is adding a storage class called "archive" to the existing options for object storage (coldline, standard, etc). Support for this should be added to this client library. This has already been implemented in the C++ client library and there is a PR in python; reference PR [here](https://github.com/googleapis/google-cloud-python/issues/9532)

This should be straightforward as ARCHIVE should just mirror existing valid storage classes. I see the following places where changes are required in order to accomplish this:

* https://github.com/googleapis/nodejs-storage/blob/master/src/storage.ts#L589 (add new constant; also update CreateBucketRequest in this file as well as documentation).
*https://github.com/googleapis/nodejs-storage/blob/master/src/bucket.ts (update all documentation).
* https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts (update documentation).
* Consider adding or modifying at least 1 test to exercise the archive storage class.

stephenplusplus:
Fixes #902


src/bucket.ts
```diff
@@ -2851,9 +2851,9 @@ class Bucket extends ServiceObject {
    * @see [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
    *
    * @param {string} storageClass The new storage class. (`standard`,
-   *     `nearline`, `coldline`, or `durable_reduced_availability`).
-   *     **Note:** The storage classes `multi_regional` and `regional`
-   *     are now legacy and will be deprecated in the future.
+   *     `nearline`, `coldline`, `archive`, or `durable_reduced_availability`).
```
tritone: durable_reduced_availability should be moved to legacy as well.

src/file.ts
```diff
@@ -3246,7 +3246,7 @@ class File extends ServiceObject<File> {
    * @see [Storage Classes]{@link https://cloud.google.com/storage/docs/storage-classes}
    *
    * @param {string} storageClass The new storage class. (`standard`,
-   *     `nearline`, or `coldline`)
+   *     `nearline`, `archive`, or `coldline`)
```
tritone: I would order these as standard, nearline, coldline, archive (archive is the coldest).

test/index.ts
```diff
@@ -497,6 +497,15 @@ describe('Storage', () => {
     });
 
     describe('storage classes', () => {
+      it('should expand metadata.archive', done => {
```
tritone: Is there a way to de-dupe the logic that's repeated for every storage class here? I'm far from a Node expert so feel free to tell me that this is impossible or a bad idea! :)

stephenplusplus: Not at all! I think that's a pretty good catch, however because this is our test suite, I usually err on the side of over-emphasis. Having each case handled in its own test makes it stand out as more deliberate than going through a loop or trying to group them all in a single test. Overall, optimization isn't a thought I hold too close to tests-- I think being more verbose and readable is a good thing, and hopefully holds future test writers to the same standard... at least, that's the thought anyway! :)

tritone:
Thanks for approving @JustinBeckwith ! We should be able to go ahead with merging on/after Jan 2, apologies for the delay here.

tritone:
Just an update; we're clear to merge this now and release the change after the release freeze has ended (between 1/2-1/8). Let me know if there are any issues with this timeline.

Issue #55, opened 2017-10-10
Smart Start Resumable Uploads


stephenplusplus:
Some environments don't have writable $HOME directories, which will return an error from gcs-resumable-upload. We should confirm we can write to the directory before attempting a resumable upload.

Here are the ideal scenarios:

#### file.createWriteStream()
No `options.resumable` argument supplied.
  - Can write to home directory? Start resumable upload.
  - Cannot write to home directory? Fall back to simple upload.

We can explain in the docs we will attempt a resumable upload, but won't guarantee it. In either case, the user does not know which was used.

#### file.createWriteStream({ resumable: true })
Resumable upload was explicitly requested.
  - Can write to home directory? Start resumable upload.
  - Cannot write to home directory? Emit an error, close the stream.

#### file.createWriteStream({ resumable: false })  
Resumable upload was explicitly denied.
  - Do not attempt to write to home directory.
  - Start simple upload.

PR #915, opened 2019-11-04
fix: do not check access of configPath

### LINKED ISSUE #909: Improper check for config directory when performing resumable upload
Thank you for all your work on this package! The handling of the config directory when performing resumable uploads has a few issues.

First, [the code does not properly consider the `configPath` option when setting the `configDir` variable](https://github.com/googleapis/nodejs-storage/blob/862fb16db2990c958c0097252a44c775431a7b3f/src/file.ts#L1768).

```typescript
const configDir = xdgBasedir.config || os.tmpdir();
```

The code should probably be:

```typescript
import * as path from "path";
const configDir = (options.configPath ? path.dirname(options.configPath) : undefined) || xdgBasedir.config || os.tmpdir();
```

Second, the [check for write access to the `configDir`](https://github.com/googleapis/nodejs-storage/blob/862fb16db2990c958c0097252a44c775431a7b3f/src/file.ts#L1770) fails if the directory does not exist, leading to a ResumableUploadError.

```typescript
      fs.access(configDir, fs.constants.W_OK, err => {
        if (err) {
          if (options.resumable) {
            const error = new ResumableUploadError(
              [
                'A resumable upload could not be performed. The directory,',
                `${configDir}, is not writable. You may try another upload,`,
                'this time setting `options.resumable` to `false`.',
              ].join(' ')
            );
            stream.destroy(error);
            return;
          }
        …
        }
      }
```

It is not necessarily a problem if the config directory does not exist because the underlying implementation will create it if is does not exist, see [gcs-resumable-upload](https://github.com/googleapis/gcs-resumable-upload/blob/63199814aab127a5c3e5c1a22ac905d5b70b2209/src/index.ts#L268-L271), [configstore](https://github.com/yeoman/configstore/blob/764ac63d17de9e3d322c1c42180a8d5bec85391a/index.js#L59), and [make-dir](https://github.com/sindresorhus/make-dir#readme). It is only a problem if you do not have write access to the deepest existing parent of `configDir`.

#### Environment details

  - OS: all
  - Node.js version: all
  - npm version: all
  - `@google-cloud/storage` version: <=4.0.0

#### Steps to reproduce

Perform an upload with resumable set to true after removing/hiding ~/.config and you get a ResumableUploadError. Here is some example code you can use to demonstrate the issue after replacing `YOUR_BUCKET` as appropriate.

```typescript
import { Storage } from "@google-cloud/storage";
async function main(): Promise<void> {
    await storage.bucket("YOUR_BUCKET")
        .upload(__filename, { resumable: true });
}
main().catch(e => {
    process.stderr.write(`Error: ${e.message}\n`);
    process.exit(1);
});
```

On a POSIX-like OS, e.g., GNU/Linux or macOS, you save the above script as "test.ts", compile it to JavaScript, and run it while hiding your `$HOME/.config` with the following command:

```console
$ HOME=/tmp GOOGLE_APPLICATION_CREDENTIALS=path/to/creds node test.js
Error: A resumable upload could not be performed. The directory, /tmp/.config, is not writable. You may try another upload, this time setting `options.resumable` to `false`.
```

This fails even though you would have permissions to create `/tmp/.config`. In other words, this succeeds:

```console
$ mkdir /tmp/.config
$ HOME=/tmp GOOGLE_APPLICATION_CREDENTIALS=path/to/creds node test.js
```

You can verify that `configPath` is _not_ checked and is created if it does not exist by changing the above example to explicitly set the `configPath` option to a non-existing directory and verifying the directory exists after running the script.

```typescript
import { Storage } from "@google-cloud/storage";
async function main(): Promise<void> {
    await storage.bucket("YOUR_BUCKET")
        .upload(__filename, { configPath: "/tmp/does/not/exist", resumable: true });
}
main().catch(e => {
    process.stderr.write(`Error: ${e.message}\n`);
    process.exit(1);
});
```

```console
$ GOOGLE_APPLICATION_CREDENTIALS=path/to/creds node test.js
$ ls -al /tmp/does/not/exist
-rw------- 1 XX YY 67 Oct 30 08:17 /tmp/does/not/exist
```

Fixing the writable check will be challenging since you would have to parse the path and check `fs.constants.W_OK` access on the `configDir` and each of its parents in turn, discarding results for paths that do not exist, and stopping once you found one that does exist and only consider that result in the test. It may just be better to let the downstream packages handle the work and properly respond to errors from them.

stephenplusplus:
Fixes #909

A couple things happened along this journey to fix #909.

I learned that the [configstore](http://gitnpm.com/configstore) `configPath` option isn't for a directory, it's actually the file name. We use that through gcs-resumable-upload to persist metadata about the progress of a resumable upload.

Even though that understanding was wrong, I didn't make any actual code changes in this PR. All I did was rephrase the documentation. Interestingly, the documentation was half-correct when it said "This maps to the configstore option by the same name." So... breaking change? I would not say it is, since code-wise, things behave the same.

Oh, what was #909 about? It's amazingly described, so you could read it just for fun, but to sum it up: when the user gave us a `configPath`, we were trying to make sure that the directory is writable. We don't need to worry about that, gcs-resumable-upload/configstore will sort all of that out, as that's more in their domain. The reason we have the writable access check is strictly for environments (GAE (see #55)) where the user doesn't even think about simple or resumable. They're just uploading a file. Sometimes, we would try to use a resumable upload that defaults to their home directory, but there wouldn't be one, or it was not writable. So, the check is strictly to make sure the "auto-detected $HOME-like environment" is writable. Otherwise, we fall back to a simple upload.

But, if a user provides `configPath`, they obviously know what they're doing, and they want a resumable upload. This PR codifies that assumption.

So, here was my fix. @ddgenome -- let me know what you think!

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1447,9 +1447,9 @@ class File extends ServiceObject<File> {
    */
   /**
    * @typedef {object} CreateResumableUploadOptions
-   * @property {string} [configPath] Where the `gcs-resumable-upload`
-   *     configuration file should be stored on your system. This maps to the
-   *     [configstore option by the same
+   * @property {string} [configPath] A full JSON file path to use with
+   *     `gcs-resumable-upload`. This maps to the [configstore option by the
+   *     same
    * name](https://github.com/yeoman/configstore/tree/0df1ec950d952b1f0dfb39ce22af8e505dffc71a#configpath).
    * @property {object} [metadata] Metadata to set on the file.
    * @property {string} [origin] Origin header to set for the upload.
@@ -1554,8 +1554,8 @@ class File extends ServiceObject<File> {
   /**
    * @typedef {object} CreateWriteStreamOptions Configuration options for File#createWriteStream().
    * @property {string} [configPath] **This only applies to resumable
-   *     uploads.** Where the `gcs-resumable-upload` configuration file should
-   * be stored on your system. This maps to the [configstore option by the same
+   *     uploads.** A full JSON file path to use with `gcs-resumable-upload`.
+   *     This maps to the [configstore option by the same
    * name](https://github.com/yeoman/configstore/tree/0df1ec950d952b1f0dfb39ce22af8e505dffc71a#configpath).
    * @property {string} [contentType] Alias for
    *     `options.metadata.contentType`. If set to `auto`, the file name is used
@@ -1759,6 +1759,11 @@ class File extends ServiceObject<File> {
     stream.on('writing', () => {
       if (options.resumable === false) {
         this.startSimpleUpload_(fileWriteStream, options);
+        return;
+      }
+
+      if (options.configPath) {
+        this.startResumableUpload_(fileWriteStream, options);
         return;
       }
 
--- a/test/file.ts
+++ b/test/file.ts
@@ -1709,6 +1709,21 @@ describe('File', () => {
       writable.write('data');
     });
 
+    it('should start a resumable upload if configPath is provided', done => {
+      const options = {
+        metadata: METADATA,
+        configPath: '/config/path.json',
+      };
+      const writable = file.createWriteStream(options);
+
+      file.startResumableUpload_ = (stream: {}, options_: {}) => {
+        assert.deepStrictEqual(options_, options);
+        done();
+      };
+
+      writable.write('data');
+    });
+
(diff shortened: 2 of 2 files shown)


callmehiphop:
LGTM but I'll give this a little more time for @ddgenome to chime in

PR #922, opened 2019-11-07
fix(docs): rendering for createReadStream method in File class

arku:
### Description

One of the comments in the documentation for `createReadStream` has the following content.

> To limit the downloaded data to only a byte range, pass an options object.

And a part of it ('object') is not rendered as a comment in https://googleapis.dev/nodejs/storage/latest/File.html#createReadStream as it is missing a `//`.

### Checklist

#### Before

image

#### After

image


stephenplusplus:
Thank you very much!

arku:
@jkwlui Thanks!

PR #940, opened 2019-11-20
fix: add createBucket file and updated relevant test

sofisl:
Fix: internal reference b/144088253

### DIFF
--- a/samples/createNewBucket.js
+++ b/samples/createNewBucket.js
@@ -15,36 +15,35 @@
  */
 
 function main(bucketName = 'my-bucket') {
-    // [START storage_create_bucket]
-    /**
-     * TODO(developer): Uncomment the following line before running the sample.
-     */
-    // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // [START storage_create_bucket]
+  /**
+   * TODO(developer): Uncomment the following line before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
 
-    // Imports the Google Cloud client library
-    const {Storage} = require('@google-cloud/storage');
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
 
-    // Creates a client
-    const storage = new Storage();
+  // Creates a client
+  const storage = new Storage();
 
-    async function createBucket() {
-        // Creates a new bucket in the Asia region with the coldline default storage
-        // class. Leave the second argument blank for default settings.
-        //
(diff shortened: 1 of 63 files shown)


sofisl:
@fhinkel - I'm not sure why the build test isn't passing. It seems to be an issue with the storage.ts test under system-test, but I didn't touch that test, and confirmed that it's up-to-date with the master branch. Not sure if I'm missing something or if it's a bug...

fhinkel:
@sofisl I don't see right away why the system test is failing. But there are hundreds of tests, I wonder if the new test is somehow messing with the system tests, because it is creating a bucket and not deleting it. 

Can you delete the existing `createBucket` code and test in the sample. And we should change the new test to delete the bucket at the end of the test run. Otherwise we're left with a bucket after running the test once.

bcoe:
@sofisl @fhinkel I have a hunch that this is a flaky test, just kicked off the build to hopefully confirm.

fhinkel:
Don't we need to delete the existing `createBucket` sample? I only see code additions.

samples/createNewBucket.js
```diff
@@ -0,0 +1,49 @@
...
+  const storage = new Storage();
+
+  async function createBucket() {
+    // Creates a new bucket in the Asia region with the coldline default storage
+    // class. Leave the second argument blank for default settings.
+    //
+    // For default values see: https://cloud.google.com/storage/docs/locations and
+    // https://cloud.google.com/storage/docs/storage-classes
+
+    await storage.createBucket(bucketName, {
```
stephenplusplus: Should we show that this returns a Bucket object?

```js
const [bucket] = await storage.createBucket(bucketName, {
  location: 'ASIA',
  storageClass: 'COLDLINE',
});
```

sofisl: Great catch! Fixed both issues. Thanks for the review!

sofisl:
@fhinkel - Lint is failing because I don't reassign the variable bucket, but that's the way the docs and @stephenplusplus recommend doing it.

Error:

Screen Shot 2019-11-25 at 12 50 06 PM

Also, I think the kokoro system test is flaky, according to @bcoe.

sofisl:
@fhinkel - I think the system test is still flaky, this is the error - https://screenshot.googleplex.com/MZ27aYXMaeD

Otherwise all the tests have passed

sofisl:
@fhinkel - waiting on CL to be approved to merge cl 284587412

fhinkel:
@sofisl Sorry I didn't see this. Always better to ping me on chat rather than relying on GitHub notifications. LGTM! Thanks

PR #941, opened 2019-11-20
test: address test flake in system-tests

bcoe:
we've been seeing a frequent test flake recently, related to credentials being unset, let's make sure after each test we set credential environment variable back.

see: #940


bcoe:
I was hoping I'd figured out what was flaking, but have not yet.

PR #944, opened 2019-11-22
fix: update PolicyDocument types

### LINKED ISSUE #943: Incorrect types on PolicyDocument interface
The types for `PolicyDocument` are incorrect with what is returned when calling `getSignedPolicy`.

The response is:

```
{
  string: '{"expiration":"2099-06-16T11:11:11.000Z","conditions":[["eq","$key","test.png"],{"bucket":"test"},["starts-with","content-type","image"],["content-length-range",0,1000000]]}',
  base64: 'asdasd=',
  signature: 'qasdasd'
}
```

But the `PolicyDocument` interface is:

```
export interface PolicyDocument {
    expiration: string;
    conditions: Array<Array<string | number>>;
    string: string;
}
```

reichjustin:
- [ X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X ] Ensure the tests and linter pass
- [X ] Code coverage does not decrease (if any source code was changed)
- [X ] Appropriate docs were updated (if necessary)

Fixes #943

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -108,9 +108,9 @@ export interface GetSignedUrlCallback {
 }
 
 export interface PolicyDocument {
-  expiration: string;
-  conditions: Array<Array<string | number>>;
   string: string;
+  base64: string;
+  signature: string;
 }
 
 export type GetSignedPolicyResponse = [PolicyDocument];


stephenplusplus:
Good catch, thank you for the fix!

PR #960, opened 2019-12-21
feat(iam): getIamPolicy to support requestedPolicyVersion

jkwlui:
Minimal code changes is involved in this PR.
We map `optionsRequestedPolicyVersion` to `requestedPolicyVersion` so the surface is uniform across products.

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -23,20 +23,11 @@ import arrify = require('arrify');
 import {Bucket} from './bucket';
 import {normalize} from './util';
 
-/**
- * @typedef {object} GetPolicyOptions
- * @property {string} [userProject] The ID of the project which will be billed for
- *     the request.
- */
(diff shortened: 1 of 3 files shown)


src/iam.ts
```diff
@@ -75,19 +66,22 @@ export interface SetPolicyCallback {
   (err?: Error | null, acl?: Policy, apiResponse?: object): void;
 }
 
-/**
- * @typedef {object} Policy
```
bcoe: curious why we're dropping these jsdoc strings, are they not used anywhere I'm guessing?

jkwlui: Not dropped, just moved downwards :) `tsc` rips the jsdoc comments above interfaces so it doesn't go into the compiled docs.

bcoe: 👍 awesome.

PR #973, opened 2020-01-06
docs(samples): storage IAM condition

jkwlui:


samples/iam.js
```diff
@@ -111,7 +187,7 @@ async function removeBucketIamMember(bucketName, roleName, members) {
   const bucket = storage.bucket(bucketName);
 
   // Gets and updates the bucket's IAM policy
-  const [policy] = await bucket.iam.getPolicy();
+  const [policy] = await bucket.iam.getPolicy({requestedPolicyVersion: 3});
```
frankyn: Add conditional policy check to make sure a condition isn't set when removing a member from a role. It explicitly tells a reader that the code ignores bindings with conditions.

jkwlui: 👍 fixed!

stephenplusplus:
System tests currently failing with:

```
  1) storage
       without authentication
         public data
           should list and download a file:
     Error: Request failed with status code 503
      at Gaxios._request (node_modules/gaxios/src/gaxios.ts:109:15)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

PR #976, opened 2020-01-07
fix: refactoring listBuckets function into its own file

sofisl:


samples/listBuckets.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright 2019 Google LLC
```
fhinkel: 2020

PR #977, opened 2020-01-07
chore: refactoring deleteBucket sample into a new file

sofisl:


laljikanjareeya:
@fhinkel  and @sofisl it would be great if we change the documentation as well while refactoring the samples.

please refer below doc sample link for delete bucket:

https://github.com/googleapis/nodejs-storage/blob/0ba69f1b3d6093701dac927fa4543d2d911ce8b0/src/bucket.ts#L656

PR #990, opened 2020-01-08
fix: refactor printFileAclForUser sample into its own file

sofisl:

### DIFF
--- a/samples/printFileAclForUser.js
+++ b/samples/printFileAclForUser.js
@@ -1,0 +1,60 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


samples/printFileAclForUser.js
```diff
@@ -0,0 +1,60 @@
...
+ * file Access Control Lists with the Google Cloud Storage API.
+ *
+ * For more information, see the README.md under /storage and the documentation
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main(
+  bucketName = 'my-bucket',
+  filename = 'test.txt',
+  userEmail = 'jdobry@google.com'
```
fhinkel: Can we use a dummy email address here? Thanks

fhinkel: Never mind, this is outside the tag :)

PR #1017, opened 2020-01-08
fix: refactor removeBucketIamMember.js

sofisl:

### DIFF
--- a/samples/removeBucketIamMember.js
+++ b/samples/removeBucketIamMember.js
@@ -1,0 +1,77 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


samples/removeBucketIamMember.js
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function main(
+  bucketName = 'my-bucket',
+  roleName = 'roles/storage.objectViewer',
+  members = 'user:test@example.com'
+) {
+  //including this logic to remove yargs functionality
```
fhinkel: This comment is confusing :) I think we can do without.

PR #1030, opened 2020-01-09
fix: refactor rotateEncryptionKey sample into its own file

sofisl:


sofisl:
Thanks for the review, @bcoe! @fhinkel showed me how to do it today, so that'll be the next step once the PR is merged :)

PR #1063, opened 2020-02-06
fix: skip validation for server decompressed objects

### LINKED ISSUE #709: CONTENT_DOWNLOAD_MISMATCH with successful file download

jkwlui:
Fixes #709

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1212,6 +1212,7 @@ class File extends ServiceObject<File> {
     // once Node 8 support is discontinued
     const throughStream = streamEvents(through()) as Duplex;
 
+    let isServedCompressed = true;
     let crc32c = true;
     let md5 = false;
 
@@ -1294,7 +1295,7 @@ class File extends ServiceObject<File> {
       //      applicable, to the user.
       const onResponse = (
         err: Error | null,
-        body: ResponseBody,
+        _body: ResponseBody,
         rawResponseStream: Metadata
       ) => {
         if (err) {
@@ -1312,7 +1313,7 @@ class File extends ServiceObject<File> {
         rawResponseStream.on('error', onComplete);
 
         const headers = rawResponseStream.toJSON().headers;
-        const isCompressed = headers['content-encoding'] === 'gzip';
+        isServedCompressed = headers['content-encoding'] === 'gzip';
         const shouldRunValidation = !rangeRequest && (crc32c || md5);
         const throughStreams: Writable[] = [];
 
@@ -1321,7 +1322,7 @@ class File extends ServiceObject<File> {
           throughStreams.push(validateStream);
         }
 
-        if (isCompressed && options.decompress) {
+        if (isServedCompressed && options.decompress) {
           throughStreams.push(zlib.createGunzip());
         }
 
@@ -1368,6 +1369,17 @@ class File extends ServiceObject<File> {
         }
 
         onCompleteCalled = true;
+
+        // TODO: Remove once the backend issue is fixed.
+        // If object is stored compressed (having metadata.contentEncoding === 'gzip')
+        // and was served decompressed, then skip checksum validation because the
+        // remote checksum is computed against the compressed version of the object.
+        // See https://github.com/googleapis/nodejs-storage/issues/709.
+        if (this.metadata.contentEncoding === 'gzip' && !isServedCompressed) {
+          onCompleteCalled = true;
+          throughStream.end();
+          return;
+        }
 
         const hashes = {
           crc32c: this.metadata.crc32c,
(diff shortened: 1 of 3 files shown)


crwilcox:
Thanks @jkwlui . I am going to make a small tweak to the TODO and get CI Green

src/file.ts
```diff
@@ -1369,6 +1370,17 @@ class File extends ServiceObject<File> {
 
         onCompleteCalled = true;
 
+        // TODO(https://github.com/googleapis/nodejs-storage/issues/709):
+        // Remove once the backend issue is fixed.
+        // If object is stored compressed (having metadata.contentEncoding === 'gzip')
+        // and was served decompressed, then skip checksum validation because the
+        // remote checksum is computed against the compressed version of the object.
+        if (this.metadata.contentEncoding === 'gzip' && !isServedCompressed) {
+          onCompleteCalled = true;
```
stephenplusplus: I think this is already set above.

system-test/storage.ts
```diff
@@ -2250,6 +2250,37 @@ describe('storage', () => {
...
+          gzip: true,
+        })
+        .then(() => {
+          tmp.setGracefulCleanup();
+          tmp.file((err, tmpFilePath) => {
+            assert.ifError(err);
+
+            const file = bucket.file(filename);
+            file
+              .createReadStream()
```
stephenplusplus: This needs an on-error handler.

system-test/storage.ts
```diff
@@ -2250,6 +2250,37 @@ describe('storage', () => {
       });
     });
 
+    it('should skip validation if file is served decompressed', done => {
+      const filename = 'logo-gzipped.png';
+      bucket
+        .upload(FILES.logo.path, {
```
feywind: Re: style comments - the mix of promises and callbacks makes me twitch a bit. Actually the preferred way now is to use async/await, so you could make the lambda passed to it() async and then await on stuff. I think there's a promisify function we have handy for tmp?

bcoe: for unit tests, I'd be tempted to just use `mktempSync`, I don't know that we need a module for this:

https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options

I'd also consider switching to to `async`/`await`, would probably unravel this a bunch.

src/file.ts
crwilcox: Good catch.

system-test/storage.ts
crwilcox: Different from the one on line 2269?

stephenplusplus: Yeah that one will catch errors from the write stream, but if the read stream has an error, it would go uncaught currently.

system-test/storage.ts
jkwlui: Ended up using `tmp.fileSync()` since it's already there :)
Probably gonna require more boilerplate code if we'd switch to using `fs.mkdTempSync`.

system-test/storage.ts
crwilcox: Looks like @jkwlui addressed this?

PR #1067, opened 2020-02-10
feat(signed-url): support virtual hosted style url

### LINKED ISSUE #1045: [Storage] V4 Signature Support for virtual hosted-style signed
Tracking issue for work defined in companion document.

[New Client Request - Support for virtual hosted-style signed URLs](https://docs.google.com/document/d/1TkoDiZjz6nINNVQrYbrfaODhmcMTR-crb94aR1cXZVA/edit#)

jkwlui:
Closes #1045


src/bucket.ts
```diff
@@ -244,6 +244,7 @@ export interface GetBucketSignedUrlConfig {
   action: 'list';
   version?: 'v2' | 'v4';
   cname?: string;
+  urlStyle?: 'virtual-host' | 'path';
```
frankyn: I'd like this to be a boolean rather than a string value if possible. 
```
virtualHostedStyle: true | false.
```

By default it's 'path' style and the documentation here can be simplified.

src/bucket.ts
```diff
@@ -2398,6 +2399,12 @@ class Bucket extends ServiceObject {
    *     Note: 'v4' supports maximum duration of 7 days (604800 seconds) from now.
    * @property {string} [version='v2'] The signing version to use, either
    *     'v2' or 'v4'.
+   * @param {boolean} [virtualHostedStyle=false] Use virtual hosted-style
+   *     URLs ('https://mybucket.storage.googleapis.com/...') instead of path-style
+   *     ('https://mybucket.storage.googleapis.com/...'). Virtual hosted-style URLs
```
frankyn: Path-style is storage.googleapis.com/mybucket.

src/bucket.ts
```diff
@@ -2398,6 +2399,12 @@ class Bucket extends ServiceObject {
...
+   *     should generally be preferred instaed of path-style URL.
+   *     Currently defaults to 'path', although this may change in a future
```
frankyn: Currently defaults to `false` for Path Style URLs.

PR #1070, opened 2020-02-12
feat(signed-url): support SHA256 for signed payloads

jkwlui:


src/file.ts
```diff
@@ -2323,6 +2328,10 @@ class File extends ServiceObject<File> {
    *     future major-version release.
    * @param {string} [config.cname] The cname for this bucket, i.e.,
    *     "https://cdn.example.com".
+   * @param {string} [config.contentSha256] The SHA256 digest value for the request
+   *     content. If you provide this value, the client must provide the HTTP header
+   *     `X-Goog-Content-SHA256` with the same value in its request, and the service
+   *     will verify that the content's SHA256 digest matches that provided here.
```
frankyn: I'd prefer if there was no helper because this doesn't translate to V2 and can be confusing.

test/signer.ts
```diff
@@ -429,12 +429,29 @@ describe('signer', () => {
...
+          CONFIG.contentSha256 = 'sha256-hash';
+          CONFIG.version = 'v4';
+          const EXPECTED_HEADER = `x-goog-content-sha256:${CONFIG.contentSha256}`;
+
+          const signedUrl = await signer.getSignedUrl(CONFIG);
+
+          assert(spy.firstCall.returnValue.endsWith(CONFIG.contentSha256));
+          assert(spy.firstCall.args[3].includes(EXPECTED_HEADER));
+
+          assert(signedUrl.includes('X-Goog-SignedHeaders=host%3Bx-goog-content-sha256'));
```
frankyn: There's a conformance test for this specific case on https://github.com/googleapis/conformance-tests/blob/master/storage/v1/v4_signatures.json#L217

src/file.ts
jkwlui: So we'll leave the responsibility to the user's side to provide the `X-Goog-Content-SHA256` header and we will populate the signed payload section of the canonical request with the provided SHA256. Does that sound okay?

frankyn: +1, I understand this to be an advanced feature. The client should replace `UNSIGNED-PAYLOAD` with the value of header `X-Goog-Content-SHA256` when it's provided by the user.

jkwlui: Thank you!

test/signer.ts
jkwlui: Gotcha. I still would like a unit test to make sure proper things are passed around. Thanks.

PR #1072, opened 2020-02-12
feat(v4): GA master branch

jkwlui:
WIP: DO NOT MERGE.

### DIFF
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -1,0 +1,350 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as crypto from 'crypto';
+import * as dateFormat from 'date-and-time';
+import * as http from 'http';
+import * as url from 'url';
+import {encodeURI, qsStringify, objectEntries} from './util';
+
+interface GetCredentialsResponse {
+  client_email?: string;
+}
+
+export interface AuthClient {
+  sign(blobToSign: string): Promise<string>;
+  getCredentials(): Promise<GetCredentialsResponse>;
+}
+
+export interface BucketI {
+  name: string;
+}
+
+export interface FileI {
+  name: string;
+}
+
+interface GetSignedUrlConfigInternal {
+  expiration: number;
+  method: string;
+  extensionHeaders?: http.OutgoingHttpHeaders;
+  cname?: string;
+  contentMd5?: string;
+  contentType?: string;
+  bucket: string;
+  file?: string;
+}
+
+interface SignedUrlQuery {
+  generation?: number;
+  'response-content-type'?: string;
+  'response-content-disposition'?: string;
+}
+
+interface V2SignedUrlQuery extends SignedUrlQuery {
+  GoogleAccessId: string;
+  Expires: number;
+  Signature: string;
+}
+
+interface V4SignedUrlQuery extends V4UrlQuery {
+  'X-Goog-Signature': string;
+}
+
+interface V4UrlQuery extends SignedUrlQuery {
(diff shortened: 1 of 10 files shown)


test/signer.ts
```diff
@@ -0,0 +1,504 @@
...
+    beforeEach(() => {
+      authClient = {
+        sign: async (_blobToSign: string) => 'signature',
+        getCredentials: async () => ({client_email: CLIENT_EMAIL}),
+      };
+      bucket = {name: BUCKET_NAME};
+      file = {name: FILE_NAME};
+    });
+
+    it('should construct a UrlSigner', () => {
```
stephenplusplus: Just for style-consistency, we would normally wrap constructor tests in a describe:

```js
describe('UrlSigner', () => {
  it('should localize authClient', () => {});

  it('should localize bucket', () => {});

  it('should localize file', () => {});
});
```

Seems excessive when you compare it to just the single test, but multiple tests should highlight to other maintainers that if they change something or add a new property, it needs another test.

src/signer.ts
```diff
@@ -0,0 +1,416 @@
...
+
+const SEVEN_DAYS = 604800;
+
+/**
+ * @const {string}
+ * @private
+ */
+export const PATH_STYLED_HOST = 'https://storage.googleapis.com';
+
+export class UrlSigner {
```
stephenplusplus: Should we:

```diff
- UrlSigner
+ URLSigner
```

? Similar examples, `encodeURIComponent`, `compute.getVMs()`.

test/signer.ts
```diff
@@ -0,0 +1,504 @@
...
+      });
+
+      describe('cname', () => {
+        it('should use a provided cname', async () => {
+          const host = 'http://www.example.com';
+          const configWithCname = Object.assign({cname: host}, CONFIG);
+
+          const signedUrl = await signer.getSignedUrl(configWithCname);
+          const expires = Math.round(Number(CONFIG.expires) / 1000);
+          const expected =
```
stephenplusplus: Checking against the completed URL means the test from `describe('getSignedUrl')` is actually testing behavior in `getSignedUrlV2`. Can we intercept the call to `getSignedUrlV2(config)` and verify config.cname is set to the user-provided input?

Then the same system throughout the rest of the tests in this file, so we can stay as local as possible to the function we are testing. It looks like we'd want new `describe` blocks for:

- getSignedUrlV2
- getSignedUrlV4
- getCanonicalHeaders
- getCanonicalRequest
- getCanonicalQueryParams
- getResourcePath
- parseExpires

test/signer.ts
```diff
@@ -0,0 +1,504 @@
...
+        sign: async (_blobToSign: string) => 'signature',
+        getCredentials: async () => ({client_email: CLIENT_EMAIL}),
+      };
+      bucket = {name: BUCKET_NAME};
+      file = {name: FILE_NAME};
+    });
+
+    it('should construct a UrlSigner', () => {
+      const signer = new UrlSigner(authClient, bucket, file);
+      assert.strictEqual(signer['authClient'], authClient);
```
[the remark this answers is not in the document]
jkwlui: `authClient` is private in `UrlSigner`, using `[]` notation to get around.

src/signer.ts
jkwlui: refactored!

test/signer.ts
jkwlui: done!

test/signer.ts
jkwlui: Make sense! I refactored all the tests in URLSigner so they are localized. Thanks @stephenplusplus

src/signer.ts
stephenplusplus: Oh no, I forgot we have "getSignedUrl()" and not "getSignedURL()". I guess we'll just have a little of both :)

test/signer.ts
```diff
@@ -0,0 +1,886 @@
...
+          bucket: bucket.name,
+        };
+      });
+
+      it('should fail for expirations beyond 7 days', () => {
+        CONFIG.expiration = NOW.valueOf() + 7.1 * 24 * 60 * 60;
+
+        assert.throws(() => {
+          signer['getSignedUrlV4'](CONFIG);
+        }, /Max allowed expiration is seven days/);
```
stephenplusplus: Can you check the full error message, `Max allowed expiration is seven days (${SEVEN_DAYS} seconds).`?

jkwlui:
@stephenplusplus Thank you for your review!

I addressed your feedback, please take a look when you have a moment.

stephenplusplus:
Thank you very much, looks great to me!

PR #1074, opened 2020-02-13
feat: enable bytes read tracking

### LINKED ISSUE #956: progress tracking when uploading big file
Maybe I missed something but couldn't find a way to track progress during uploading. Is it possible to do it now?

I think it would be great to have an option for example:

`const options = {
   onProgress: (data) => {
          console.log(data.loadedBytes)
       }
  };`

`await storage.bucket(bucketName).upload(filename, options);`

AVaksman:
Design: Changes to `file.createWriteStream()`
- Introduce a `ProgressStream` class that extends from native `Trasform`
   - on every chunk emits a `'progress'` event with total calculated `bytesRead` value.
- add an instance of ProgressStream to the pipe.
- If `options.onUploadProgress` is provided by user add a `'progress'` listener that passes the total 
   `bytesRead` value to `options.onUploadProgress`.

Fixes #956


src/file.ts
```diff
@@ -1752,12 +1754,20 @@ class File extends ServiceObject<File> {
...
+      });
+    }
+
     const fileWriteStream = duplexify();
 
     const stream = streamEvents(
       pumpify([
         gzip ? zlib.createGzip() : through(),
         validateStream,
+        progressStream,
```
stephenplusplus: Referring back to an old comment (https://github.com/AVaksman/nodejs-storage/pull/12#discussion_r364463846), was it possible to move this below the write stream?

src/file.ts
```diff
@@ -1752,12 +1754,20 @@ class File extends ServiceObject<File> {
       md5,
     });
 
+    const progressStream = new ProgressStream();
+    if (options.onUploadProgress) {
+      progressStream.on('progress', bytesRead => {
+        options.onUploadProgress!({bytesRead});
```
stephenplusplus: [previous comment](https://github.com/AVaksman/nodejs-storage/pull/12#discussion_r364464162)

We should just forward this event to the returned stream, rather than the handler (IMO). Keeps with the streaming API, so we don't have to mix callbacks/streams.

From `bucket#upload`, it should accept `onUploadProgress` there:

```js
Bucket.prototype.upload = function (opts) {
  file.createWriteStream()
    .on('progress', opts.onUploadProgress)
}
```

But I don't believe `file#createWriteStream` should accept the option itself.

system-test/storage.ts
```diff
@@ -2791,6 +2792,21 @@ describe('storage', () => {
     });
   });
 
+  describe('bucket upload with progress', () => {
+    it('show bytes sent', async () => {
```
stephenplusplus: This test might need to be reworked, because it looks like it requires a human to watch the output and verify there is output? I think you could just check that the `bytesRead` argument in the `onUploadProgress` event is `> 0 && <= fileSize`.

src/file.ts
stephenplusplus: Could we also just pass back `bytesRead` as opposed to boxing it in an object? Or if there's anything else we should pass back, let's do it now.

stephenplusplus (changes requested):

stephenplusplus:
@Avaksman I forgot about this: https://github.com/googleapis/gcs-resumable-upload/blob/4cec8f74a408ea8dc03e3395555189fa6b09add0/src/index.ts#L451

Looks like we could just forward that event to the user (during resumable uploads, anyway)

AVaksman:
> @AVaksman I forgot about this: https://github.com/googleapis/gcs-resumable-upload/blob/4cec8f74a408ea8dc03e3395555189fa6b09add0/src/index.ts#L451
> 
> Looks like we could just forward that event to the user (during resumable uploads, anyway)

That's good!
Should we worry about the simple uploads as well (for consistency)?

stephenplusplus:
> Should we worry about the simple uploads as well (for consistency)?

Yes, I suppose we should. What do you think?

AVaksman:
Blocked by on googleapis/nodejs-common/pull/540 to merge

src/file.ts
AVaksman: With new changes `prgressStream` is not used any more.

src/file.ts
AVaksman: with new changes `bucket#upload` and `file#save` accept `onUploadProgress`

system-test/storage.ts
AVaksman: Reworked!

src/file.ts
AVaksman: Done

src/file.ts
```diff
@@ -1698,7 +1701,7 @@ class File extends ServiceObject<File> {
       pumpify([
         gzip ? zlib.createGzip() : through(),
         validateStream,
-        fileWriteStream,
+        fileWriteStream.on('progress', evt => stream.emit('progress', evt)),
```
stephenplusplus: Can you put this event handler assignment on its own line? Just makes it easier to see what the code does line-by-line.

stephenplusplus (approved):
Just a subjective suggestion, but LGTM.

src/file.ts
```diff
@@ -1701,7 +1701,9 @@ class File extends ServiceObject<File> {
       pumpify([
         gzip ? zlib.createGzip() : through(),
         validateStream,
-        fileWriteStream.on('progress', evt => stream.emit('progress', evt)),
+        fileWriteStream.on('progress', evt => {
```
stephenplusplus: Sorry, I was thinking more like:

```ts
fileWriteStream.on('progress', evt => {
  stream.emit('progress', evt);
});

const stream = streamEvents(
  pumpify([
    gzip ? zlib.createGzip() : through(),
    validateStream,
    fileWriteStream,
  ])
) as Duplex;
```

AVaksman:
@bcoe 
Now that [@google-common v3](https://github.com/googleapis/nodejs-common/pull/542) is released and add `progress events` from simple upload are available ([#540](https://www.github.com/googleapis/nodejs-common/issues/540)) ([1834059](https://www.github.com/googleapis/nodejs-common/commit/18340596ecb61018e5427371b9b5a120753ec003)) all system tests are passing.
However this PR should not be merged until Node 8 is depricated.

stephenplusplus (approved):

bcoe (approved):
I think this feature is really slick, if @stephenplusplus has given his 👍 I say ship it.

gajus:
There does not seem to be a way to track progress using the [`upload` method](https://googleapis.dev/nodejs/storage/latest/Bucket.html#upload).

gajus:
I was wrong. `onUploadProgress` is supported, but not documented. Seems like a documentation bug.

PR #1079, opened 2020-02-18
feat: allow date to be specified for v4 signed URLs

### LINKED ISSUE #1041: Add option to specify date in getSignedUrl
https://cloud.google.com/storage/docs/access-control/signed-urls

From the documentation, signed urls support the following headers.
```
X-Goog-Date: The date and time the signed URL became usable, in the ISO 8601 basic format YYYYMMDD'T'HHMMSS'Z'.

X-Goog-Expires: The length of time the signed URL remained valid, measured in seconds from the value in X-Goog-Date. In this example the Signed URL expires in 15 minutes. The longest expiration value is 604800 seconds (7days).
```

However, with respect to the javascript api, only the expiration is a valid input

https://github.com/googleapis/nodejs-storage/blob/b1c60bbfb11319a6543b0f137e3c2e3d1db9dd11/src/file.ts#L68

```
export interface GetSignedUrlConfig {
  action: 'read' | 'write' | 'delete' | 'resumable';
  version?: 'v2' | 'v4';
  cname?: string;
  contentMd5?: string;
  contentType?: string;
  expires: string | number | Date;
  extensionHeaders?: http.OutgoingHttpHeaders;
  promptSaveAs?: string;
  responseDisposition?: string;
  responseType?: string;
}
```

In the code, `X-Goog-Date` is always set to `Date.now` and `X-Goog-Expires` is calculated relative to `Date.now`.

I propose adding an option to specify the the `date` field and default it to `Date.now`. 

My use case is, for a short time, I would like the same call to `getSignedUrl` to return the same value. As is, this is impossible, since `Date.now` is forever changing.

I can take a stab at adding this feature if it is accepted.

rhodgkins:

Fixes #1041 🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -72,6 +72,7 @@ export interface GetSignedUrlConfig {
   contentMd5?: string;
   contentType?: string;
   expires: string | number | Date;
+  usableFrom?: string | number | Date;
   extensionHeaders?: http.OutgoingHttpHeaders;
   promptSaveAs?: string;
   responseDisposition?: string;
@@ -80,6 +81,7 @@ export interface GetSignedUrlConfig {
 
 interface GetSignedUrlConfigInternal {
   expiration: number;
+  usableFrom: Date;
   method: string;
   name: string;
   resource?: string;
@@ -2409,6 +2411,10 @@ class File extends ServiceObject<File> {
    *     signed url.
    * @param {string} [config.responseType] The response-content-type parameter
    *     of the signed url.
+   * @param {*} [config.usableFrom=Date.now()] A timestamp when this link became usable. Any value
+   *     given is passed to `new Date()`.
+   *     Note: Use for 'v4' only.
+   *     See [reference]{@link https://cloud.google.com/storage/docs/access-control/signed-urls#example}
    * @param {GetSignedUrlCallback} [callback] Callback function.
    * @returns {Promise<GetSignedUrlResponse>}
    *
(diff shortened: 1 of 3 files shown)


src/file.ts
```diff
@@ -72,6 +72,7 @@ export interface GetSignedUrlConfig {
   contentMd5?: string;
   contentType?: string;
   expires: string | number | Date;
+  usableFrom?: string | number | Date;
```
frankyn: Please add a system test to confirm this works with the service.

rhodgkins: Done - also added conformance test case

AVaksman:
The [documentation](https://cloud.google.com/storage/docs/access-control/signed-urls#example) reads.

> X-Goog-Date: The date and time the signed URL **became usable**, in the ISO 8601 basic format YYYYMMDD'T'HHMMSS'Z'

Does that mean that if `X-Goog-Date:` header is set with a date and time _**in the future**_ the `signed url` should fail if used prior to `X-Goog-Date:` set date?

If yes, than we should system-test for that.
@frankyn

conformance-test/test-data/v4SignedUrl.json
```diff
@@ -41,6 +41,16 @@
         "timestamp": "20190301T090000Z",
         "expectedUrl": "https://storage.googleapis.com/test-bucket/test-object?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=test-iam-credentials%40dummy-project-id.iam.gserviceaccount.com%2F20190301%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20190301T090000Z&X-Goog-Expires=20&X-Goog-SignedHeaders=host&X-Goog-Signature=9669ed5b10664dc594c758296580662912cf4bcc5a4ba0b6bf055bcbf6f34eed7bdad664f534962174a924741a0c273a4f67bc1847cef20192a6beab44223bd9d4fbbd749c407b79997598c30f82ddc269ff47ec09fa3afe74e00616d438df0d96a7d8ad0adacfad1dc3286f864d924fe919fb0dce45d3d975c5afe8e13af2db9cc37ba77835f92f7669b61e94c6d562196c1274529e76cfff1564cc2cad7d5387dc8e12f7a5dfd925685fe92c30b43709eee29fa2f66067472cee5423d1a3a4182fe8cea75c9329d181dc6acad7c393cd04f8bf5bc0515127d8ebd65d80c08e19ad03316053ea60033fd1b1fd85a69c576415da3bf0a3718d9ea6d03e0d66f0"
     },
+    {
```
frankyn: Remove this as it's not necessary here.

system-test/storage.ts
```diff
@@ -3277,6 +3277,19 @@ describe('storage', () => {
       assert.strictEqual(body, localFile.toString());
     });
 
+    it('should create a signed read url with usableFrom', async () => {
```
frankyn: Pending reply from the Storage backend team. I'll get back to you as soon as I get an update.

JustinBeckwith:
@frankyn @jkwlui this PR is getting crusty :) What are the plans here?

frankyn:
Apologies for the delay, we can move forward with this change. @rhodgkins could you resolve conflicts with master?

shaffeeullah:
Feature is implemented by https://github.com/googleapis/nodejs-storage/pull/1328

PR #1085, opened 2020-02-27
feat(bucket): support single source in combine (#1076)

### LINKED ISSUE #1076: Allow 1 source inputs in Bucket.combine
The [Bucket.combine](https://googleapis.dev/nodejs/storage/latest/Bucket.html#combine) operation requires 2 or more sources.

It'd be great if the constraint was changed to 1 or more sources. It's allowed by the API and can be done in `gsutil compose`.

I would find this very useful, because then I don't need to write a special case to handle 1 source inputs.

surjikal:
Fixes #1076
 
tldr: This is already supported by `gsutil compose` and the api.


test/bucket.ts
```diff
@@ -473,14 +473,14 @@ describe('Bucket', () => {
   });
 
   describe('combine', () => {
-    it('should throw if invalid sources are not provided', () => {
```
surjikal: Fixed inaccuracy with the error message.

surjikal:
Rebased

stephenplusplus:
@frankyn do you think we should support this here?

jkwlui:
> @frankyn do you think we should support this here?

I think it is fine here, since others have supported this behaviour.

surjikal:
Thanks!

PR #1116, opened 2020-03-20
samples: add v4 signed policy sample

frankyn:


samples/generateV4SignedPolicy.js
```diff
@@ -0,0 +1,64 @@
...
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage();
+
+  async function generateV4SignedPolicy() {
+    // These options will allow temporary uploading of a file
+    // through an HTML form.
+    const file = myBucket.file('my-file');
```
jkwlui: myBucket needs to be defined; use `filename` instead of `my-file`

```javascript
const bucket = storage.bucket(bucketName);
const file = bucket.file(filename)
```

samples/generateV4SignedPolicy.js
```diff
@@ -0,0 +1,64 @@
...
+    const expires = Date.now() + 600 * 1000; //  10 minutes
+    const options = {
+      expires,
+      fields: {'x-goog-meta-test': 'data'},
+    };
+
+    // Get a v4 signed URL for uploading file
+    await file.getSignedPolicyV4(options, function(err, policy) {
+      // Create an HTML form with the provided policy
+      console.log(`<form action='${policy.url}' method='POST' enctype="multipart/form-data">\n`);
```
jkwlui: This line and below:
`console.log` comes with carriage return, so `\n` is not needed here :)

samples/generateV4SignedPolicy.js
```diff
@@ -0,0 +1,64 @@
...
+    // through an HTML form.
+    const file = myBucket.file('my-file');
+    const expires = Date.now() + 600 * 1000; //  10 minutes
+    const options = {
+      expires,
+      fields: {'x-goog-meta-test': 'data'},
+    };
+
+    // Get a v4 signed URL for uploading file
+    await file.getSignedPolicyV4(options, function(err, policy) {
```
jkwlui: here we're mixing async/await with callback.

Change to:
```javascript
const [response] = await file.getSignedPolicyV4(options);

// do something with
response.url
response.fields
```

jkwlui:
@frankyn patched up the comments and made tests pass. PTAL!

samples/generateV4SignedPolicy.js
```diff
@@ -0,0 +1,69 @@
...
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main(bucketName = 'my-bucket', filename = 'test.txt') {
+  // [START storage_generate_post_policy_url_v4]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // const bucketName = 'Name of a bucket, e.g. my-bucket';
+  // const filename = 'File to access, e.g. file.txt';
```
frankyn: File to create

samples/generateV4SignedPolicy.js
```diff
@@ -0,0 +1,69 @@
...
+      `<form action='${response.url}' method='POST' enctype="multipart/form-data">`
+    );
+    // Include all fields returned in the HTML form as they're required
+    for (const name of Object.keys(response.fields)) {
+      const value = response.fields[name];
+      console.log(`  <input name='${name}' value='${value}' type='hidden'/>`);
+    }
+    console.log("  <input type='file' name='file'/>");
+    console.log("  <input type='submit' value='Upload File' name='submit'/>");
+    console.log('</form>');
```
frankyn: Could you instead return the form as a string?

PR #1120, opened 2020-03-20
types: split out SignedUrlResponse

jkwlui:
VS code complained of duplicate local variable with the referencing to self in #1119. I think this is a little bit cleaner.


stephenplusplus:
Way better, thank you!

PR #1135, opened 2020-03-28
fix(bucket): Only disable resumable uploads for bucket.upload (fixes #1133)

### LINKED ISSUE #1133: bucket.upload doesn't retry failed resumable requests
#### Environment details

  - OS: any
  - Node.js version: any
  - npm version: any
  - `@google-cloud/storage` version: 4.6.0

#### Steps to reproduce

  1. Use `bucket.upload` with a file size over the [`RESUMABLE_THRESHOLD `](https://github.com/googleapis/nodejs-storage/blob/6833dfe9bf013a582ff46620b8035fc47ec38d0f/src/bucket.ts#L372) to upload
  2. If `/home/.config` is not writable the upload fails

Do the same steps but with `file.createWriteStream` and a file stream and the upload works.

From looking at the code, `bucket.upload` sets [`options.resumable`](https://github.com/googleapis/nodejs-storage/blob/6833dfe9bf013a582ff46620b8035fc47ec38d0f/src/bucket.ts#L3457) and that means the [retry check](https://github.com/googleapis/nodejs-storage/blob/6833dfe9bf013a582ff46620b8035fc47ec38d0f/src/file.ts#L1723) in `file.createWriteStream` doesn't work.

Note I discovered this issue by trying a resumable upload on Cloud Run which doesn't allow writing to `/home/.config` which doesn't seem right!

rhodgkins:

Fixes #1133 

I've added a test case for this by stubbing `fs.stat` - I hope that's ok!

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -3454,7 +3454,10 @@ class Bucket extends ServiceObject {
           return;
         }
 
-        options.resumable = fd.size > RESUMABLE_THRESHOLD;
+        if (fd.size <= RESUMABLE_THRESHOLD) {
+          // Only disable resumable uploads so createWriteStream still attempts them and falls back to simple upload.
+          options.resumable = false;
+        }
 
         upload();
       });
--- a/test/testdata/dummylargetestfile.txt
+++ b/test/testdata/dummylargetestfile.txt
@@ -1,3 +1,0 @@
-This is a test file!
-
-When using fs.stat should be stubbed to return a large file size
--- a/test/testdata/dummylargetestfile.txt
+++ b/test/testdata/dummylargetestfile.txt
@@ -1,0 +1,3 @@
+This is a test file!
+
+When using fs.stat should be stubbed to return a large file size
--- a/test/bucket.ts
+++ b/test/bucket.ts
@@ -2381,19 +2381,74 @@ describe('Bucket', () => {
       bucket.upload(textFilepath, options, assert.ifError);
     });
 
-    it('should force a resumable upload', done => {
-      const fakeFile = new FakeFile(bucket, 'file-name');
-      const options = {destination: fakeFile, resumable: true};
-      fakeFile.createWriteStream = (options_: CreateWriteStreamOptions) => {
-        const ws = new stream.Writable();
-        ws.write = () => true;
-        setImmediate(() => {
-          assert.strictEqual(options_.resumable, options.resumable);
-          done();
-        });
-        return ws;
-      };
-      bucket.upload(filepath, options, assert.ifError);
+    describe('resumable uploads', () => {
+      const dummyLargeFilepath = path.join(
+        __dirname,
+        '../../test/testdata',
+        'dummylargetestfile.txt'
+      );
+      const dummyLargeFileStat = require('fs').statSync(dummyLargeFilepath);
+      // Set size greater than threshold
+      dummyLargeFileStat.size = 5000001;
+
+      let sandbox: sinon.SinonSandbox;
+
+      beforeEach(() => {
+        sandbox = sinon.createSandbox();
+
+        const statStub = sandbox.stub(require('fs'), 'stat');
+        statStub
+          .withArgs(dummyLargeFilepath)
+          .callsArgWithAsync(1, null, dummyLargeFileStat);
+        statStub.callThrough();
+      });
+
+      afterEach(() => sandbox.restore());
+
+      it('should force a resumable upload', done => {
+        const fakeFile = new FakeFile(bucket, 'file-name');
+        const options = {destination: fakeFile, resumable: true};
+        fakeFile.createWriteStream = (options_: CreateWriteStreamOptions) => {
+          const ws = new stream.Writable();
+          ws.write = () => true;
+          setImmediate(() => {
+            assert.strictEqual(options_.resumable, options.resumable);
+            done();
+          });
+          return ws;
+        };
+        bucket.upload(filepath, options, assert.ifError);
+      });
+
+      it('should not pass resumable option to createWriteStream when file size is greater than minimum resumable threshold', done => {
+        const fakeFile = new FakeFile(bucket, 'file-name');
+        const options = {destination: fakeFile};
+        fakeFile.createWriteStream = (options_: CreateWriteStreamOptions) => {
+          const ws = new stream.Writable();
+          ws.write = () => true;
+          setImmediate(() => {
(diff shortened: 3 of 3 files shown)


src/bucket.ts
```diff
@@ -3454,7 +3454,10 @@ class Bucket extends ServiceObject {
           return;
         }
 
-        options.resumable = fd.size > RESUMABLE_THRESHOLD;
+        if (fd.size <= RESUMABLE_THRESHOLD) {
+          // Only disable resumable uploads so createWriteStream still attempts them and falls back to simple upload.
```
stephenplusplus: Doesn't this have the same effect as the removed code?

rhodgkins: It doesn't explicitly set `resumable` to `true` so when the below check is done (Line 1723), it will fallback to a simple upload (I've mentioned this in more detail in [#1133](https://github.com/googleapis/nodejs-storage/issues/1133#issuecomment-605323416)

https://github.com/googleapis/nodejs-storage/blob/6833dfe9bf013a582ff46620b8035fc47ec38d0f/src/file.ts#L1721-L1741

stephenplusplus: Sorry, I'm caught up now :)

This change would ultimately opt the user out of resumable upload behavior by default when they pass a large file. That would mean the threshold concept wouldn't have a purpose, but it has been working for us so far.

The error returned about the directory that isn't writable isn't a retryable error, since resumable upload relies on using that directory to persist data that allow it to resume the upload if it fails during transmission. The message suggests using "resumable: false", but that's only a suggestion, as the user may prefer to resolve the error and gain the benefits of a resumable upload.

rhodgkins: >This change would ultimately opt the user out of resumable upload behavior by default when they pass a large file

It wouldn't - from my understanding, passing `resumable=false` to `File.createWriteStream` would be the only thing which prevents a resumable upload? `File.createWriteStream` always tries to do a resumable upload unless `resumable` is `false`.

This change is just so that `Bucket.upload` performs the same as `File.createWriteStream` when no `resumable` option is passed to `File.createWriteStream`. `File.createWriteStream` currently ignores the error returned about the directory that isn't writable and continues to a simple upload, shouldn't `Bucket.upload` match this?

stephenplusplus: > shouldn't Bucket.upload match this?

That's a fair question. `createWriteStream()` defaults to resumable because it's the best-case scenario for a user. But, it doesn't know the incoming data size, so if it turns out a resumable upload doesn't work, then "oh well, maybe we were wrong and the file isn't that big anyway" and a non-resumable upload is fine. With `bucket.upload()`, we know the file size, and if it's large enough that we think resumable would be best, we opt the user into it, as opposed to being passive about it.

I could go either way here, I think. It's nice to just make things work without obstructions or annoying errors, but it's hard to know if we would be disappointing other users who would prefer to have the perks of resumable uploads, and appreciate the error.

@jkwlui @frankyn @bcoe any thoughts?

jkwlui: I am leaning towards matching `bucket.upload()`'s behaviour with `file.createWriteStream()`'s so we reduce the complexity of different upload strategies, but @stephenplusplus's point about informing user about the advantage of resumable uploads is also very valid - I wonder if better documentation about the benefit of explicitly opting resumable upload and having the user check that the config dir is writable in their environment would be a fair middle ground here. 

Nevertheless, we should probably align our libraries strategies across languages. Summarizing the current behaviour for those who's just getting caught up with this change:

For Bucket.upload() where the file size is known in advance, we default resumable upload to true if `size > RESUMABLE_THRESHOLD`, which will result in error if the config dir is not writable (like in Cloud Run as pointed out by @rhodgkins ).

Whereas for `file.createWriteStream()` where a user is expected to pipe a readable stream to and the file size is *not* known in advance, we default to simple upload if not specified.

@frankyn

stephenplusplus: > Whereas for file.createWriteStream() where a user is expected to pipe a readable stream to and the file size is not known in advance, we default to simple upload if not specified.

We actually default to resumable, but just silently fallback to a simple upload if it didn't work.

jkwlui: Thanks, edited the comment.

JustinBeckwith:
@jkwlui any movement on this PR?

jkwlui:
@stephenplusplus let's move forward with this PR, if everything looks okay to you.

stephenplusplus:
Thanks @rhodgkins! Great work on the tests 👍

stephenplusplus:
@rhodgkins would you mind updating the PR against our latest master? You might also have the option on this page to give us write access to your fork, which would allow us to make the change. (I'm not sure that's a thing, but I recall seeing the option somewhere before)

rhodgkins:
@stephenplusplus done (sorry about the noise, I tired to update via GitHub UI and it wasn't the email I signed the CLA with, so I rebased and force pushed...!)

stephenplusplus:
The linter had a small issue with how `fs` was being imported, but when I looked closer at the tests, I thought I would do a small refactor of the mocking logic to sync with how we're currently doing it. Here's a diff you can apply, which should 🤞 get us green lights from CI.

Here's a gist: https://gist.github.com/stephenplusplus/3dd358c7f98aeca2b55a7355750a3e5f

And to quickly apply it:
```sh
$ curl https://gist.githubusercontent.com/stephenplusplus/3dd358c7f98aeca2b55a7355750a3e5f/raw/05525a9344c6151b1e7d9b362980dceb658f58f3/storage-1135.diff | git apply 
`

**Comment by rhodgkins** [comment]
@stephenplusplus patch applied - also rebased from master

File: test/testdata/dummylargetestfile.txt [hunk 2]
```diff
@@ -0,0 +1,3 @@
+This is a test file!
```
stephenplusplus, line 1: Sorry I missed this @rhodgkins. We can remove this file now.

rhodgkins:
@stephenplusplus done! Though no idea why `npm install` on windows is failing...

stephenplusplus:
Yeah, I'm stumped on that, too. I am working on a workaround at the moment. Sorry for the delay getting this merged. I'll update soon.

stephenplusplus:
The fix should be in master. Can you update with master one last time? 🙏

rhodgkins:
@stephenplusplus nice - all sorted

stephenplusplus:
Phew! Thank you very much for finding the problem, the fix, and the patience to deal with us :)

PR #1156, opened 2020-04-12
build!: drop support for node.js 8.x

JustinBeckwith:


package.json
```diff
@@ -5,17 +5,14 @@
...
   "repository": "googleapis/nodejs-storage",
   "main": "./build/src/index.js",
   "types": "./build/src/index.d.ts",
   "files": [
     "build/src",
-    "!build/src/**/*.map",
-    "AUTHORS",
-    "CONTRIBUTORS",
-    "COPYING"
+    "!build/src/**/*.map"
```
bcoe, line 15: I've been wanting to start shipping source-maps again, as it can help significantly with debugging (I've been finding it very useful with the `--enable-source-map` flag occasionally.

JustinBeckwith: +1 to that.  It's @soldair you have to convince!

PR #1161, opened 2020-04-22
fix: StorageOptions.apiEndpoint overrides simple and resumable uploads

### LINKED ISSUE #898: Simple file upload doesn't respect apiEndpoint
File uploading does not respect `apiEndpoint` or the environment variable `STORAGE_EMULATOR_HOST`. Instead it uses a hardcoded constant `STORAGE_UPLOAD_BASE_URL`

#### Environment details
  - OS: `Mac 10.14.6`
  - Node.js version: `12.12.0`
  - npm version: `6.11.3`
  - `@google-cloud/storage` version: `4.0.0`

jkwlui:
- `StorageOptions.apiEndpoint` overrides simple upload url, if set.


src/storage.ts
```diff
@@ -388,9 +389,15 @@ export class Storage extends Service {
     options = Object.assign({}, options, {
       apiEndpoint: options.apiEndpoint || 'storage.googleapis.com',
     });
-    const url =
-      process.env.STORAGE_EMULATOR_HOST ||
-      `https://${options.apiEndpoint}/storage/v1`;
+
+    // Overrides options.apiEndpoint.
```
stephenplusplus, line 393: If the user provides "options.apiEndpoint", that should override the env var. Similar to options.keyFilename beating ADC.

src/storage.ts
```diff
@@ -388,9 +389,15 @@ export class Storage extends Service {
...
+    const EMULATOR_HOST = process.env.STORAGE_EMULATOR_HOST;
```
stephenplusplus, line 394: Let's document the format expected for the env var. It looks like it needs to be a full protocol://host/path, however in the following line, the protocol will be lost.

To compare, nodejs-datastore accepts 'https://localhost:8080' or 'localhost:8080'

https://github.com/googleapis/nodejs-datastore/blob/68790d26d69f48d780930f90be6758f7431f61c9/src/index.ts#L97
https://github.com/googleapis/nodejs-datastore/blob/68790d26d69f48d780930f90be6758f7431f61c9/src/index.ts#L866

Can/should we accept both here?

src/storage.ts
```diff
@@ -388,9 +389,15 @@ export class Storage extends Service {
...
+    if (typeof EMULATOR_HOST === 'string') {
+      options.apiEndpoint = new URL(EMULATOR_HOST).host;
```
stephenplusplus, line 396: Maybe we should just accept raw user-input-- the whole thing, protocol, port, and path. We can also append "/storage/v1" to it to take the path of least surprise for the user (and less cognitive overload for us).

stephenplusplus (changes requested):

src/storage.ts
jkwlui: I'm not really sure what we want to do here -

So we have `STORAGE_EMULATOR_HOST` that provides the full URL to override `baseUrl` upon which all requests are based, except simple upload and resumable upload.

`StorageOptions.apiEndpoint` also does not override simple upload and resumable upload, since it only affects the baseUrl set, as far as I can tell.

Simple upload currently uses `STOARGE_UPLOAD_BASE_URL` which is hardcoded to `https://storage.googleapis.com/upload/storage/v1/b`.

Resumable upload currently uses `https://${this.apiEndpoint}/upload/storage/v1/b`, where `apiEndpoint` defaults to `storage.googleapis.com`, which is overridable through `cfg.apiEndpoint` passed to `resumableUpload.createURI`, but `apiEndpoint` doesn't seem to be passed. 
https://github.com/googleapis/gcs-resumable-upload/blob/master/src/index.ts#L212
https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts#L1520-L1536

stephenplusplus: Thank you for tracking all of those scenarios down. It is a bit messy, eh?

> So we have STORAGE_EMULATOR_HOST that provides the full URL to override baseUrl upon which all requests are based, except simple upload and resumable upload.

AFAIK, `STORAGE_EMULATOR_HOST` is not an official, Google-made env var. So we get a bit of play here to define it ourselves.

We should have it also affect the simple and resumable paths.

> StorageOptions.apiEndpoint also does not override simple upload and resumable upload, since it only affects the baseUrl set, as far as I can tell.

We should make sure it overrides it when `apiEndpoint` is set, too.

> Simple upload currently uses [...] https://storage.googleapis.com/upload/storage/v1/b
> Resumable upload currently uses https://${this.apiEndpoint}/upload/storage/v1/b

Let's just sync these up. 

Overall, what I think would be least surprising for users:

For here-ish: https://github.com/googleapis/nodejs-storage/blob/5a8eea3aeab45d63964e134ea79310c4fddf0c24/src/storage.ts#L391

```ts
const url =
  sanitize(options.apiEndpoint) || 
  sanitize(process.env.STORAGE_EMULATOR_HOST) ||
  'https://storage.googleapis.com'

url += '/storage/v1'

function sanitize(url) {
  // prepend default protocol if necessary
}
```

I know those upload URLs are hardcoded in file.ts, so we'd have to pick them up from the instance instead.

- Users would always handle the full URL path, as opposed to having to figure out which part we cut out
- It would allow both protocol and port to be overridden to match their mock env

Is this solution possible, or maybe not even a good option for what you think users would want?

ojhaujjwal:
Can we please prioritize this? I am affected by this.

jkwlui:
Yes @ojhaujjwal, sorry for the delay!

src/storage.ts
jkwlui: it appears our benchmarking fake serves the API without the '/storage/v1' prefix.

src/file.ts
```diff
@@ -3568,11 +3561,15 @@ class File extends ServiceObject<File> {
       options
     );
 
+    const apiEndpoint = this.storage.apiEndpoint;
+    const bucketName = this.bucket.name;
+    const uri = `https://${apiEndpoint}/upload/storage/v1/b/${bucketName}/o`;
```
stristr, line 3566:
```suggestion
    const uri = `${this.storage.url}/upload/storage/v1/b/${bucketName}/o`;
```

This ensures host protocols other than `https://` work here, in conjunction with googleapis/gcs-resumable-upload#349.

src/storage.ts
stephenplusplus: We could probably change our benchmark mock? I still have the same concerns about what a user would find obvious when it comes to what path we send a request to. The easiest thing IMO is to only allow overriding the section in bold: **protocol://host:port**/path, regardless of when path starts with "storage/v1" or anything else.

jkwlui:
@stephenplusplus can you help take a look so I'm not missing anything?

src/storage.ts
```diff
@@ -417,6 +429,13 @@ export class Storage extends Service {
     this.getHmacKeysStream = paginator.streamify('getHmacKeys');
   }
 
+  private static sanitizeEndpoint(url: string) {
```
stephenplusplus, line 432: This should be tested in its own "describe('sanitizeEndpoint')". (You could probably just copy gcs-resumable-upload's)

src/storage.ts
stephenplusplus: @jkwlui Is this how it works currently? ^ the part about replacing the section in bold.

src/storage.ts
stephenplusplus: @jkwlui

stephenplusplus:
@jkwlui I pinged you in a couple of threads in this PR. It looks good to me after those.

src/storage.ts
jkwlui: Thanks for bringing it up @stephenplusplus. I've decided to keep `STORAGE_EMULATOR_HOST` working as it is right now, since other languages also depend on the benchmark mock. I understand this is inflexible and potentially confusing, but since this should be a private option I think it's okay for now.

stristr:
Hey @jkwlui, one more thing—looks like `startResumableUpload_` is not passing `apiEndpoint` in

```typescript
const uploadStream = resumableUpload.upload(...);
```

stephenplusplus (approved):

rajwilkhu:
Thanks for pushing this change - was going to create a PR to do the same until I saw this. When is this scheduled for release?

stristr:
> When is this scheduled for release?

@rajwilkhu, this was released in version 5.0.1.

PR #1167, opened 2020-04-27
fix: sync to googleapis/conformance-tests@fa559a1

jkwlui:
Updated conformance-tests to reflect signing bucket name into fields.


frankyn:
This will fail until #1168 is merged.

PR #1168, opened 2020-04-27
fix(v4-policy): add missing bucket field

### LINKED ISSUE #1164: Storage Update Conformance Tests
The Storage library support for Post Policy V4 has a security risk that is addressed in the latest PR to Storage conformance tests: https://github.com/googleapis/conformance-tests/pull/31

The fix is to add `{\"bucket\":\"bucket-name\"}` to Post Policy V4 policy which is then signed.

Here's an example solution in Ruby: https://github.com/googleapis/google-cloud-ruby/pull/5702

jkwlui:
fixes #1164

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2586,6 +2586,7 @@ class File extends ServiceObject<File> {
 
       fields = {
         ...fields,
+        bucket: this.bucket.name,
         key: this.name,
         'x-goog-date': nowISO,
         'x-goog-credential': credential,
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -2586,6 +2586,7 @@ class File extends ServiceObject<File> {
 
       fields = {
         ...fields,
+        bucket: this.bucket.name,
```
frankyn, line 2589: Only add bucket to policy and not fields.

PR #1176, opened 2020-05-01
chores(build): add conformance-test to workflow

jkwlui:


jkwlui:
@bcoe can I get your help taking a look at the workflow config?

I'm not sure at what point in time we stopped running conformance test from CI, so this adds it back in.

.github/workflows/conformance-test.yaml
```diff
@@ -0,0 +1,17 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: ci
```
bcoe, line 6: we should probably call this something other than CI, perhaps `conformance`?

PR #1188, opened 2020-05-12
refactor: drop explicit dependency on node-fetch

JustinBeckwith:


bcoe (approved):

stephenplusplus:
It could be a bit disorienting for the user to not recognize the request library. They'll probably be able to sort out what we mean just by looking at the code. However, users hunting down samples are already learning one new thing, so having to spend any more mental cycles to unwind the transport library could be bad timing.

jkwlui (approved):

JustinBeckwith:
eh

PR #1190, opened 2020-05-12
feat!: automatically detect contentType if not provided

### LINKED ISSUE #1184: Automatically set content-type based on file type
While uploading a file to a Storage Bucket, the API automatically sets the content-type to `application/octet-stream` instead of setting it to what it should be. e.g. image/png etc.

Instead of having to manually set this metadata or it being set to `application/octet-stream`, let the API automatically set this based on the file extension.
Check for each file type extension and then set the metadata manually.

atn:
Fixes #1184  🦕

In a rush, I didn't have time to edit tests.ts file. It would be appreciated if someone could!

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -3438,12 +3438,6 @@ class Bucket extends ServiceObject {
       });
     }
 
-    const contentType = mime.contentType(path.basename(pathString));
-
-    if (contentType && !options.metadata.contentType) {
-      options.metadata.contentType = contentType;
-    }
-
     if (options.resumable !== null && typeof options.resumable === 'boolean') {
       upload();
     } else {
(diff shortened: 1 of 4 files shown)


austtn:
@googlebot I signed it!

stephenplusplus:
Thank you! I will take it over :)

src/bucket.ts
```diff
@@ -3438,12 +3438,6 @@ class Bucket extends ServiceObject {
       });
     }
 
-    const contentType = mime.contentType(path.basename(pathString));
```
stephenplusplus, line 3441: Removed as this is identically covered inside of `createWriteStream()` when this code makes its way there.

austtn:
@googlebot I consent.

jkwlui (approved):

PR #1198, opened 2020-05-19
fix(types): assert in typescript 3.7

jkwlui:
CI started to fail overnight on some assertions in our tests. I believe this has to do with [Assertion Functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) in TS3.7.

What I basically did in this PR is simply patching the breakrages, as I'm not sure how to proceed otherwise.

Calling on all node/ts experts @googleapis/yoshi-nodejs

### DIFF
--- a/test/bucket.ts
+++ b/test/bucket.ts
@@ -2069,7 +2069,7 @@ describe('Bucket', () => {
       };
 
       FakeServiceObject.prototype.request = ((
-        reqOpts: DecorateRequestOptions,
+        reqOpts: DecorateRequestOptions
       ) => {
         assert.strictEqual(reqOpts.qs.userProject, USER_PROJECT);
         assert.strictEqual(reqOpts.qs, options.qs);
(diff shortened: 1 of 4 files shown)


test/bucket.ts
```diff
@@ -281,7 +281,9 @@ describe('Bucket', () => {
       });
 
       const bucket = new Bucket(storageInstance, BUCKET_NAME);
-      assert(bucket instanceof ServiceObject);
+      // Using assert.strictEqual instead of assert to prevent
+      // coercing of types.
+      assert.strictEqual(bucket instanceof ServiceObject, true);
```
jkwlui, line 286: This causes the compiler to recognise `bucket` as type `ServiceObject`, rather than `any`, which causes the line below to fail.

I'm starting to think we should probably rewrite our tests in `sinon`.

bcoe (approved):

PR #1208, opened 2020-06-02
feat: allow setting timeouts on uploads

### LINKED ISSUE #1206: Add operation timeout option
We're occasionally seeing very long times (>6minutes) for operations trying to write tiny 90byte files to GCS which then throws the exception mentioned in https://github.com/googleapis/nodejs-storage/issues/833. The services getting the errors are hosted on managed Cloud Run.

The code that (occasionally) triggers the timeout is basically:

````
    const someObject = { key: "value" };
    const options = {
      resumable: false,
      contentType: 'application/json'
    };
    await file.save(JSON.stringify(someObject), options);
````

It would be useful to be able to set the timeouts explicitly to a lower value, e.g. 30 seconds rather than 6 minutes.

### PR DESCRIPTION
**Pull request by stephenplusplus**
Fixes #1206

All of these should work:

```ts
await bucket.upload('file/path.txt', {timeout: 1});

await file.save('content', {timeout: 1});

file.createWriteStream({resumable: false, timeout: 1});
```


jkwlui:
Would it be better if we have the timeout option in a nested object? This separates the options that's related to the request versus those belonging to the API.

Something like:
```js
await bucket.upload('file/path.txt', {reqOpts: {timeout: 1}});
```

stephenplusplus:
I do like the separation of that from a maintainer's perspective, but I think a lot of users just want a logical name for what they need, and "timeout" is about as intuitive as we can go. Otherwise, they'd have to learn and remember "reqOpts" (or whatever we might end up calling it). Some other sibling options exist such as "resumable", "encryptionKey", "userProject" which also affect the request.

If we had designed this in from the start, though, I would very much favor all of our methods accepting a "requestOptions" object that could have control over the request parameters (query string, headers, etc). We have the request interceptors system, which can cover that in most cases, but it's not as easy to use.

JustinBeckwith (approved):

jkwlui (approved):

PR #1215, opened 2020-06-11
fix: sample - cacheControl is set under metadata, not top level as comment suggests

jgresham:

### DIFF
--- a/samples/fileSetMetadata.js
+++ b/samples/fileSetMetadata.js
@@ -45,9 +45,11 @@ function main(bucketName = 'my-bucket', filename = 'file.txt') {
 
         // Note or actionable items for user e.g. uniqueId,
         // object description or other useful information.
+        // cacheControl is set here (default value for public buckets is 'public, max-age=3600')
         metadata: {
           description: 'file description...',
           modified: '1900-01-01',
+          cacheControl: 'no-cache, max-age=0'
         },
       });
 


jkwlui:
`cacheControl` metadata should be on the top level as it exists as part of an Object's [resource](https://cloud.google.com/storage/docs/json_api/v1/objects#resource).

jgresham:
> `cacheControl` metadata should be on the top level as it exists as part of an Object's [resource](https://cloud.google.com/storage/docs/json_api/v1/objects#resource).

@jkwlui Sounds good to me. Just FYI, it currently doesn't work set at the top level using version `"@google-cloud/storage": "^5.0.1",`

jkwlui:
@jgresham can you share the code attempting to set `cacheControl` at the top-level?

I have this and I can verify `cacheControl` is set on the remote object.
```js
  async function setFileMetadata() {
    // Set file metadata.
    const [metadata] = await storage
      .bucket(bucketName)
      .file(filename)
      .setMetadata({
        // Predefinded metadata for server e.g. 'cacheControl', 'contentDisposition',
        // 'contentEncoding', 'contentEncoding', 'contentLanguage', 'contentType'
        contentDisposition: 'attachment; filename*=utf-8\'\'"anotherImage.jpg"',
        contentType: 'image/jpeg',
        cacheControl: 'private',

        // Note or actionable items for user e.g. uniqueId,
        // object description or other useful information.
        metadata: {
          description: 'file description...',
          modified: '1900-01-01',
        },
      });

    console.log(metadata);
  }
```

jgresham:
@jkwlui 

```
const { Storage } = require("@google-cloud/storage")
const storage = new Storage({
  projectId: PROJECT_ID,
})
let imageMetadata = {
      contentType: "image/jpeg",
      // cacheControl: "no-cache, max-age=0",   <--- does not work here
      metadata: {
        // default GCP Storage file caching is "public, max-age=3600"
        // override so that users immediately see new files after updating
        cacheControl: "no-cache, max-age=0",
      },
    }
let imageBucket = storage.bucket("my-bucket")
let bucketFile = imageBucket.file("file-path/file-name")
let writeStream = bucketFile.createWriteStream(imageMetadata)
let writeStreamResp = await promiseWriteStream(imageBuffer, writeStream)
```

jkwlui:
Ah I see.

so from the documentation of `File.createWriteStream()`, the first argument is expected to have type `CreateWriteStreamOptions`. This is **not** the metadata of the object. However with that being said, we did provide convenient options here like `contentType` and `predefinedAcl` which actually maps back to the `metadata` field of the option. See https://github.com/googleapis/nodejs-storage/blob/master/src/file.ts#L1693-L1715

The `metadata` field of `CreateWriteStreamOptions` actually corresponds to the top-level metadata as is expected for `file.setMetadata()`.

jgresham:
Thank you for the details @jkwlui! I closed the PR.

PR #1240, opened 2020-07-09
feat: remove through2 dependency

laljikanjareeya:
Fixes #1239

PR #1238, opened 2020-07-08
docs: update outdated references to signBlob API documentation

kimroen:
Updates the inline reference documentation to point to the new Service Account Credentials API
instead of the deprecated methods in the Cloud IAM API.

This reflects the actual usage of the library.

Fixes #1231

PR #1241, opened 2020-07-09
test: should throw original error message on non JSON response on large metadata

laljikanjareeya:
In recent PR it's failing.
Ref:
#1240 
#1238

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2143,19 +2143,19 @@ describe('storage', () => {
       };
 
       const file = bucket.file('large-metadata-error-test');
-      try {
-        // Save file with metadata size more then 2MB.
-        await file.save('test', {
-          resumable: false,
-          metadata: {
+      await assert.rejects(
+        async () => {
+          await file.save('test', {
+            resumable: false,
             metadata: {
-              custom: largeCustomMeta(2.1e6),
+              metadata: {
+                custom: largeCustomMeta(2.1e6),
+              },
             },
-          },
-        });
-      } catch (err) {
-        assert(err!.message.includes('Metadata part is too large.'));
-      }
+          });
+        },
+        (err: Error) => err.message.indexOf('Metadata part is too large.') > -1
+      );
     });
 
     it('should read a byte range from a file', done => {
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2143,19 +2143,16 @@ describe('storage', () => {
       };
 
       const file = bucket.file('large-metadata-error-test');
-      await assert.rejects(
-        async () => {
-          await file.save('test', {
-            resumable: false,
+      await assert.rejects(async () => {
+        await file.save('test', {
+          resumable: false,
+          metadata: {
             metadata: {
-              metadata: {
-                custom: largeCustomMeta(2.1e6),
-              },
+              custom: largeCustomMeta(2.1e6),
             },
-          });
-        },
-        (err: Error) => err.message.indexOf('Metadata part is too large.') > -1
-      );
+          },
+        });
+      }, /Metadata part is too large/);
     });
 
     it('should read a byte range from a file', done => {
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2143,21 +2143,19 @@ describe('storage', () => {
       };
 
       const file = bucket.file('large-metadata-error-test');
-      // Save file with metadata size more then 2MB.
-      await file.save(
-        'test',
-        {
+      try {
+        // Save file with metadata size more then 2MB.
+        await file.save('test', {
           resumable: false,
           metadata: {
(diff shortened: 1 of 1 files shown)


system-test/storage.ts
```diff
@@ -2144,19 +2144,18 @@ describe('storage', () => {
 
       const file = bucket.file('large-metadata-error-test');
       // Save file with metadata size more then 2MB.
-      await assert.rejects(
-        async () => {
-          await file.save('test', {
-            resumable: false,
+      await file.save(
```
stephenplusplus, line 2147: Let's not mix `await` with the callback usage below. You could use the `done` system, or try/catch.

stephenplusplus (changes requested):

stephenplusplus (approved):

system-test/storage.ts
stephenplusplus: Thanks!

JustinBeckwith (changes requested):
Please use the `test: ` prefix for changes that don't affect users.  Anything with `fix: ` will trigger a release, which is not what we want here

system-test/storage.ts
```diff
@@ -2143,22 +2143,19 @@ describe('storage', () => {
       };
 
       const file = bucket.file('large-metadata-error-test');
-      // Save file with metadata size more then 2MB.
```
JustinBeckwith, line 2146: Please switch this back to an `assert.rejects`.  When you try to do an assert in the catch, there's a chance the method just doesn't throw.  The `await assert.rejects` is careful about making sure both that it throws, and that it gets the right error message.

JustinBeckwith (changes requested):

system-test/storage.ts
stephenplusplus: I was wondering about that. I saw the try/catch used in other tests, so we may have some more work to do across the libraries to be safe.

JustinBeckwith: Yeah, I try to call em out when I see them.  Things were messy until `assert.rejects` made it's way consistently into node core.

stephenplusplus:
> Anything with fix: will trigger a release, which is not what we want here

I agree this is a "test:" as opposed to a "fix:", but am confused about the repercussions of mis-labeling. I assume the downsides of unnecessary release PRs are potential resource exhaustion, in terms of GCP operations and CI overloading. However, I thought there was always a pending release PR. Is that not the case, they only get opened when the first fix/feat is merged? It seems like a narrow slice in time where we sit on master without having an auto-gen'd fix or dependency update, etc. So without knowing more, it sounds like it could be more trouble than its worth to not just let all commits trigger creating a release PR, if necessary?

I'm not heavily invested in either option, this just turned out to be a good time to learn more about how this works, what we're going for, and why.

JustinBeckwith:
Yeah it's a really good question.  The release PR is only created when there are changes on master that would trigger a release - like a `fix` or `feat`.  By landing things with `fix`, it means a few things will happen:
- This will show up in the customer facing changelog that gets autogenerated.  Customers probably don't care that we fixed our own tests :) 
- It will open up the tracking PR

We are trying to keep the number of open PRs to a minimum where we can.  We also want to release bug fixes to developers as quickly as we can after them landing on master, so changes don't pile up.  As a result of both of these, it leads to us often merging release PRs with one or two fixes (a good thing!).  Cutting a bunch of releases that have no customer impacting changes is kinda not great, as it creates a lot of churn.

system-test/storage.ts
laljikanjareeya: @JustinBeckwith  Thanks for bringing this up.

system-test/storage.ts
```diff
@@ -2155,9 +2154,7 @@ describe('storage', () => {
             },
           });
         },
-        {
-          message: 'Cannot parse response as JSON: Metadata part is too large.',
-        }
+        (err: Error) => err.message.indexOf('Metadata part is too large.') > -1
```
JustinBeckwith, line 2157: Should be able to simplify this entire line to:
```js
/Metadata part is too large/
```

That method takes an optional regex, or function to do the check.

stephenplusplus (approved):

system-test/storage.ts
```diff
@@ -2143,22 +2143,16 @@ describe('storage', () => {
       };
 
       const file = bucket.file('large-metadata-error-test');
-      // Save file with metadata size more then 2MB.
-      await assert.rejects(
-        async () => {
-          await file.save('test', {
-            resumable: false,
+      await assert.rejects(async () => {
```
JustinBeckwith, line 2146: You can drop the `async` and `await here!  This is as easy as:
```js
await assert.rejects(
  file.save('test', {
    ....
  })
);
```

JustinBeckwith (approved):

PR #1248, opened 2020-07-16
chore: update createBucketRequest in createNewBucket.js

kembala:
Example is outdated as the parameter type is:
}
```
export interface CreateBucketRequest {
    archive?: boolean;
    coldline?: boolean;
    cors?: Cors[];
    dra?: boolean;
    multiRegional?: boolean;
    nearline?: boolean;
    regional?: boolean;
    requesterPays?: boolean;
    retentionPolicy?: object;
    standard?: boolean;
    userProject?: string;
    location?: string;
    versioning?: Versioning;
}
```


sofisl (approved):

sofisl:
Hi @kemenesbalazs, thanks for your contribution! Before we can merge, can you sign the CLA? (see comment above). Thanks!

stephenplusplus:
I think this is a TypeScript issue. We should definitely be supporting `storageClass`, it just looks like the annotation doesn't show that.

kemenesbalazs:
@googlebot I signed it!

stephenplusplus:
@kemenesbalazs Thank you for pointing this out. The real issue was that the TypeScript interface was not allowing `storageClass` to be user-defined. I've added that to the interface in a new PR: #1299.

PR #1282, opened 2020-09-01
fix: enable retries for simpleUpload_

### LINKED ISSUE #1233: Error 502 (Server Error)!!1
#### Environment details

  - OS: Debian
  - Node.js version: 14
  - `@google-cloud/storage` version: 5.1.1

#### Steps to reproduce

```js
await googleStorageBucket.upload(transcodedFilePath, {
  destination: 'images/' + uid,
  gzip: false,
  metadata: {
    contentType: 'image/webp',
    storageClass: 'STANDARD',
  },
  onUploadProgress: createFileUploadProgressEventHandler(
    'transcoded image',
    uid,
    transcodedFileStat.size,
  ),
  predefinedAcl: 'bucketOwnerFullControl',
  resumable: false,
  validation: 'crc32c',
});
```

Occasionally fails with:

```yaml
statusCode:     502
statusMessage:  Bad Gateway
headers:
  User-Agent:        gcloud-node-storage/5.1.1
  x-goog-api-client: gl-node/14.5.0 gccl/5.1.1
  Authorization:     Bearer [..]
  Content-Type:      multipart/related; boundary=925dbc29-dd20-4554-b36b-89ebb41850c2
href:    https://storage.googleapis.com/upload/storage/v1/b/contrawork/o?uploadType=multipart&name=images%2F0c902ad8-403f-6000-9758-fe7194509706&predefinedAcl=bucketOwnerFullControl
body:
  """
    <!DOCTYPE html>
    <html lang=en>
      <meta charset=utf-8>
      <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
      <title>Error 502 (Server Error)!!1</title>
      <style>
        *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
      </style>
      <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
      <p><b>502.</b> <ins>That’s an error.</ins>
      <p>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.  <ins>That’s all we know.</ins>

  """

```

There appears to be no built in logic to retry these errors.

I worked around this issue by adding a generic retry wrapper above the `googleStorageBucket.upload` function. However, (1) I am still getting many of these errors logged (2) this [retry logic] looks like something that should be handled by `@google-cloud/storage`

### LINKED ISSUE #1281: Retry Strategy should be enabled by default for multipart uploads
Library issue, that does not enable retries for multipart uploads and requires enabling retries automatically using `autoRetry: true`. Resumable uploads does support retries by default so there's an inconsistency here. I don't see any issue with enabling autoRetry by default for multipart uploads.

cc: @crwilcox @dmahugh

Internal bug: 166293186

The workaround is to enable retries doing the following:

```javascript
const {Storage} = require('@google-cloud/storage');

// Your Google Cloud Platform project ID
const projectId = 'project-id';

// Creates a client
const storage = new Storage({
  projectId: projectId,
});

// The name for the new bucket
const contents = 'This is the contents of the file.';

storage
  .bucket(bucketName)
  .file('test')
  .save(contents, {
      autoRetry: true, // <<-- required to enable retries for multipart.
      resumable: false, // means, use multipart upload instead of resumable upload.
      }, function(err) {
    if (!err) {
      // File written successfully.
      console.log("File was written!");
    }
  });
```

#### Environment details

  - Node.js version: 12.7.0
  - `@google-cloud/storage` version: 5.3.0

frankyn:

Rationale for setting value is because default value is set to [0](https://github.com/googleapis/nodejs-common/blob/dbaad170a2e3a6785568523086f15d88fb34eaca/src/util.ts#L465).

Fixes #1281 🦕
Fixes #1233 

Repro steps:

## Setup testbench

1. Clone https://github.com/frankyn/google-cloud-cpp.git repository.
1. Checkout branch nodejs-multipart
1. Change Dir to google/cloud/storage/testbench
1. Follow set up in README.

## Run test script:

```javascript
const {Storage} = require('@google-cloud/storage');

// Your Google Cloud Platform project ID
const projectId = 'spec-test-ruby-samples';

// Creates a client
const apiEndpoint = 'http://localhost:8080';
const storage = new Storage({
  projectId: projectId,
  apiEndpoint,
});

// The name for the new bucket
const bucketName = 'anima-frank';
const contents = 'This is the contents of the file.';

storage.bucket(bucketName).create(function(err, bucket, apiResponse) {
  if (!err) {
    // The bucket was created successfully.
    console.log("Bucket created.");
    // Creates the new bucket
    storage
    .bucket(bucketName)
    .file('test')
    .save(contents, {
      resumable: false,
      },
      function(err) {
    if (!err) {
      // File written successfully.
      console.log("File was written!");
    }
    console.log(JSON.stringify(err));
    });
  }
});
```

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -3582,6 +3582,7 @@ class File extends ServiceObject<File> {
     const uri = `${apiEndpoint}/upload/storage/v1/b/${bucketName}/o`;
 
     const reqOpts: DecorateRequestOptions = {
+      maxRetries: 3,
       qs: {
         name: this.name,
       },
--- a/test/file.ts
+++ b/test/file.ts
@@ -4188,6 +4188,7 @@ describe('File', () => {
       makeWritableStreamOverride = (stream: {}, options_: any) => {
         assert.strictEqual(options_.metadata, options.metadata);
         assert.deepStrictEqual(options_.request, {
+          maxRetries: 3,
           qs: {
             name: file.name,
             predefinedAcl: options.predefinedAcl,


crwilcox (approved):

crwilcox:
LGTM and obeys the interface. And seeing as you took it through the paces with the testbench 💯

frankyn:
It retries, but it doesn't succeed on a following upload. May be what you were mentioning to me yesterday @stephenplusplus

stephenplusplus:
I don't think we can do this without any repercussions, as we set `maxRetries` to 0 by default due to issues with streaming uploads: https://github.com/googleapis/nodejs-common/pull/268

I think the only way we could allow retries for writable streams is when we already have all of the data to be written in the memory of the application. That is exactly the case for `file.save()`, but not for `file.createWriteStream({ resumable: false })`. The difference, in code:

```js
// This would be retryable
const blob = fs.readFileSync('./my-image.jpg')
await file.save(blob)

// This would not be retryable
file.createReadStream('./my-image.jpg')
  .pipe(file.createWriteStream({resumable: false}))
```

The problem is that a readable stream, by design, flushes its data from memory before it asks for newer data. So, once each chunk is passed to the final stream, the writable stream (`file.createWriteStream()`), it's gone and we can't get it back. So if the user has an issue half way through an upload, we cannot retry. But, `file.save()` is unique in that stores all of the data in memory. So, if we have to retry at any point, we can just fire the request again.

Now, here's where I lose my grip on understanding what's going on... a multipart upload is a single POST request, so even if the user is sending a large stream of data through `file.createWriteStream({resumable: false})`, all of the data *should* be collected into one place before the final request is sent. And that would make it retryable. But, we don't do that collecting ourselves, and I'm not even sure how it works-- this is the crucial line: https://github.com/googleapis/nodejs-common/blob/981a17650007d5c2a8b30051909baf2cc795d05d/src/util.ts#L478. We embed a stream as a part of the request multipart. It used to be the `request` library that knew how to handle that, and I'm not even sure without looking deeper what the new request library we're using is, and what it is doing with that stream.

The issues that were addressed in https://github.com/googleapis/nodejs-common/pull/268 could be gone, if that was back when we used `request`. Or, we could resurface them with this change. For now, I'd like to take a bit more time to reproduce and investigate before knowing if this solution will be side-effect free. Sorry for the delay!

stephenplusplus:
I haven't dug in yet, but for a quick solution, the following code should work:

```js
const saveWithRetries = (file, data, options, callback) => {
  if (typeof options === 'function') {
    callback = options;
    options = {};
  }

  const apiEndpoint = file.storage.apiEndpoint;
  const bucketName = file.bucket.name;
  const metadata = options.metadata || {};

  let reqOpts = {
    method: 'POST',
    uri: `${apiEndpoint}/upload/storage/v1/b/${bucketName}/o`,
    qs: {
      name: file.name,
      uploadType: 'multipart',
    },
    multipart: [
      {
        'Content-Type': 'application/json',
        body: JSON.stringify(metadata),
      },
      {
        'Content-Type': metadata.contentType || 'application/octet-stream',
        body: data,
      },
    ],
  };

  if (this.generation !== undefined) {
    reqOpts.qs.ifGenerationMatch = file.generation;
  }
  if (this.kmsKeyName !== undefined) {
    reqOpts.qs.kmsKeyName = file.kmsKeyName;
  }
  if (typeof options.timeout === 'number') {
    reqOpts.timeout = options.timeout;
  }
  if (options.userProject || this.userProject) {
    reqOpts.qs.userProject = options.userProject || file.userProject;
  }
  if (options.predefinedAcl) {
    reqOpts.qs.predefinedAcl = options.predefinedAcl;
  } else if (options.private) {
    reqOpts.qs.predefinedAcl = 'private';
  } else if (options.public) {
    reqOpts.qs.predefinedAcl = 'publicRead';
  }

  file.request(reqOpts, callback);
};

const bucket = storage.bucket('my-bucket');
const file = bucket.file('my-file');

saveWithRetries(file, 'my new data', callback);
```

The function would have to be manually promisified by the user, if necessary.

PR #1299, opened 2020-09-21
fix(types): add `CreateBucketRequest.storageClass`

stephenplusplus:
Closes #1248

The `CreateBucketRequest` interface was not accepting the `storageClass` string.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -83,6 +83,7 @@ export interface CreateBucketRequest {
   requesterPays?: boolean;
   retentionPolicy?: object;
   standard?: boolean;
+  storageClass?: string;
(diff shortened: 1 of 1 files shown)


laljikanjareeya (commented):
@stephenplusplus do we need this as all the classes available on the create bucket request
You can create bucket with storage class as below:

```ts
await storage.createBucket(bucketName, {
    [storageClassName]: true,
});
```

stephenplusplus:
Yes, that is our decoration to make things easier for the user, but we should always handle the REST-defined input as well.

laljikanjareeya (approved):

PR #1300, opened 2020-09-21
test: fix flaky test should get metadata for an HMAC key  

### LINKED ISSUE #1287: storage HMAC keys: should get metadata for an HMAC key failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: fd82e2880ffa274072810ac9a472576786238733
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/5caf977e-c4b2-4618-a6cb-6d10fa1e8a89), [Sponge](http://sponge2/5caf977e-c4b2-4618-a6cb-6d10fa1e8a89)
status: failed
<details><summary>Test output</summary><br><pre>Access ID not found in project.
Error: Access ID not found in project.
    at new ApiError (node_modules/@google-cloud/common/build/src/util.js:59:15)
    at Util.parseHttpRespBody (node_modules/@google-cloud/common/build/src/util.js:194:38)
    at Util.handleResp (node_modules/@google-cloud/common/build/src/util.js:135:117)
    at /tmpfs/src/github/nodejs-storage/node_modules/@google-cloud/common/build/src/util.js:434:22
    at onResponse (node_modules/retry-request/index.js:214:7)
    at /tmpfs/src/github/nodejs-storage/node_modules/teeny-request/build/src/index.js:219:13
        -> /tmpfs/src/github/nodejs-storage/node_modules/teeny-request/src/index.ts:325:11
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)</pre></details>

laljikanjareeya:
`deleteHmacKeys` is deleting all the keys available in project. While multiple build runs concurrently on CI there is a possibility that one is trying to get metadata at the same time another build deleted all the keys from project.

Fixes #1287

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2954,6 +2954,9 @@ describe('storage', () => {
 
     before(async () => {
       await deleteStaleHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
+      if (SECOND_SERVICE_ACCOUNT) {
+        await deleteStaleHmacKeys(SECOND_SERVICE_ACCOUNT, HMAC_PROJECT!);
+      }
     });
 
     it('should create an HMAC key for a service account', async () => {
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2952,14 +2952,6 @@ describe('storage', () => {
 
     let accessId: string;
 
-    before(async () => {
-      await deleteHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
-    });
-
-    after(async () => {
-      await deleteHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
-    });
-
     it('should create an HMAC key for a service account', async () => {
       const [hmacKey, secret] = await storage.createHmacKey(SERVICE_ACCOUNT, {
         projectId: HMAC_PROJECT,
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2950,7 +2950,12 @@ describe('storage', () => {
     const SECOND_SERVICE_ACCOUNT =
       process.env.HMAC_KEY_TEST_SECOND_SERVICE_ACCOUNT;
 
+    const acessIdsToCleanup: string[] = [];
     let accessId: string;
+
+    after(async () => {
+      await deleteHmacKeys(acessIdsToCleanup);
+    });
 
     it('should create an HMAC key for a service account', async () => {
       const [hmacKey, secret] = await storage.createHmacKey(SERVICE_ACCOUNT, {
@@ -2959,6 +2964,8 @@ describe('storage', () => {
       // We should always get a 40 character secret, which is valid base64.
       assert.strictEqual(secret.length, 40);
       accessId = hmacKey.id!;
+      // key should be deleted incase delete test fails.
+      acessIdsToCleanup.push(accessId);
       const metadata = hmacKey.metadata!;
       assert.strictEqual(metadata.accessId, accessId);
       assert.strictEqual(metadata.state, 'ACTIVE');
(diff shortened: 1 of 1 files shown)


system-test/storage.ts
```diff
@@ -2967,6 +2964,8 @@ describe('storage', () => {
       // We should always get a 40 character secret, which is valid base64.
       assert.strictEqual(secret.length, 40);
       accessId = hmacKey.id!;
+      // key should be deleted incase delete test fails.
```
AVaksman, line 2967:
```suggestion
      // key should be deleted in case delete test fails.
```

stephenplusplus:
An issue with this approach is that if this test run fails for some reason, or the delete operation fails, those resources will continue to exist and never be picked up by another test run for removal. The usual approach we do is to get the resources that are stale (generally, were created one hour ago or more), and delete those. That way, we don't worry about deleting the ones we just created-- they will be caught and deleted in a future test run.

Is a structure like that possible for HMAC keys?

AVaksman:
Maybe can try getting a list of all keys and then filter by both `hmacKey.metadata!.state` and `hmacKey.metadata!.timeCreated`

laljikanjareeya:
@stephenplusplus @AVaksman implemented delete stale HMAC key logic, it will delete all the keys that has been created before one hour.

stephenplusplus (approved):
Thank you!

system-test/storage.ts
```diff
@@ -2953,11 +2953,7 @@ describe('storage', () => {
     let accessId: string;
 
     before(async () => {
-      await deleteHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
-    });
-
-    after(async () => {
-      await deleteHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
+      await deleteStaleHmacKeys(SERVICE_ACCOUNT, HMAC_PROJECT!);
```
AVaksman, line 2956: Should also add here
```ts
await deleteStaleHmacKeys(SECOND_SERVICE_ACCOUNT, HMAC_PROJECT!);
```
in case a previous test crashed?

system-test/storage.ts
```diff
@@ -3022,11 +3018,10 @@ describe('storage', () => {
     });
 
     describe('second service account', () => {
-      before(async function () {
+      before(function () {
```
AVaksman, line 3021: We do not seem to clean up Hmac keys from the `SECOND_SERVICE_ACCOUNT`?

Should add the logic?
1. Capture `accessId` in the `'should create key for a second service account'` test
2. Delete the Hmac key by accessId in the `after`

WDYT?

laljikanjareeya: Done!!

AVaksman (approved):

system-test/storage.ts
```diff
@@ -3022,17 +3021,25 @@ describe('storage', () => {
...
+
+      after(async () => {
+        const hmacKey = storage.hmacKey(accessId, {projectId: HMAC_PROJECT});
+        await hmacKey.setMetadata({state: 'INACTIVE'});
+        await hmacKey.delete();
       });
 
       it('should create key for a second service account', async () => {
-        await storage.createHmacKey(SECOND_SERVICE_ACCOUNT!, {
+        const [hmacKey] = await storage.createHmacKey(SECOND_SERVICE_ACCOUNT!, {
```
stephenplusplus, line 3039: Does this have the same issue where these could be un-deleted if there's a test failure / system crash / operation fails? Do we need to loop over both service accounts in the global before hook?

stephenplusplus (changes requested):
Sorry...

system-test/storage.ts
laljikanjareeya: `Service Account` key will delete in delete test, while `Second Service Account` key will delete in `after` hook.
In case if any of above fails the `deleteStaleHmacKeys` will delete it in next cycle after one hour.

stephenplusplus: Sorry, I missed that line of code 🤦🏻 Thanks!

stephenplusplus (approved):

PR #1301, opened 2020-09-22
docs: storage class docs in create bucket

laljikanjareeya:
Recently added `storageClass` in `CreateBucketRequest` in [PR](https://github.com/googleapis/nodejs-storage/pull/1299).


src/storage.ts
```diff
@@ -529,6 +529,8 @@ export class Storage extends Service {
    *     Force the use of the User Project metadata field to assign operational
    *     costs when an operation is made on a Bucket and its objects.
    * @property {boolean} [standard=true] Specify the storage class as Standard.
+   * @property {string} [storageClass] The new storage class. (`standard`,
```
stephenplusplus, line 532: I believe we need a complete list from here: https://github.com/googleapis/nodejs-storage/blob/a1135afdce9daf44eea6d7ff8c82d541e413a661/src/storage.ts#L635-L641

stephenplusplus (changes requested):

src/storage.ts
laljikanjareeya: Done!

stephenplusplus (approved):

PR #1304, opened 2020-09-27
refactor(deps): replace concat-stream with get-stream

TrySound:
Ref https://packagephobia.com/result?p=concat-stream https://packagephobia.com/result?p=get-stream

Concat-stream is a big package with not relevant in node 10 dependencies.
In this diff I replaced it with dependency-free get-stream with promise
based api.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -24,7 +24,7 @@ import {
 import {promisifyAll} from '@google-cloud/promisify';
 
 import compressible = require('compressible');
-import concat = require('concat-stream');
+import getStream = require('get-stream');
 import * as crypto from 'crypto';
 import * as dateFormat from 'date-and-time';
 import * as extend from 'extend';
@@ -1267,12 +1267,10 @@ class File extends ServiceObject<File> {
       ) => {
         if (err) {
           // Get error message from the body.
-          rawResponseStream.pipe(
-            concat(body => {
-              err.message = body.toString();
-              throughStream.destroy(err);
-            })
-          );
+          getStream(rawResponseStream).then(body => {
+            err.message = body;
+            throughStream.destroy(err);
+          });
 
           return;
         }
@@ -1992,7 +1990,12 @@ class File extends ServiceObject<File> {
         .on('error', callback)
         .on('finish', callback);
     } else {
-      fileStream.on('error', callback).pipe(concat(callback.bind(null, null)));
+      getStream
+        .buffer(fileStream)
+        .then(
+          callback.bind(null, null) as (contents: Buffer) => void,
+          callback as (error: RequestError) => void
+        );
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -1267,12 +1267,10 @@ class File extends ServiceObject<File> {
       ) => {
         if (err) {
           // Get error message from the body.
-          rawResponseStream.pipe(
```
JustinBeckwith, line 1270: In the general sense ... are we making this harder than it needs to be?

```js
const results = [];
rawResponseStream
  .on('data', results.push)
  .on('end', () => {
    err.message = results.concat();
    throughStream.destroy(err);
  });
});
```

Adding @stephenplusplus, the stream whisperer, to chime in :)

TrySound: Can this be a problem with utf-8 when ligatures will be in separate chunk from characters?

stephenplusplus: @TrySound I believe get-stream is doing essentially the example @JustinBeckwith provided, but let me know if I missed it. I think the point is, we could probably duplicate the logic in both of these modules relatively easily and avoid any extra dependencies. In either case, I would still lean towards using get-stream just for simplicity. Especially since it's already a huge win, size-wise, over concat-stream.

src/file.ts
```diff
@@ -1992,7 +1990,12 @@ class File extends ServiceObject<File> {
         .on('error', callback)
         .on('finish', callback);
     } else {
-      fileStream.on('error', callback).pipe(concat(callback.bind(null, null)));
+      getStream
+        .buffer(fileStream)
+        .then(
```
stephenplusplus, line 1995: This might be easier to read if it was something like:

```js
getStream
  .buffer(fileStream)
  .then(contents => callback(null, contents))
  .catch(callback)
```

I know you copied our existing `callback.bind(null, null)` (thank you for the attention to detail), but I think arrow functions are new since we wrote this, so we can "upgrade" now, if you're willing :)

stephenplusplus:
@TrySound thank you for noticing this and helping out!

src/file.ts
TrySound: Not really easier. Refinement will be invalidated and callback will become nullable again.

stephenplusplus: True, I think TS makes both options much less readable :) I think "bind" was a well-known, useful solution in the old days, but I don't know how many generations of developers are left who will see that and know what's really happening or why. And with arrow functions being an option, I don't think we have a good explanation anymore. Thanks for making the change.

TrySound: I used optional chaining to simplify this. Looks like everything still works.

stephenplusplus (approved):

stephenplusplus:
Thanks!

PR #1311, opened 2020-10-13
feat: add File#rename

### LINKED ISSUE #1308: Their should be a function to rename file
Their should be a method to rename the existing file rather than copying it with different name and the deleting the previous version. Like we have the rename functionality in the developer console.

AVaksman:

Fixes #1308

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -249,6 +249,10 @@ export interface MoveOptions {
   userProject?: string;
 }
 
+export type RenameOptions = MoveOptions;
+export type RenameResponse = MoveResponse;
+export type RenameCallback = MoveCallback;
+
 export type RotateEncryptionKeyOptions = string | Buffer | EncryptionKeyOptions;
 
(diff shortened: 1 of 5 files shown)


src/file.ts
```diff
@@ -3233,6 +3237,103 @@ class File extends ServiceObject<File> {
...
+   */
+  /**
+   * @typedef {object} RenameOptions Configuration options for File#move(). See an
+   *     [Object
+   * resource](https://cloud.google.com/storage/docs/json_api/v1/objects#resource).
+   * @param {string} [userProject] The ID of the project which will be
+   *     billed for the request.
+   */
+  /**
+   * Rename this file
```
stephenplusplus, line 3269: We should explain probably what "move()" does, that this is not an atomic operation, and also that this is an alias for "move()".

stephenplusplus (approved):

JustinBeckwith (approved):

PR #1317, opened 2020-10-19
fix(docs): explain manual pagination and add usage sample

### LINKED ISSUE #1277: Cloud Storage getFiles() documentation does not explain how to use pageToken
I'm trying to use Google Cloud Storage [getFiles](https://googleapis.dev/nodejs/storage/latest/Bucket.html#getFiles) to delete a large collection of files. I assume I should get the files in batches and delete them that way; I see a `pageToken` option that looks like it should be used for that. But nowhere in the doc does it say where that `pageToken` is returned to me, so I can pass it to the next iteration. The only lines for `pageToken` in the [Global](https://googleapis.dev/nodejs/storage/latest/global.html#GetFilesOptions) docs are for places to pass it in, not calls that return it.

I've seen some examples that hint that if you pass a callback rather than expecting a Promise, the callback may get the pageToken as an extra arg; but the [doc for getFilesCallback](https://googleapis.dev/nodejs/storage/latest/global.html#GetFilesCallback) doesn't say anything about that and I don't want to depend on undocumented behavior. I've seen some old closed issues on this github but nothing that explains the proper method.

What's the proper method for getting pagination to work for `getFiles`?

AVaksman:

Fixes #1277


samples/listFilesPaginate.js
```diff
@@ -0,0 +1,54 @@
...
+
+      // nextQuery will have all the custom properties originally set in
+      // getFilesOptions together with nextPageToken already set.
+      getFilesOptions = nextQuery;
+    } while (getFilesOptions);
+
+    console.log('Files:');
+    allFiles.forEach(file => {
+      console.log(file.name);
+    });
```
stephenplusplus, line 48: This is always tricky to explain. I think if a user is opting-out of auto-pagination, we don't need to show them how to recreate it. Just explaining the basic set of tools should suffice, maybe like:

```js
const [files, queryForPage2] = await bucket.getFiles({autoPaginate: false});

// Get the next set of results
if (queryForPage2 !== null) {
  const [files, queryForPage3] = await bucket.getFiles(queryForPage2);
}
```

AVaksman: I was trying to make a sample that will run a few cycles.

Available options

1. `autoPaginate: false` - will return 1 page of up to 1000 results, in a simple sample there won't be a nextQuery
2. `maxResults: <more than 1 page worth>` - will autoPaginate on default up to set value. Not really relevant for a simple sample
3. `maxResults: <more than 1 page worth>` and `autoPaginate: false` - will return only 1 page worth of results
4. `maxResults: <less then 1 page>` (autoPaginate value irrelevant) - will return up to set value

Option 4 seemed possible in a simple sample.

WDYT?

stephenplusplus: Do you mean the motivation is to make the sample testable? I don't think that's as important as explaining the concept, since we have extensive unit and system test coverage in the respective places (paginator / storage libs).

AVaksman: Done

samples/listFilesPaginate.js
```diff
@@ -0,0 +1,51 @@
...
+    // Get the next set of results
+    if (queryForPage2 !== null) {
+      const [files, queryForPage3] = await bucket.getFiles(queryForPage2);
+      files.forEach(file => {
+        console.log(file.name);
+      });
+
+      if (queryForPage3 !== null) {
+        // Request the nest set of results...
+      }
```
stephenplusplus, line 44:
```suggestion
      // If necessary, continue cursoring using "queryForPage3"
```

samples/listFilesPaginate.js
```diff
@@ -0,0 +1,51 @@
...
+  const storage = new Storage();
+
+  async function listFilesPaginated() {
+    const bucket = storage.bucket(bucketName);
+    const [files, queryForPage2] = await bucket.getFiles({autoPaginate: false});
+    console.log('Files:');
+    files.forEach(file => {
+      console.log(file.name);
+    });
+    // Get the next set of results
```
stephenplusplus, line 35:
```suggestion

    // Page through the next set of results using "queryForPage2"
```

samples/listFilesPaginate.js
```diff
@@ -0,0 +1,51 @@
...
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage();
+
+  async function listFilesPaginated() {
+    const bucket = storage.bucket(bucketName);
+    const [files, queryForPage2] = await bucket.getFiles({autoPaginate: false});
+    console.log('Files:');
```
stephenplusplus, line 31:
```suggestion

    console.log('Files:');
```

stephenplusplus (changes requested):

samples/listFilesPaginate.js
AVaksman: Will try, linter may complain though.

AVaksman: > error  'queryForPage3' is assigned a value but never used  no-unused-vars

samples/listFilesPaginate.js
stephenplusplus: I was just suggesting we add a new line before the `console.log`, which it looks like that sample does as well. Sorry for the confusion.

samples/listFilesPaginate.js
stephenplusplus: I think I've seen that rule ignored for the sake of samples. I forget how specific the rule was-- is it possible to ignore that rule break with a "ignore" line at the top of the file / beginning of the function (outside the printable area)?

AVaksman: Done

samples/listFilesPaginate.js
AVaksman: Added a line

stephenplusplus: Sorry, this new line should go above the `console.log`:

```js
const [files, queryForPage2] = await bucket.getFiles({autoPaginate: false});

console.log('Files:');
files.forEach(file => {
  console.log(file.name);
});
```

AVaksman: ✔️ 
Thanks!

stephenplusplus (approved):

shaffeeullah:
@AVaksman can this be merged?

AVaksman:
> @AVaksman can this be merged?
Yes, it is ready to be merged.
Thank you

PR #1318, opened 2020-10-19
fix: self-upload files for Unicode system tests

shaffeeullah:
The two unicode validation tests depended on a bucket which has been deleted. This PR changes those tests so that they create their own data.

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -890,10 +890,8 @@ describe('storage', () => {
       const name = 'Caf\u00e9';
       const expectedContents = 'Normalization Form C';
 
-      //Data setup
-      await bucket.file(name).save(expectedContents);
-
       const file = bucket.file(name);
+      await file.save(expectedContents);
 
       return file
         .get()
@@ -913,10 +911,8 @@ describe('storage', () => {
       const name = 'Cafe\u0301';
       const expectedContents = 'Normalization Form D';
 
-      //Data setup
-      await bucket.file(name).save(expectedContents);
-
       const file = bucket.file(name);
+      await file.save(expectedContents);
 
       return file
         .get()
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -880,17 +880,20 @@ describe('storage', () => {
 
     let bucket: Bucket;
 
-    before(() => {
-      bucket = storage.bucket('storage-library-test-bucket');
+    before(async () => {
+      [bucket] = await storage.createBucket(generateName());
     });
 
     // Normalization form C: a single character for e-acute;
     // URL should end with Cafe%CC%81
-    it('should not perform normalization form C', () => {
+    it('should not perform normalization form C', async () => {
       const name = 'Caf\u00e9';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form C';
+
+      //Data setup
+      await bucket.file(name).save(expectedContents);
+
+      const file = bucket.file(name);
 
       return file
         .get()
@@ -906,11 +909,14 @@ describe('storage', () => {
 
     // Normalization form D: an ASCII character followed by U+0301 combining
     // character; URL should end with Caf%C3%A9
-    it('should not perform normalization form D', () => {
+    it('should not perform normalization form D', async () => {
       const name = 'Cafe\u0301';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form D';
+
+      //Data setup
+      await bucket.file(name).save(expectedContents);
+
+      const file = bucket.file(name);
 
       return file
         .get()


system-test/storage.ts
```diff
@@ -880,18 +880,21 @@ describe('storage', () => {
...
       const name = 'Caf\u00e9';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form C';
 
+      //Data setup
+      await bucket.file(name).save(expectedContents);
+
+      const file = bucket.file(name);
+
```
stephenplusplus, line 897:
```suggestion
      const file = bucket.file(name);
      await file.save(expectedContents);

```

system-test/storage.ts
```diff
@@ -906,12 +909,15 @@ describe('storage', () => {
...
       const name = 'Cafe\u0301';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form D';
 
+      //Data setup
+      await bucket.file(name).save(expectedContents);
+
+      const file = bucket.file(name);
+
```
stephenplusplus, line 920:
```suggestion
      const file = bucket.file(name);
      await file.save(expectedContents);

```

stephenplusplus (changes requested):

stephenplusplus:
I think this might be a "chore" instead of a "fix".

system-test/storage.ts
```diff
@@ -880,18 +880,19 @@ describe('storage', () => {
...
+    it('should not perform normalization form C', async () => {
       const name = 'Caf\u00e9';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form C';
 
+      const file = bucket.file(name);
+      await file.save(expectedContents);
+
       return file
```
AVaksman, line 896: Should this logic be switched to use `await` as well?

system-test/storage.ts
```diff
@@ -906,12 +907,13 @@ describe('storage', () => {
...
+    it('should not perform normalization form D', async () => {
       const name = 'Cafe\u0301';
-      const file = bucket.file(name);
-
       const expectedContents = 'Normalization Form D';
 
+      const file = bucket.file(name);
+      await file.save(expectedContents);
+
       return file
```
AVaksman, line 917: Should this logic be switched to use `await` as well?

system-test/storage.ts
```diff
@@ -880,18 +880,19 @@ describe('storage', () => {
 
     let bucket: Bucket;
 
-    before(() => {
-      bucket = storage.bucket('storage-library-test-bucket');
+    before(async () => {
+      [bucket] = await storage.createBucket(generateName());
```
JustinBeckwith, line 884: This looks like it's creating buckets, but possibly not cleaning them up after.  Do we have code that will come back around and delete it, or is it possible we're leaking buckets (which is a funny turn of phrase now that I think about it)

system-test/storage.ts
JustinBeckwith: Yes :) It's a wonder this test wasn't flaky before!

system-test/storage.ts
shaffeeullah: Good catch! I will make that fix.

system-test/storage.ts
shaffeeullah: Which logic are you referring to here? `bucket.file(` or the `return` statement?

system-test/storage.ts
stephenplusplus: Yep, the after hook deletes all buckets that were created (as long as the "generateName()" function was used for a name).

system-test/storage.ts
```diff
@@ -880,18 +880,19 @@ describe('storage', () => {
 
     let bucket: Bucket;
 
-    before(() => {
-      bucket = storage.bucket('storage-library-test-bucket');
+    before(async () => {
```
shaffeeullah, line 883: @JustinBeckwith / @AVaksman best practice question: should this `before` clause be returning something? or is it okay the way it is?

system-test/storage.ts
shaffeeullah: @JustinBeckwith best practice question: should the `after()` clause that deletes the bucket wait for the bucket to actually be deleted? (`return await bucket.delete();` instead of `return bucket.delete()`)

system-test/storage.ts
AVaksman: IMHO it is correct as it is right now.

system-test/storage.ts
JustinBeckwith: So this is important!  I always `await`, and then `return`.  I personally prefer to keep those on separate lines for complete clarity.  This is a good read on the subject:
https://jakearchibald.com/2017/await-vs-return-vs-return-await/

system-test/storage.ts
AVaksman: `file.get()` and `receivedFile.downLoad()` functions

system-test/storage.ts
JustinBeckwith: Nah this is fine

system-test/storage.ts
AVaksman: I believe the clean up [after hook](https://github.com/googleapis/nodejs-storage/blob/ea35986a08c6c049865fbdab937c9b2b32cf15af/system-test/storage.ts#L161-L163) logic properly [awaits](https://github.com/googleapis/nodejs-storage/blob/ea35986a08c6c049865fbdab937c9b2b32cf15af/system-test/storage.ts#L3699) for all the buckets to be deleted 👍 .

frankyn (approved):

frankyn:
Pending approval from Node.js Experts

stephenplusplus (approved):

system-test/storage.ts
shaffeeullah: What is the difference between `.then()` and `await`? From what I understand, `then` will wait for the prior code to return before executing. Is this understanding incorrect?

AVaksman: `await X()` suspends execution until `X`'s async call completes.
`promise.then(X)` adds the `X` to the callback chain and continues the current execution.
```ts
      function greeting() {
        return new Promise(resolve => {
          resolve(console.log('Hi'));
        });
      }
      function hru() {
        console.log('How are you?');
      }
      function cu() {
        console.log('See you later');
      }
```
```ts
     greeting().then(() => hru());
     cu();
```
will print
`Hi`
`See you later`
`How are you?`
vs
```ts
      await greeting();
      hru();
      cu();
```
will print
`Hi`
`How are you?`
`See you later`

shaffeeullah: This is a super helpful explanation, thank you so much! I really appreciate it. In the case of our return statement here, can you please clarify your concern? In your first example above, `Hi` will always print before `How are you?`, correct? And so in the case of our return statement, it will wait for `file.get()` before the next step and `receivedFile.download()` before the last step.

AVaksman: The concern is that it may continue to the the next test after firing `file.get()` before getting to `.then()`'s and not getting to the `assert`s.

AVaksman (approved):

JustinBeckwith (approved):

Issue #697, opened 2019-05-06
Get long lived download URLs in addition to SignedURLs


theochampion:
Hi there,

I noticed that there is currently no way to get a long-lived download URL from an uploaded resource using the node admin package which causes problems when storing generated URLs in external DB as they expire after some time.

Would it be possible to generate long-lived URLs just like the `getDownloadUrl()` function from the client SDKs
(https://firebase.google.com/docs/reference/android/com/google/firebase/storage/StorageReference.html#getDownloadUrl() ). 

Cheers!

avin3sh:
From [GCP Storage Docs](https://cloud.google.com/storage/docs/access-public-data),



>     Get the name of the bucket containing the public data.
> 
>     Use the following URI to access an object in the bucket:
> 
>     http://storage.googleapis.com/[BUCKET_NAME]/[OBJECT_NAME]
> 
> For example, the Google public bucket gcp-public-data-landsat contains the Landsat public dataset. You can link to the publicly shared object LC08/PRE/063/046/LC80630462016136LGN00/LC80630462016136LGN00_B11.TIF with the link:
> 
> http://storage.googleapis.com/gcp-public-data-landsat/LC08/PRE/063/046/LC80630462016136LGN00/LC80630462016136LGN00_B11.TIF
> 
> This link does not require authentication in order to use. It is suitable, for example, as a link in a web page, or for downloading with a command-line tool such as cURL.

theochampion:
@avin3sh thanks for the reply, this is the kind of predictable URLs i'd be looking for, however, this only applies to public data.

What I need is a predictable URL that still expects authentication (via firebase auth in my case) to access the data

tdkehoe:
I write a book-length :-) summary of the [problems associated with the three types of download URLs](https://stackoverflow.com/questions/42956250/get-download-url-from-file-uploaded-with-cloud-functions-for-firebase/56010225?noredirect=1#comment98914254_56010225) for Google Cloud Storage, on Stack Overflow.

AVaksman:
The maximum duration that the signed url should be valid for is mentioned [here](https://cloud.google.com/storage/docs/gsutil/commands/signurl#options)

AVaksman:
Currently 7 day limit is only true for V4 signatures and API will throw a `Max allowed expiration is seven days (604800 seconds).` error.
V2 signatures still support longer duration.

theochampion:
[...]

PR #1321, opened 2020-10-22
feat: Experimental getFirebaseDownloadUrl API

### LINKED ISSUE #244: getSignedURL() in Google Cloud Function produces link that works for several days, then returns "SignatureDoesNotMatch" error
#### Environment details

  - OS: Heroku dyno (Ubuntu 16.04)
  - Node.js version: 8.6
  - npm version: 6.0.0
  - `@google-cloud/storage` version: 1.6.0

#### Steps to reproduce

Repo of relevant code:
https://github.com/colinjstief/getSignedUrl-example

Lots of this code is copied/pasted straight from official example ([link here](https://github.com/firebase/functions-samples/tree/master/generate-thumbnail)). Code in my repo runs as a Google Cloud Function and produces a link that works for a few days (maybe 7?), and then seems to expire with "SignatureDoesNotMatch" 403 response. Full message reads:

"The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method."

URL seems valid, including this query parameter: '...&Expires=16730323200&Signature=...'

Both the initial request and the later request (after apparent expiration) are performed the same ways: either through an axios.get() or simply placing in browser bar. Both work at first, then both fail later.

Same issue seems to be reported here GoogleCloudPlatform/google-cloud-node#1976, here googleapis/nodejs-storage#144, and here https://github.com/firebase/functions-samples/issues/360

hiranya911:

Fixes #244 (using this old issue temporarily to kickoff the discussion) 🦕

### Background

Currently this library only supports generating signed URLs. Signed URLs have an expiry time, and are attached to a private key. They become invalid when the expiry time is up or when the private key gets cycled out. Firebase Storage (which operates on top of GCS) supports a notion of permanent download URLs. These URLs remain valid until the developer explicitly revokes them, and therefore are better for saving into a database or some other permanent record keeping system.

### Problem

There's currently a feature gap in the GCS SDK due to it only supporting signed URLs. Server-side Firebase developers have no easy way for generating Firebase download URLs using this library (or the Firebase Admin SDK that re-exports this library). This has resulted in several reported issues and questions on SO. 

The temporary nature of signed URLs often causes problems in environments like GCF where developers use application default credentials. The underlying service account automatically gets cycled out within a week or so, leaving all the signed URLs invalid. Firebase download URLs are a better fit for such use cases.

For more context:

* https://stackoverflow.com/questions/42956250/get-download-url-from-file-uploaded-with-cloud-functions-for-firebase
#244
* https://github.com/googleapis/nodejs-storage/issues/697
* https://github.com/firebase/extensions/issues/140#issuecomment-595195942

### Solution

Adding a new `getFirebaseDownloadUrl()` method to the `File` class that generates Firebase Storage download URLs (token based access control). 

Just wanted to know what the community thinks about this approach, and possible ways we can improve it.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -271,6 +271,15 @@ export enum ActionToHTTPMethod {
   resumable = 'POST',
 }
 
+export type GetFirebaseDownloadUrlResponse = GetSignedUrlResponse;
+export type GetFirebaseDownloadUrlCallback = GetSignedUrlCallback;
+
+interface FirebaseStorageMetadata {
+  metadata?: {
+    firebaseStorageDownloadTokens?: string;
+  };
+}
+
 /**
  * Custom error type for errors related to creating a resumable upload.
  *
@@ -2853,6 +2862,90 @@ class File extends ServiceObject<File> {
       .then(signedUrl => callback!(null, signedUrl), callback!);
   }
 
+  getFirebaseDownloadUrl(): Promise<GetFirebaseDownloadUrlResponse>;
+  getFirebaseDownloadUrl(callback: GetFirebaseDownloadUrlCallback): void;
+  /**
+   * @typedef {array} GetFirebaseDownloadUrlResponse
+   * @property {object} 0 The Firebase Storage download URL.
+   */
+  /**
+   * @callback GetFirebaseDownloadUrlCallback
+   * @param {?Error} err Request error, if any.
+   * @param {object} url The Firebase Storage download URL.
+   */
+  /**
+   * Get a Firebase download URL.
+   *
+   * Firebase download URLs use the `firebasestorage.googleapis.com` hostname, and have a static
+   * token attached to them. Unlike signed URLs, Firebase download URLs do not expire after a
+   * specific time period. Instead they remain valid until the underlying token is explicitly
+   * revoked. Developers can revoke these tokens from the Firebase Console or by directly modifying
+   * the object metadata.
+   *
+   * @param {GetFirebaseDownloadUrlCallback} [callback] Callback function.
+   * @returns {Promise<GetFirebaseDownloadUrlResponse>}
+   *
+   * @example
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const myBucket = storage.bucket('my-bucket');
+   *
+   * const file = myBucket.file('my-file');
+   *
+   * //-
+   * // Generate a URL that allows temporary access to download your file.
+   * //-
+   * const request = require('request');
+   *
+   * file.getFirebaseDownloadUrl(function(err, url) {
+   *   if (err) {
+   *     console.error(err);
+   *     return;
+   *   }
+   *
+   *   // The file is now available to read from this URL.
+   *   request(url, function(err, resp) {
+   *     // resp.statusCode = 200
+   *   });
+   * });
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * file.getFirebaseDownloadUrl().then(function(data) {
+   *   const url = data[0];
+   * });
+   */
+  getFirebaseDownloadUrl(
+    callback?: GetFirebaseDownloadUrlCallback
+  ): void | Promise<GetFirebaseDownloadUrlResponse> {
+    this.ensureDownloadToken()
+      .then(token => {
+        return `https://firebasestorage.googleapis.com/v0/b/${
+          this.bucket.name
+        }/o/${encodeURIComponent(this.name)}?alt=media&token=${token}`;
+      })
+      .then(url => callback!(null, url), callback!);
+  }
+
+  private ensureDownloadToken(): Promise<string> {
+    return this.getMetadata().then(response => {
+      const metadata = response[0] as FirebaseStorageMetadata;
+      if (metadata?.metadata?.firebaseStorageDownloadTokens) {
+        return metadata.metadata.firebaseStorageDownloadTokens;
+      }
+
+      const token = generateDownloadToken();
+      return this.setMetadata({
+        metadata: {
+          firebaseStorageDownloadTokens: token,
+        },
+      }).then(() => {
+        return token;
+      });
+    });
+  }
+
   isPublic(): Promise<IsPublicResponse>;
   isPublic(callback: IsPublicCallback): void;
   /**
@@ -3744,13 +3837,28 @@ class File extends ServiceObject<File> {
   }
 }
 
+function generateDownloadToken(length = 32): string {
+  const alphabet =
+    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  let text = '';
+  for (let i = 0; i < length; i++) {
+    text += alphabet.charAt(random(alphabet.length));
+  }
+
+  return text;
+}
+
+function random(max: number): number {
+  return Math.floor(Math.random() * max);
+}
+
 /*! Developer Documentation
  *
  * All async methods (except for streams) will return a Promise in the event
  * that a callback is omitted.
  */
 promisifyAll(File, {
-  exclude: ['request', 'setEncryptionKey'],
+  exclude: ['ensureDownloadToken', 'request', 'setEncryptionKey'],
 });
 
 /**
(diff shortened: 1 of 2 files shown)


frankyn:
@crwilcox I think offline discussions left off on not moving forward with specific Firebase features in the Storage library. Do you know if that has changed?

tonyjhuang:
Hiranya, could you explain what the intended usecase for this api is and what problem it is intended to fix?

src/file.ts
```diff
@@ -2853,6 +2862,90 @@ class File extends ServiceObject<File> {
...
+   * file.getFirebaseDownloadUrl().then(function(data) {
+   *   const url = data[0];
+   * });
+   */
+  getFirebaseDownloadUrl(
+    callback?: GetFirebaseDownloadUrlCallback
+  ): void | Promise<GetFirebaseDownloadUrlResponse> {
+    this.ensureDownloadToken()
+      .then(token => {
+        return `https://firebasestorage.googleapis.com/v0/b/${
```
tonyjhuang, line 2924: This shouldn't change your PR but FYI we're considering eventually building out an endpoint that users can call to disable download tokens on a per project or per bucket level.

src/file.ts
```diff
@@ -3744,13 +3837,28 @@ class File extends ServiceObject<File> {
   }
 }
 
+function generateDownloadToken(length = 32): string {
```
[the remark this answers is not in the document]
hiranya911: My gut feeling is that integrating with an external endpoint unnecessarily complicates this implementation. Is there any particular benefit in using the remote service for token generation?

hiranya911:
@tonyjhuang I've added more background info and context links to the issue description.

src/file.ts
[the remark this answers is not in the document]
tonyjhuang: My initial thought was that this implementation of storing download tokens GCS bucket metadata may change in the future but given the fact that our other client SDKs make a similar assumption, this should be fine.

src/file.ts
```diff
@@ -2853,6 +2862,90 @@ class File extends ServiceObject<File> {
...
+   * });
+   */
+  getFirebaseDownloadUrl(
+    callback?: GetFirebaseDownloadUrlCallback
+  ): void | Promise<GetFirebaseDownloadUrlResponse> {
+    this.ensureDownloadToken()
+      .then(token => {
+        return `https://firebasestorage.googleapis.com/v0/b/${
+          this.bucket.name
+        }/o/${encodeURIComponent(this.name)}?alt=media&token=${token}`;
```
avolkovi, line 2926: alt=media will return the file, not the metadata, if you want the metadata take that parameter out

avolkovi: oh j/k, you're just creating the url here, my bad

src/file.ts
```diff
@@ -2853,6 +2862,90 @@ class File extends ServiceObject<File> {
...
+      })
+      .then(url => callback!(null, url), callback!);
+  }
+
+  private ensureDownloadToken(): Promise<string> {
+    return this.getMetadata().then(response => {
+      const metadata = response[0] as FirebaseStorageMetadata;
+      if (metadata?.metadata?.firebaseStorageDownloadTokens) {
+        return metadata.metadata.firebaseStorageDownloadTokens;
```
avolkovi, line 2935: this is actually an array of strings, you'll want to parse and return the first one

src/file.ts
```diff
@@ -2853,6 +2862,90 @@ class File extends ServiceObject<File> {
...
+      }
+
+      const token = generateDownloadToken();
+      return this.setMetadata({
+        metadata: {
+          firebaseStorageDownloadTokens: token,
+        },
+      }).then(() => {
```
avolkovi, line 2943: why not await/async?

hiranya911: This file already heavily uses then/catch, so I was trying to be consistent. I can change this to async/await if needed.

src/file.ts
hiranya911: Doesn't seem to be the case for files uploaded from the Firebase Console. Here's the metadata for one such file:

```
metadata:
   { firebaseStorageDownloadTokens: 'da8c9c47-16c1-4667-9dac-e534a33c434e' }
```

Does this mean we should assume the type of this property is `string | string[]`?

avolkovi: `string | string[]` looks right to me, we initialize with just one token, but multiple tokens can exist (try adding one in console)

src/file.ts
avolkovi: just a nit, feel free to leave for a separate pr/never

src/file.ts
hiranya911: Done

tonyjhuang (approved):
LGTM thanks for the new feature!

src/file.ts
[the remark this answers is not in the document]
crwilcox: @tonyjhuang what are `our other client SDKs`? Where else is this code today?

crwilcox:
Is there a reason this code cannot be added to the Firebase Admin SDK instead of the GCS Library?

> Server-side Firebase developers have no easy way for generating Firebase download URLs using this library (or the Firebase Admin SDK that re-exports this library). This has resulted in several reported issues and questions on SO.

It seems that this is used by that subset of users. While I haven't looked up stats, but feel this might be 'bloat' in the GCS library, since we have many customers who wouldn't be using this added functionality.

Is there a technical reason to favor this approach?

hiranya911:
We (well, at least I) slightly prefer this option because Firebase Admin SDKs don't really have any GCS code. They are just a facade to initialize and use the official GCS clients like `@google-cloud/storage`. Given that there are issues like #697 reported on this repo, it seems like a good fit here.

crwilcox:
The code seems to (at a high level) construct a URI to `firebasestorage.googleapis.com`.  Will this URI work for all library users, or only ones who have used firebase with the project associated with the storage bucket?

hiranya911:
I think at the least the Firebase API should be enabled in the project. @avolkovi @tonyjhuang can you confirm?

src/file.ts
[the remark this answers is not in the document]
tonyjhuang: @crwilcox You can find them under the Firebase github org: https://github.com/firebase/

https://github.com/firebase/firebase-ios-sdk/blob/14764b8d60a6ad023d8fa5b7f81d42378d92e6fe/FirebaseStorage/Sources/FIRStorageGetDownloadURLTask.m#L94
https://github.com/firebase/firebase-android-sdk/blob/e65f4876ff61b5791986c249b21cb094c37bd29e/firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java#L80

tonyjhuang:
@crwilcox the `firebasestorage.googleapis.com` URI that `getFirebaseDownloadUrl()` will only work if the bucket has previously initialized Cloud Storage for Firebase in their Firebase project. Otherwise I would expect the server to return an error.

crwilcox:
So, to summarize:
- this feature only works for firebase users
- we instruct firebase users to use the admin sdk.
- the admin sdk includes this library, @google-cloud/storage

I truly don't meant to be obtuse, but I can't figure out why we wouldn't favor adding this to the Admin SDK, as we are certain users are using firebase, and we already include a dependency on @google-cloud/storage, so it wouldn't cause additional dependencies there.

hiranya911:
The simplest reason is that this is what would result in the best developer experience for the user. Following looks and feels quite natural:

```js
const bucket = admin.storage().bucket();
const file = bucket.file('test.txt');
console.log(file.getFirebaseDownloadUrl());
```

And also like I mentioned earlier, we don't have any actual GCS API in the Admin SDK to add anything to. It's literally just an initialization shim that returns a Bucket object from this package. This means we have 2 options at the Admin SDK end (in the order of my personal preference):

1. Provide some sort of a util API: `admin.storage().getDownloadUrl(file)`, which looks rather awkward to use.
2. Monkeypatch this library to dynamically add the new method to `File` class, which won't work across all languages.

So from a dev experience point of view, we only have suboptimal solutions at our end. Hence the suggested API in this PR.

nicoqh:
Seems to me that the burden of adding this Firebase-specific feature to GCS is greater than making space for it on the Firebase Admin SDK, where it (IMO) belongs.

kouariga:
Hi @hiranya911 @frankyn @tonyjhuang, has there been any update on this issue? I really appreciate the suggested feature and would like to see it merged.

PR #1323, opened 2020-10-22
fix: error if both `storageClass` and specific storage class name are provided

AVaksman:
`Storage#createBucket` accepts both REST `storageClass` type and a convenience specific name of the class type parameter. Creates confusion when both types provided at the same time.

Is this considered a breaking change?


src/storage.ts
```diff
@@ -641,6 +641,11 @@ export class Storage extends Service {
         delete body[storageClass];
       }
     });
+    if (body.storageClass && metadata.storageClass) {
```
stephenplusplus, line 644: This will throw if a user provides only one storageClass as well, such as this:

```js
it('should allow a user-specified storageClass', done => {
  storage.createBucket(
    BUCKET_NAME,
    {
      storageClass: 'nearline',
    },
    done
  );
});
```
```
  1) Storage
       createBucket
         should allow a user-specified storageClass:
     Error: Both `nearline` and `storageClass` were provided.
      at Storage.createBucket (build/src/storage.js:309:19)
          -> /Users/stephen/dev/nodejs-storage/src/storage.ts:645:13
      at Context.<anonymous> (build/test/index.js:439:21)
          -> /Users/stephen/dev/nodejs-storage/test/index.ts:592:15
      at processImmediate (internal/timers.js:456:21)
```

stephenplusplus (changes requested):

src/storage.ts
AVaksman: Done, and added the test.
Thanks.
PTAL

stephenplusplus (approved):

shaffeeullah:
@AVaksman can this be merged?

AVaksman:
> @AVaksman can this be merged?

Yes, I believe it is ready to be merged.
Thank you

shaffeeullah:
@AVaksman @stephenplusplus is this a breaking change? I see that you have this question in your PR description, but it is unanswered in the thread.

PR #1324, opened 2020-10-22
feat: public url of a file

### LINKED ISSUE #931: Request for `url`-like prop/method on files uploaded to storage
I'd like to get the URL of an image upon saving it to Storage, so I can save it in Firestore right away and display the image later. Right now there is `getSignedUrl()` and `metadata.mediaLink` (for downloading) but apparently we can view images with the following URLs:

_Public files_
> https://storage.googleapis.com/[BUCKET_NAME]/[IMAGE_PATH]

_Files viewable with access key_
> https://firebasestorage.googleapis.com/v0/b/[BUCKET_NAME]/o/[IMAGE_PATH]?alt=media&token=[ACCESS_TOKEN]

But there is no property or method that returns these URLs. It seems like people are just stitching them together in the understanding that this is a hack that might break in the future if the actual URL format happens to change.

I'd like a property or method to exist on a bucket file instance that would return either of the above mentioned URLs depending on whether the file is public or not, or whether an access token exists in the metadata.

Stitching the URL together by myself.

shaffeeullah:
Created a function publicUrl equivalent to [this python function](https://github.com/googleapis/python-storage/blob/2dc91c947e3693023b4478a15c460693808ea2d9/google/cloud/storage/blob.py#L317).

Fixes https://github.com/googleapis/nodejs-storage/issues/931

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -3061,6 +3061,25 @@ class File extends ServiceObject<File> {
     );
   }
 
+  /**
+   * The public URL of this File
+   * Use {@link File#makePublic} to enable anonymous access via the returned URL.
+   *
+   * @returns {string}
+   *
+   * @example
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const bucket = storage.bucket('albums');
+   * const file = bucket.file('my-file');
+   *
+   * file.publicUrl();
+   *  // Will return "https://storage.googleapis.com/albums/my-file"
+   */
+  publicUrl(): string {
+    return `${STORAGE_POST_POLICY_BASE_URL}/${this.bucket.name}/${this.name}`;
+  }
+
   move(
     destination: string | Bucket | File,
     options?: MoveOptions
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -3061,6 +3061,25 @@ class File extends ServiceObject<File> {
...
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const bucket = storage.bucket('albums');
+   * const file = bucket.file('my-file');
+   *
+   * file.publicUrl();
+   *  // Will return "https://storage.googleapis.com/albums/my-file"
+   */
+  publicUrl(): string {
+    return `${STORAGE_POST_POLICY_BASE_URL}/${this.bucket.name}/${this.name}`;
```
frankyn, line 3080: Please use: apiEndpoint from: https://github.com/googleapis/nodejs-storage/blob/master/src/storage.ts#L391

sofisl (approved):

src/file.ts
```diff
@@ -3061,6 +3061,25 @@ class File extends ServiceObject<File> {
...
+   *
+   * @returns {string}
+   *
+   * @example
+   * const {Storage} = require('@google-cloud/storage');
+   * const storage = new Storage();
+   * const bucket = storage.bucket('albums');
+   * const file = bucket.file('my-file');
+   *
+   * file.publicUrl();
```
frankyn, line 3076: Set publicUrl to a const variable.

PR #1328, opened 2020-10-28
feat: accessibleAt v4 signed urls

### LINKED ISSUE #1041: Add option to specify date in getSignedUrl

shaffeeullah:
Resolves merge conflicts from https://github.com/googleapis/nodejs-storage/pull/1079

Fixes #1041 🦕


src/file.ts
```diff
@@ -2723,6 +2724,10 @@ class File extends ServiceObject<File> {
    * @param {string} [config.responseDisposition] The
    *     [response-content-disposition parameter](http://goo.gl/yMWxQV) of the
    *     signed url.
+   * @param {*} [config.usableFrom=Date.now()] A timestamp when this link became usable. Any value
```
frankyn, line 2727: Is the name `usableFrom` readable to you or do you have recommendations for changing it?

system-test/storage.ts
```diff
@@ -3442,6 +3442,35 @@ describe('storage', () => {
...
+      const expiresDate = new Date();
+      const expiresMinutes = expiresDate.getMinutes();
+      const [signedReadUrl] = await file.getSignedUrl({
+        version: 'v4',
+        action: 'read',
+        usableFrom: usableFromDate.setMinutes(usableFromMinutes + 60),
+        expires: expiresDate.setMinutes(expiresMinutes + 90),
+      });
+      const res = await fetch(signedReadUrl);
+      const body = await res.text();
```
frankyn, line 3470: Instead of validating the text. Validate the res status code which is less brittle.

test/signer.ts
```diff
@@ -168,6 +168,108 @@ describe('signer', () => {
...
+          const query = {
+            'X-Goog-Date': expectedUsableFrom,
+          };
+          assert(signedUrl.includes(qsStringify(query)));
+        });
+
+        it('should accept strings', async () => {
+          const usableFrom = '12-12-2099';
+          const usableFromDate = new Date(usableFrom);
+          const expectedUsableFrom = '20991212T000000Z';
```
frankyn, line 230: Could you update this to be provided from the Date object instead?

test/signer.ts
```diff
@@ -168,6 +168,108 @@ describe('signer', () => {
...
+
+          const signedUrl = await signer.getSignedUrl({
+            version: 'v4',
+            method: 'GET',
+            usableFrom,
+            expires: usableFromDate.valueOf() + 5000,
```
frankyn, line 236: Why use `.valueOf()` here instead of setX()?

test/signer.ts
```diff
@@ -168,6 +168,108 @@ describe('signer', () => {
         });
       });
 
+      describe('usableFrom', () => {
+        it('should set correct settings if usableFrom provided', async () => {
+          const authClientSign: sinon.SinonStub<
+            [string],
+            Promise<string>
+          > = sandbox.stub(authClient, 'sign').resolves('signature');
+          const usableFrom = new Date(1581984000000);
```
frankyn, line 177: Nit: Make int constants named variables in a before() call so it can be reused.

test/signer.ts
```diff
@@ -168,6 +168,108 @@ describe('signer', () => {
...
+              version: 'v4',
+              method: 'GET',
+              usableFrom,
+              expires: Date.now() + 5000,
+            });
+          }, /The usable from date provided was invalid\./);
+        });
+
+        it('should throw if an expiration date from the before usableFrom date is given', () => {
+          const expires = Date.now() + 5;
```
frankyn, line 259: Based on our pairing:

`+ 5` is may not work as expected, could you update the line to increment based on the unit? e.g. second, minute, hour.

src/signer.ts
```diff
@@ -408,6 +419,16 @@ export class URLSigner {
 
     return Math.round(expiresInMSeconds / 1000); // The API expects seconds.
   }
+
+  parseUsableFrom(usableFrom?: string | number | Date): number {
```
frankyn, line 423: Did you try refactoring `parseUsableFrom()` into `parseExpires()` they look really similar?

src/signer.ts
```diff
@@ -140,6 +147,7 @@ export class URLSigner {
     const config: GetSignedUrlConfigInternal = Object.assign({}, cfg, {
       method,
       expiration: expiresInSeconds,
+      usableFrom: new Date(1000 * usableFromInSeconds),
```
frankyn, line 150: Nit: not sure I understand `1000 * ` in this case, is it needed? (Comparing to expiration and it's not used in this way.

frankyn (changes requested):

test/signer.ts
shaffeeullah: I have updated it to be 1 day difference. I am not incrementing by unit as to avoid converting to a date and then back to milliseconds. Do you think I should perform this conversion?

test/signer.ts
shaffeeullah: `setX` modifies and returns the original object

sofisl (approved):

sofisl:
@shaffeeullah, I reviewed this for readability, but will refer to @frankyn for library-specific code review.

src/file.ts
shaffeeullah: @frankyn what do you think of `accessibleAt`? (thanks @bcoe for the suggestion!)

src/file.ts
```diff
@@ -2747,7 +2746,8 @@ class File extends ServiceObject<File> {
    *
    * const config = {
    *   action: 'read',
-   *   expires: '03-17-2025'
+   *   expires: '03-17-2025',
+   *   accessibleAt: ''03-13-2025'
```
frankyn, line 2750: Double `'`.

Recommend adding another example separate from the simple Signed URL example.

frankyn (changes requested):
I saw you decided on `accessibleAt` sounds good and acking.

You have a linter issue and added a nit for the added example.

src/file.ts
```diff
@@ -2758,6 +2762,30 @@ class File extends ServiceObject<File> {
...
    * //-
+   * // Generate a URL that allows temporary access to download your file.
+   * // Access will begin at accessibleAt and end at expires.
+   * //-
+   * const request = require('request');
+   *
+   * const config = {
+   *   action: 'read',
+   *   expires: '03-17-2025',
+   *   accessibleAt: '03-13-2025'
```
frankyn, line 2773: One potential issue with `accessibleAt` is that it may not fit well with using a Signed URL to upload data only downloading data.

WDYT?

shaffeeullah: From how I read it, 'accessible' refers to the link and not the file. The link to upload is only accessible from the date specified. I think it works.

frankyn (approved):

PR #1347, opened 2020-12-01
feat: allow ignoring 404 errors during delete() operations

### LINKED ISSUE #1235: FR: Add File.delete options property to count missing file as success
I frequently find myself performing `delete` operations on files that may or may not exist. Similar to how Cloud Firestore's `delete` operation works, I would like for the Cloud Storage `delete` operation to consider a non-existent file as a successful deletion. In other words, I want my `delete` operation to be considered successful if the file does not exist after the operation completes- regardless of if it did exist before the operation.

The `delete` operation ([docs here](https://googleapis.dev/nodejs/storage/latest/File.html#delete)) takes an optional `options` object, whose only parameter is currently `userProject`. I would like a new property to be available that I could use to enable this mechanism on a per-operation basis. The syntax might look something like:

```
await myBucket.file("thisFileMightNotExist.txt").delete({mustExist: false})
```

The two alternatives are to either 1) run a `File.exists` operation first, or 2) handle deletion error in catch block. Both alternatives work, they're just not very elegant. Alternative 1 _technically_ also leaves an unlikely but possible scenario where the file is removed after the the `exists` operation but before the `delete` operation.

 **Additional context**
None.

stephenplusplus:
Fixes #1235

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "api-documenter": "api-documenter yaml --input-folder=temp"
   },
   "dependencies": {
-    "@google-cloud/common": "^3.3.0",
+    "@google-cloud/common": "^3.5.0",
     "@google-cloud/paginator": "^3.0.0",
     "@google-cloud/promisify": "^2.0.0",
     "arrify": "^2.0.0",
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -666,6 +666,8 @@ class Bucket extends ServiceObject {
        *
        * @method Bucket#delete
        * @param {DeleteBucketOptions} [options] Configuration options.
+       * @param {boolean} [options.ignoreNotFound = false] Ignore an error if
+       *     the bucket does not exist.
        * @param {DeleteBucketCallback} [callback] Callback function.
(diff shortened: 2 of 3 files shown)


AVaksman:
Do [`DeleteFileOptions`](https://github.com/googleapis/nodejs-storage/blob/cbb2c258dcaed5cd475481932844458cc520a5f8/src/file.ts#L174) and [`DeleteBucketOptions`](https://github.com/googleapis/nodejs-storage/blob/cbb2c258dcaed5cd475481932844458cc520a5f8/src/bucket.ts#L171) interfaces need to be updated?

shaffeeullah:
Should there be system/unit tests for this functionality?

stephenplusplus:
@shaffeeullah I don't believe we should do that here, as that testing is done in the common library, where the actual code for the `ignoreNotFound` option lives. We don't even have a function body for "delete()" in this layer-- it is entirely defined and executed from the common library's `ServiceObject.prototype.delete` method.

shaffeeullah (approved):

PR #1354, opened 2020-12-07
docs: add properties and missing methods

### LINKED ISSUE #1227: Include object properties in API reference
As far as I can tell, the API reference only details the methods available on an object, and not the properties.
For example, a `File` object has properties `name`, `metadata`, etc, but this is nowhere to be found in the API reference.
Please add documentation for object properties.

stephenplusplus:
Fixes #1227 

This PR:

- Adds `name` and `metadata` to the Class documentation, where applicable
- Adds missing methods to the docs
  - Bucket.acl methods, Bucket.iam methods, Bucket.getFilesStream()
  - File.acl methods, File.iam methods
  - Storage.acl methods, Storage.getBucketsStream(), Storage.getHmacKeysStream()

It seems to be a matter of location to entice the JSDocs parser to recognize an instance-property as a "Member" and a constructor-defined method (e.g. `this.getBucketsStream = () => {...}`) as a documentable method.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -379,6 +379,197 @@ type MakeAllFilesPublicPrivateResponse = [File[]];
 const RESUMABLE_THRESHOLD = 5000000;
 
 /**
+ * Get and set IAM policies for your bucket.
+ *
+ * @name Bucket#iam
+ * @mixes Iam
+ *
+ * @see [Cloud Storage IAM Management](https://cloud.google.com/storage/docs/access-control/iam#short_title_iam_management)
+ * @see [Granting, Changing, and Revoking Access](https://cloud.google.com/iam/docs/granting-changing-revoking-access)
+ * @see [IAM Roles](https://cloud.google.com/iam/docs/understanding-roles)
+ *
+ * @example
+ * const {Storage} = require('@google-cloud/storage');
+ * const storage = new Storage();
+ * const bucket = storage.bucket('albums');
+ *
+ * //-
+ * // Get the IAM policy for your bucket.
+ * //-
+ * bucket.iam.getPolicy(function(err, policy) {
+ *   console.log(policy);
+ * });
+ *
+ * //-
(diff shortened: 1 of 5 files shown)


src/bucket.ts
```diff
@@ -378,6 +378,197 @@ type MakeAllFilesPublicPrivateResponse = [File[]];
...
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * myBucket.acl.add(options).then(function(data) {
+ *   const aclObject = data[0];
+ *   const apiResponse = data[1];
+ * });
+ *
+ * @example <caption>include:samples/acl.js</caption>
+ * region_tag:storage_print_bucket_acl
+ * Example of printing a bucket's ACL:
```
shaffeeullah, line 483: Should there be an example here?

src/bucket.ts
```diff
@@ -378,6 +378,197 @@ type MakeAllFilesPublicPrivateResponse = [File[]];
...
+ *
+ * @example <caption>include:samples/acl.js</caption>
+ * region_tag:storage_print_bucket_acl_for_user
+ * Example of printing a bucket's ACL for a specific user:
```
shaffeeullah, line 487: Should there be an example here?

src/bucket.ts
```diff
@@ -378,6 +378,197 @@ type MakeAllFilesPublicPrivateResponse = [File[]];
...
+ *
+ * @example <caption>include:samples/acl.js</caption>
+ * region_tag:storage_add_bucket_owner
+ * Example of adding an owner to a bucket:
```
shaffeeullah, line 491: Should there be an example here? (Same for below this as well.)

src/bucket.ts
stephenplusplus: It looks weird, but here is how that actually renders: 
image

shaffeeullah (approved):

PR #1355, opened 2020-12-07
feat: support metadata updates from makePrivate() methods

### LINKED ISSUE #1353: make private and set MetaData in one step
I am using your storage system, but I realize that when you want to make a public or private file and also set metadata to said file, this has to be done in two steps, there is some way how it can be done in a single line 
Ej: `await bucket.file(destination).makePrivate().setMetadata({metadata});`
because this slows down the app and also translates at a higher cost

image

stephenplusplus:
Fixes #1353


shaffeeullah (approved):
Thanks for implementing this!

PR #1370, opened 2020-12-23
fix: deprecated directory. prefix should be used instead

### LINKED ISSUE #1343: Constant value delimiter for directory option for bucket.getFiles
I was studying bucket.getFiles() options and found a property called "directory".  I couldn't understand what this did or how it should work so I examined the source code and came to this source:

https://github.com/googleapis/nodejs-storage/blob/master/src/bucket.ts#L2234

at that location, we find the implementation of the directory option property.  If I am reading it correctly, it appears to assume that the delimiter will be fixed to be '/'.  While I fully agree that I can't imagine anyone wanting a delimiter other than this, our documentation does allow for arbitrary delimiters as specified by the delimiter option.  This report asks for y'all to consider whether or not the "directory" option needs to consider the value that the programmer passed in for "delimiter".

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1343 🦕

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -113,7 +113,8 @@ export interface EnableLoggingOptions {
 export interface GetFilesOptions {
   autoPaginate?: boolean;
   delimiter?: string;
-  directory?: string; //deprecated
+  /** @deprecated */
+  directory?: string;
   endOffset?: string;
   includeTrailingDelimiter?: boolean;
   prefix?: string;
@@ -2196,7 +2197,7 @@ class Bucket extends ServiceObject {
    *     truncated after the delimiter, returned in `apiResponse.prefixes`.
    *     Duplicate prefixes are omitted.
    * @deprecated @property {string} [directory] Filter results based on a directory name, or
-   *     more technically, a "prefix". Deprecated. Use prefix instead.
+   *     more technically, a "prefix". Assumes delimeter to be '/'. Deprecated. Use prefix instead.
    * @property {string} [endOffset] Filter results to objects whose names are
    * lexicographically before endOffset. If startOffset is also set, the objects
    * listed have names between startOffset (inclusive) and endOffset (exclusive).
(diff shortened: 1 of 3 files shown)


src/bucket.ts
```diff
@@ -2195,8 +2195,8 @@ class Bucket extends ServiceObject {
    *     names, aside from the prefix, contain delimiter will have their name
    *     truncated after the delimiter, returned in `apiResponse.prefixes`.
    *     Duplicate prefixes are omitted.
-   * @property {string} [directory] Filter results based on a directory name, or
-   *     more technically, a "prefix".
+   * @deprecated @property {string} [directory] Filter results based on a directory name, or
+   *     more technically, a "prefix". Deprecated. Use prefix instead.
```
frankyn, line 2199: Please add a note that directory uses delimiter='/' in case folks get confused. 

Recommend including an example of using prefix and delimiter below to reference.

src/bucket.ts
```diff
@@ -113,7 +113,7 @@ export interface EnableLoggingOptions {
 export interface GetFilesOptions {
   autoPaginate?: boolean;
   delimiter?: string;
-  directory?: string;
+  directory?: string; //deprecated
```
frankyn, line 116: Is this the recommended approach to declare deprecation in an interface? I don't see another example in the project that deprecates this way, and want to confirm.

frankyn (changes requested):

src/bucket.ts
shaffeeullah: Hi Frank, [this example](https://github.com/googleapis/nodejs-storage/blob/6ac93b823966081403e3269adced7a898ffdd4cc/src/bucket.ts#L2333) includes prefix and delimiter. Are you looking for something different?

frankyn (approved):

PR #1385, opened 2021-01-20
fix: retry multipart uploads on File.save()

### LINKED ISSUE #1281: Retry Strategy should be enabled by default for multipart uploads

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1281 🦕


src/file.ts
```diff
@@ -3560,15 +3564,31 @@ class File extends ServiceObject<File> {
...
     const options =
       typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
-
-    const writable = this.createWriteStream(options)
-      .on('error', callback!)
-      .on('finish', callback!);
-    if (options.onUploadProgress) {
-      writable.on('progress', options.onUploadProgress);
+    const isMultipart = options.resumable === false;
+    const returnValue = retry(async (bail: (err: Error) => void) => {
```
frankyn, line 3568: This is a future todo: How can retry config be uniform between this retry logic versus what's in nodejs-common repo?

src/file.ts
```diff
@@ -3816,7 +3836,7 @@ class File extends ServiceObject<File> {
  * that a callback is omitted.
  */
 promisifyAll(File, {
-  exclude: ['publicUrl', 'request', 'setEncryptionKey'],
+  exclude: ['publicUrl', 'request', 'save', 'setEncryptionKey'],
```
frankyn, line 3839: is this considered a breaking change?

test/file.ts
```diff
@@ -3853,6 +3865,151 @@ describe('File', () => {
...
+        };
+        await file.save(BUFFER_DATA, options);
+        assert.strictEqual(streamFinished, true);
+      });
+
+      it('buffer upload should retry on first failure', async () => {
+        const options = {resumable: false};
+        let retryCount = 0;
+        file.createWriteStream = (options_: {}) => {
+          class DelayedStream extends Transform {
```
frankyn, line 3935: Can you make this a test helper class that can be deduped for each use-case?

frankyn (changes requested):

src/file.ts
shaffeeullah: It is not. Instead of `save` being promisified here at the end, it is promisified in the function itself.

src/file.ts
```diff
@@ -3517,6 +3519,8 @@ class File extends ServiceObject<File> {
    * Resumable uploads are automatically enabled and must be shut off explicitly
    * by setting `options.resumable` to `false`.
    *
+   * Multipart uploads with retryable error codes will be retried 10 times with exponential backoff.
+   *
    * <p class="notice">
    *   There is some overhead when using a resumable upload that can cause
```
tritone, line 3525: Out of curiosity, is this true? The node library doesn't automatically choose multipart for small files?

src/file.ts
```diff
@@ -3517,6 +3519,8 @@ class File extends ServiceObject<File> {
    * Resumable uploads are automatically enabled and must be shut off explicitly
    * by setting `options.resumable` to `false`.
    *
+   * Multipart uploads with retryable error codes will be retried 10 times with exponential backoff.
```
tritone, line 3522: Do resumable uploads not retry currently? If they do, probably good to add docs about both.

src/file.ts
```diff
@@ -3560,15 +3564,31 @@ class File extends ServiceObject<File> {
...
-
-    const writable = this.createWriteStream(options)
-      .on('error', callback!)
-      .on('finish', callback!);
-    if (options.onUploadProgress) {
-      writable.on('progress', options.onUploadProgress);
+    const isMultipart = options.resumable === false;
+    const returnValue = retry(async (bail: (err: Error) => void) => {
+      await new Promise<void>((resolve, reject) => {
+        const writable = this.createWriteStream(options)
```
tritone, line 3570: Do end users ever use streams directly, or is it more typical to use save?

src/file.ts
frankyn: It does not have a short circuit on that. It depends on a config option named `resumable` to be set to false for a multipart. By default it's always true (resumable)

src/file.ts
frankyn: That's the open question raised in today's meeting. We aren't sure. It might not a use-case for `file.save()` and @shaffeeullah is looking into it.

src/file.ts
```diff
@@ -3587,7 +3587,7 @@ class File extends ServiceObject<File> {
     if (!callback) {
       return returnValue;
     } else {
-      return returnValue.then(callback);
+      return returnValue.then(callback).catch(callback);
```
frankyn, line 3590: This looks wrong, but I'm not sure. Should there be two different values for `then()` and `catch()`?

bcoe: The first argument to the closure passed to `catch` will be `err` if it's invoked, the `then` in this case will be invoked with no arguments. So, I believe this behavior works as expected `callback()` when there's a success `callback(err)` on failure.

We should make sure there's a test.

src/file.ts
shaffeeullah: Which logic in `nodejs-common` are you referring to here?

src/file.ts
shaffeeullah: @bcoe I believe the error case is already covered by [this system test ](https://github.com/googleapis/nodejs-storage/blob/1d0a7a6aa1c016cc73da2088c083ca571d7aebd9/system-test/storage.ts#L1493) as well as [this unit test](https://github.com/googleapis/nodejs-storage/blob/1d0a7a6aa1c016cc73da2088c083ca571d7aebd9/test/file.ts#L3979). The success case is covered by multiple tests such as [this one ](https://github.com/googleapis/nodejs-storage/blob/1d0a7a6aa1c016cc73da2088c083ca571d7aebd9/test/file.ts#L3944). Is there another method of testing this that you envision?

src/file.ts
frankyn: I was referring to [autoRetry and maxRetries](https://github.com/googleapis/nodejs-common/blob/611d16ba465b53d5eeb3ad59d37a2501a77c3b87/src/util.ts#L118-L134) options.

frankyn (approved):
LGTM!

src/file.ts
shaffeeullah: @frankyn Do you know if there is existing documentation on resumable upload retries? I'm struggling to find it. @tritone, good callout for adding this documentation, but I'm not sure it belongs here. Multipart retries happen in this function (`file.save`), so it is documented on the function. If resumable upload retries happen, they happen in another function.

src/file.ts
shaffeeullah: PR here: https://github.com/googleapis/nodejs-storage/pull/1388

src/file.ts
shaffeeullah: If I understand correctly, this logic is retrying certain requests. What we're doing here is retrying all the logic, not just the request. I believe that in order to do something similar, we'll have to refactor `File.save` so it passes `data` to `File.createWriteStream`. Then, on request failure, we can do a similar retry.

src/file.ts
```diff
@@ -64,6 +64,8 @@ import {
 const duplexify: DuplexifyConstructor = require('duplexify');
 import {normalize, objectKeyToLowercase, unicodeJSONStringify} from './util';
 import {GaxiosError, Headers, request as gaxiosRequest} from 'gaxios';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
bcoe, line 67: It would be worth figuring out how to use this library with an `import`.

Did we try:

```
import retry from 'async-retry'
```

test/file.ts
```diff
@@ -3853,6 +3865,129 @@ describe('File', () => {
 
   describe('save', () => {
     const DATA = 'Data!';
+    const BUFFER_DATA = Buffer.from(DATA, 'utf8');
+
+    class DelayedStreamNoError extends Transform {
+      _transform(chunk: any, _encoding: string, done: Function) {
```
bcoe, line 3871: might b e worth giving chunk a type, I believe it will be a `Buffer` or `string`.

test/file.ts
```diff
@@ -3853,6 +3865,129 @@ describe('File', () => {
...
+          return new DelayedStream500Error(retryCount);
+        };
+        await file.save(BUFFER_DATA, options);
+        assert.ok(retryCount === 2);
+      });
+
+      it('non-multipart upload should not retry', async () => {
+        const options = {resumable: true};
+        let retryCount = 0;
+        file.createWriteStream = (options_: {}) => {
```
bcoe, line 3966: Can you just get rid of the `option_` if you're not using it? alternatively putting a leading `_` rather than trailing `_` might satisfy the linter.

bcoe (approved):

frankyn (approved):

PR #1388, opened 2021-01-22
fix: specified acceptable types for File.save data parameter

shaffeeullah:
Acceptable types for the `File.save` data parameter were documented to be `any`. However, this is not the case, as only `string` and `Buffer` types can be passed in to this parameter. If anything else is passed in, it fails with the following error:
image

This PR updates the `data` parameter to reflect current functionality.


frankyn (approved):
LGTM

src/file.ts
```diff
@@ -3551,7 +3552,7 @@ class File extends ServiceObject<File> {
    */
   save(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
bcoe, line 3554: We can get rid of `// eslint-disable-next-line @typescript-eslint/no-explicit-any`.

bcoe (changes requested):
If there were no other values for data we accepted that would not have resulted in throwing, this seems like a reasonable change without calling it breaking.

I think we should be careful about releasing this, and be prepared to rollback however -- so I'd be tempted to release your `retry` logic first, and then release this a couple days later.

frankyn (changes requested):
I missed a few lines:

https://github.com/googleapis/nodejs-storage/blob/03f6cebd1a977b56434ecc84bd7abaa4261356cc/src/file.ts#L3498-L3503

@bcoe do you have ideas how this could break? (I'm by no means a Node.js expert.)

shaffeeullah:
@bcoe does this line need to be removed as well? https://github.com/googleapis/nodejs-storage/blob/11e94456c91dcd3442f5ec33bc9bf50c22aa67a4/src/file.ts#L3559

bcoe (approved):

frankyn (approved):

Issue #1284, opened 2020-09-03
Provide a way to pass in an authenticated auth client


JustinBeckwith:
In the GAPIC clients, I can do something like this:

```ts
const {GoogleAuth} = require('google-auth-library');
const {CloudBuildClient} = require('@google-cloud/cloudbuild');

const auth = new GoogleAuth();
const buildy = new CloudBuildClient({ auth: auth });
```

This is nice because I can create a single auth client, and then share it across multiple services.  It drastically reduces process startup when metadata servers are involved.  However ... I can't seem to find a way to do this with our hand written libraries :(  Is there a way I don't know about to pass a pre-obtained auth client?

stephenplusplus:
This was meant to be possible, so I'm not sure if we lost it along the way, or we accidentally left some wires unconnected. But it should be easy enough. PR coming!

stephenplusplus:
PR sent: https://github.com/googleapis/nodejs-common/pull/623

stephenplusplus:
This was released in [@google-cloud/common@3.4.0](https://github.com/googleapis/nodejs-common/releases/tag/v3.4.0) and is usable immediately with new @google-cloud/storage installations:

```js
const {GoogleAuth} = require('google-auth-library');
new Storage({ authClient: new GoogleAuth() });
```

oscar60310:
Hi @stephenplusplus , it seems we need to use `ServiceOptions` instead of `GoogleAuthOptions` as the base interface of storage options to apply your [changes](https://github.com/googleapis/nodejs-common/pull/623) at nodejs-common.

https://github.com/googleapis/nodejs-storage/blob/master/src/storage.ts#L48

stephenplusplus:
@oscar60310 thanks! I sent a PR here: https://github.com/googleapis/nodejs-storage/pull/1391, please take a look!

PR #1391, opened 2021-01-27
fix: deprecate `options.promise` and sync options with Service

stephenplusplus:
As brought up in https://github.com/googleapis/nodejs-storage/issues/1284#issuecomment-768004492 (thanks, @oscar60310!), the options to the Storage constructor is `GoogleAuthOptions`, but I believe it was meant to be `ServiceOptions`:

```ts
// From @google-cloud/common
export interface ServiceOptions extends GoogleAuthOptions {
  authClient?: GoogleAuth;
  interceptors_?: Interceptor[];
  email?: string;
  token?: string;
  timeout?: number; // http.request.options.timeout
  userAgent?: string;
}

// From @google-cloud/storage
export interface StorageOptions extends GoogleAuthOptions {
  autoRetry?: boolean;
  maxRetries?: number;
  promise?: typeof Promise;
  userAgent?: string;
  apiEndpoint?: string;
}
```

A couple of other fixes and modernizations:

- `autoRetry` was not passed to the Service constructor
- `maxRetries` was not passed to the Service constructor
- The `promise` property was dropped a while ago: https://github.com/googleapis/nodejs-common/issues/107

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {GoogleAuthOptions, Metadata, Service} from '@google-cloud/common';
+import {Metadata, Service, ServiceOptions} from '@google-cloud/common';
 import {paginator} from '@google-cloud/paginator';
 import {promisifyAll} from '@google-cloud/promisify';
 
@@ -45,7 +45,7 @@ export interface CreateBucketQuery {
   userProject: string;
 }
 
-export interface StorageOptions extends GoogleAuthOptions {
+export interface StorageOptions extends ServiceOptions {
   autoRetry?: boolean;
   maxRetries?: number;
   promise?: typeof Promise;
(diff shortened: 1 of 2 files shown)


shaffeeullah (approved):

JustinBeckwith (changes requested):
Lets hold up on the breaking change here.  Lets cover in the chat?

src/storage.ts
```diff
@@ -45,11 +45,9 @@ export interface CreateBucketQuery {
   userProject: string;
 }
 
-export interface StorageOptions extends GoogleAuthOptions {
+export interface StorageOptions extends ServiceOptions {
   autoRetry?: boolean;
   maxRetries?: number;
-  promise?: typeof Promise;
```
JustinBeckwith, line 51: Which it likely doesn't work anymore, yanking this triggers the breaking change, which I want to avoid.  Can we just leave this here, and maybe flag it as deprecated?  To your point, I don't think it works anymore, and probably should have been removed in the last semver major.  

I think we should create a tracking issue for future breaking changes we want to make, so we can queue a few of these up as they come.

stephenplusplus: 👍🏻 Restored with note.

JustinBeckwith (approved):

PR #1400, opened 2021-02-10
docs: Clarify how to specify projectID in samples

tritone:
Adds a comment to quickstart sample as well as bucket creation
samples (which send the project ID in the request) to clarify that
Storage can be initialized for a specific project.

Fixes an internal feedback bug on these samples.

### DIFF
--- a/samples/createBucketWithStorageClassAndLocation.js
+++ b/samples/createBucketWithStorageClassAndLocation.js
@@ -40,6 +40,9 @@ function main(
   // Creates a client
   const storage = new Storage();
 
+  // Creates a client using a specific project ID.
+  // const storage = new Storage({"my-gcp-project"});
+
   async function createBucketWithStorageClassAndLocation() {
     // For default values see: https://cloud.google.com/storage/docs/locations and
     // https://cloud.google.com/storage/docs/storage-classes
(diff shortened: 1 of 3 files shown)


samples/createBucketWithStorageClassAndLocation.js
```diff
@@ -40,6 +40,9 @@ function main(
   // Creates a client
   const storage = new Storage();
 
+  // Creates a client using a specific project ID.
+  // const storage = new Storage({"my-gcp-project"});
```
JustinBeckwith, line 44: We would actually want:
```js
const storage = new Storage({ projectId: 'my-gcp-project' });
```

JustinBeckwith (changes requested):
I worry a little here from a consistency standpoint.  We have thousands of samples where we don't specifically show identifying a custom Project Id, why are these particular ones special?

In 99/100 cases, I would expect folks want to use the same project id that either:
- Is specified by `gcloud config set projectId
- Defined by the `GCLOUD_PROJECT` env var
- Obtained from the GCE metadata service

These libraries do that detection automatically, so I want to be *super* careful on how we advise folks here.

samples/createBucketWithStorageClassAndLocation.js
frankyn, line 44: Could you include the path to credentials as well?

frankyn (changes requested):

shaffeeullah:
Resolved by https://github.com/googleapis/nodejs-storage/pull/1413

PR #1406, opened 2021-02-18
fix(types): support metadata override in file.copy()

### LINKED ISSUE #1398: Allow setting metadata when copying
We store some metadata on images regarding which post a certain image belongs to.
We have some cloud functions that listen to new files being added and depending on the metadata we do some actions.

We now want to be able to copy a file to a new location.
But because our triggers fire of as soon as we use the copy command we would like to be able to specify what should happen with the metadata.

https://googleapis.dev/nodejs/storage/latest/File.html#copy
https://googleapis.dev/nodejs/storage/latest/global.html#CopyOptions

Either we should be able to say, skip copying the metadata.
```
const file = bucket.file('my-image.png');
file.copy('my-image-copy.png', {
  metadata: null
}
```
Or we should be able to set the metadata from the copy method.

```
const file = bucket.file('my-image.png');
file.copy('my-image-copy.png', {
  metadata: {
    thisIsACopy: true
  }
}
```

## What we have tried

- downloading the file and uploading again. (takes to long, and means extra costs)
- removing the metadata on the source before copying (not ideal because then we need to put the metadata back again)

Would this be possible?

stephenplusplus:
Fixes #1398 

`file.copy()` allows overwriting the original file's metadata:

```js
const originalFile = bucket.file('original-file-name')
const [copiedFile] = await originalFile.copy('new-file-name', {metadata: {thisIsACopy: 'true'}})
const [metadata] = await copiedFile.getMetadata()
// metadata.metadata.thisIsACopy === 'true'
```

But, we didn't document it through JSDocs or type docs. This fixes that and adds a system test.

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -2810,6 +2810,24 @@ describe('storage', () => {
       await Promise.all([file.delete, copiedFile.delete()]);
     });
 
+    it('should copy an existing file and overwrite metadata', async () => {
+      const opts = {
+        destination: 'CloudLogo',
+        metadata: {
+          metadata: {
+            originalProperty: 'true',
+          },
+        },
+      };
+      const [file] = await bucket.upload(FILES.logo.path, opts);
+      const copyOpts = {metadata: {newProperty: 'true'}};
+      const [copiedFile] = await file.copy('CloudLogoCopy', copyOpts);
+      const [metadata] = await copiedFile.getMetadata();
+      assert.strictEqual(typeof metadata.metadata.originalProperty, 'undefined');
+      assert.strictEqual(metadata.metadata.newProperty, 'true');
+      await Promise.all([file.delete, copiedFile.delete()]);
+    });
+
     it('should respect predefined Acl at file#copy', async () => {
       const opts = {destination: 'CloudLogo'};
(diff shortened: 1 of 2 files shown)


system-test/storage.ts
```diff
@@ -2810,6 +2810,27 @@ describe('storage', () => {
...
+      const opts = {
+        destination: 'CloudLogo',
+        metadata: {
+          metadata: {
+            originalProperty: 'true',
+          },
+        },
+      };
+      const [file] = await bucket.upload(FILES.logo.path, opts);
+      const copyOpts = {metadata: {newProperty: 'true'}};
```
shaffeeullah, line 2823: I'm a little confused about how this interface works. Here you have `newProperty`, in the PR description you have `thisIsACopy`, and above you have `originalProperty`. I see above you specify this field as type `Metadata`, which is in @google-cloud/common as `any`. How does the code know what to do with this property? Which one of these should it be specified as? Can we improve documentation for this?

stephenplusplus: These are arbitrary property names and values. `metadata` is part of a GCS Object that can be used for user-defined key/values.

shaffeeullah (approved):

PR #1407, opened 2021-02-21
feat: add generation support to createReadStream for download of other versions

andreyk-code:


shaffeeullah:
Hi @andreynocap , this feature is already implemented in the library. It can be used via the `generation` property on `File`. Example:
```
const remoteFile = bucket.file('image.png');
remoteFile.generation = 23409239040
// Then create read stream
remoteFile.createReadStream()
  .on('error', function(err) {})
  .on('response', function(response) {  
  })
  .on('end', function() {
     // The file is fully downloaded.
  })
  .pipe(fs.createWriteStream(localFilename));
  ```

Another example of using this property is in this [sample](https://cloud.google.com/storage/docs/using-object-versioning#copy) .

shaffeeullah:
Closing due to feature already existing in the library.

PR #1416, opened 2021-03-03
docs(samples): standardize samples

stephenplusplus:
Standardize samples with the canonical samples resource, [Java](https://github.com/googleapis/google-cloud-java/tree/master/google-cloud-examples/src/main/java/com/google/cloud/examples/storage), following the [guidelines](https://docs.google.com/document/d/1HXdroH24LfNRlsuWYXH5wOeaAC9y73WVnUI-eJSBqgw/edit?resourcekey=0-G_6ID_1JogvVhxoQNggFLw#).


samples/downloadPublicFile.js
```diff
@@ -43,15 +43,14 @@ function main(
 
   async function downloadPublicFile() {
     const options = {
-      // The path to which the file should be downloaded, e.g. "./file.txt"
```
stephenplusplus, line 46: Removed, since this is described where `destFileName` is defined.

samples/listFiles.js
```diff
@@ -21,6 +21,7 @@
  */
 
 function main(bucketName = 'my-bucket') {
+  // @TODO There are two samples with ID: "storage_list_files"
```
stephenplusplus, line 24: ⚠️

samples/listFilesPaginate.js
```diff
@@ -14,6 +14,7 @@
 /* eslint-disable no-unused-vars */
 
 function main(bucketName = 'my-bucket') {
+  // @TODO There are two samples with ID: "storage_list_files"
```
stephenplusplus, line 17: ⚠️

samples/addBucketLabel.js
```diff
@@ -41,12 +41,13 @@ function main(
...
   };
+
   async function addBucketLabel() {
     await storage.bucket(bucketName).setLabels(labels);
-    console.log(`Added label to bucket ${bucketName}.`);
+    console.log(`Added label to bucket ${bucketName}`);
   }
 
-  addBucketLabel();
+  addBucketLabel().catch(console.error);
```
JustinBeckwith, line 50: 👋   we don't want these `.catch(console.error)` things spread about.  This isn't how customers would handle errors in production, so we want to let it throw.  Adding in @fhinkel who owns the node.js rubric to chime in as well.

samples/getMetadata.js
```diff
@@ -41,34 +41,44 @@ function main(bucketName = 'my-bucket', filename = 'test.txt') {
...
       .getMetadata();
 
-    console.log(`File: ${metadata.name}`);
     console.log(`Bucket: ${metadata.bucket}`);
-    console.log(`Storage class: ${metadata.storageClass}`);
-    console.log(`Self link: ${metadata.selfLink}`);
-    console.log(`ID: ${metadata.id}`);
-    console.log(`Size: ${metadata.size}`);
-    console.log(`Updated: ${metadata.updated}`);
+    console.log(`CacheControl: ${metadata.cacheControl}`);
```
JustinBeckwith, line 45: Not required, but throwing this out there ...  
```js
for (const key in metadata) {
  console.log(`${key}: ${metadata[key]}`);
}
```

JustinBeckwith (commented):
(my feedback is nonblocking)

samples/getMetadata.js
stephenplusplus: Haha, I like that as well. The parent sample calls them out line by line, which I assume is to show the sample reader what they'll get without requiring them to run it. It's definitely not the most future-proof plan, but assuming these properties won't be removed, the annoyance of typing them all out is already spent :)

samples/addBucketLabel.js
stephenplusplus: Haha, whoops. It seemed like more samples had them than not, so I caught up the ones that didn't. I'll wait for @fhinkel confirmation, but happy to revert and remove any others.

samples/listFilesPaginate.js
shaffeeullah: Great catch!! Can you please change the region tag for this sample to be `storage_list_files_paginated`?

JesseLovelace (changes requested):
Hi Stephen, thanks for this!

It looks like the comments that contextualize parameters are not following the format laid out in the guidelines, please review section 2. 

For example, from addBucketLabel.js:

`// const bucketName = 'Name of a bucket, e.g. my-bucket';`

Should instead be

```
// The ID of your GCS bucket
// const bucketName = 'your-unique-bucket-name';
```

stephenplusplus:
Thanks for the review @JesseLovelace! Sorry that I rushed my understanding of section 2. I went back over and followed the guidelines more closely now. Please take a look and let me know if I missed anything!

samples/notifications.js
```diff
@@ -1,168 +0,0 @@
-// Copyright 2019 Google LLC
```
stephenplusplus, line 1: I removed this file, since the samples within it are located in individual files.

samples/listFilesPaginate.js
stephenplusplus: Done!

JesseLovelace (approved):

JesseLovelace:
Thanks Stephen!

PR #1419, opened 2021-03-10
fix(perf): pull hashes without refreshing metadata

### LINKED ISSUE #1338: Sequential requests when downloading file
I'm using the `File.download` method and was wondering if its performance could be improved.

Currently it makes two requests sequentially: 

1. /storage/.../key?alt=media to download the file
2. /storage/.../key to download file metadata

Would it be possible to make those requests in parallel? The second request adds another 20 - 250ms to the download.

stephenplusplus:
Fixes #1338

This PR narrows the conditions for when a metadata refresh is required during a file download.

Before:

1. user calls file.createReadStream()
1. file finishes downloading
1. we call `file.getMetadata()` to get the remote file's hashes, **even if the user disabled data integrity (validation: `false`)**
1. if validation wasn't explicitly disabled, we confirm data integrity, comparing locally computed hashes vs. hashes from the API

After:

1. user calls file.createReadStream()
1. our API call to begin the download grabs the remote file's hashes from the response headers (`x-goog-hash`)
1. file finishes downloading
1. if validation wasn't explicitly disabled, we confirm data integrity, comparing locally computed hashes vs. hashes from the first call to the API (step 2)

So, one less API call.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1211,8 +1211,6 @@ class File extends ServiceObject<File> {
     let crc32c = true;
     let md5 = false;
 
-    let refreshedMetadata = false;
-
     if (typeof options.validation === 'string') {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       (options as any).validation = (options.validation as string).toLowerCase();
@@ -1221,6 +1219,8 @@ class File extends ServiceObject<File> {
     } else if (options.validation === false) {
       crc32c = false;
     }
+
+    const shouldRunValidation = !rangeRequest && (crc32c || md5);
 
     if (rangeRequest) {
       if (
@@ -1268,6 +1268,8 @@ class File extends ServiceObject<File> {
         qs: query,
       };
 
+      const hashes: {crc32c?: string; md5?: string} = {};
+
       this.requestStream(reqOpts)
         .on('error', err => {
           throughStream.destroy(err);
@@ -1307,10 +1309,22 @@ class File extends ServiceObject<File> {
 
         const headers = rawResponseStream.toJSON().headers;
         isServedCompressed = headers['content-encoding'] === 'gzip';
-        const shouldRunValidation = !rangeRequest && (crc32c || md5);
         const throughStreams: Writable[] = [];
 
         if (shouldRunValidation) {
+          // The x-goog-hash header should be set with a crc32c and md5 hash.
+          // ex: headers['x-goog-hash'] = 'crc32c=xxxx,md5=xxxx'
+          if (typeof headers['x-goog-hash'] === 'string') {
+            headers['x-goog-hash']
+              .split(',')
+              .forEach((hashKeyValPair: string) => {
+                const delimiterIndex = hashKeyValPair.indexOf('=');
+                const hashType = hashKeyValPair.substr(0, delimiterIndex);
+                const hashValue = hashKeyValPair.substr(delimiterIndex + 1);
+                hashes[hashType as 'crc32c' | 'md5'] = hashValue;
+              });
+          }
+
           validateStream = hashStreamValidation({crc32c, md5});
           throughStreams.push(validateStream);
         }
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -1307,10 +1309,22 @@ class File extends ServiceObject<File> {
...
+          // The x-goog-hash header should be set with a crc32c and md5 hash.
+          // ex: headers['x-goog-hash'] = 'crc32c=xxxx,md5=xxxx'
+          if (typeof headers['x-goog-hash'] === 'string') {
+            headers['x-goog-hash']
+              .split(',')
+              .forEach((hashKeyValPair: string) => {
+                const delimiterIndex = hashKeyValPair.indexOf('=');
+                const hashType = hashKeyValPair.substr(0, delimiterIndex);
+                const hashValue = hashKeyValPair.substr(delimiterIndex + 1);
+                hashes[hashType as 'crc32c' | 'md5'] = hashValue;
```
frankyn, line 1324: Reviewed this PR with @shaffeeullah on md5 hashes she'll follow-up on remaining Node.js specific changes:

1. Composed objects will not support md5. Could you add a warning / message to users that it can't be validated because it's a composed object (only md5: https://cloud.google.com/storage/docs/composite-objects#metadata).
2. Could you also add a unit test for this case?


Example of a non-composed GCS Object response headers:
```
HTTP/1.1 200 OK
X-goog-metageneration: 53
Content-location: https://storage.googleapis.com/storage/v1/b/anima-frank/o/coderfrank?alt=media
Content-disposition: attachment
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Vary: Origin,X-Origin
Content-length: 0
Cache-control: no-cache, no-store, max-age=0, must-revalidate
Etag: CNerzaGOku8CEDU=
X-goog-generation: 1614705373435351
Pragma: no-cache
X-guploader-uploadid: ABg5-UzkhLZtBh_nh_MyJJUrxf2cptQxl5InjXd8jYRhUN31wS_DHiHBctrKCu7RLyi0q_5cZvAdHpj1D7nS9FsgsbI
Date: Mon, 15 Mar 2021 20:48:06 GMT
X-goog-hash: crc32c=AAAAAA==,md5=1B2M2Y8AsgTpgAmY7PhCfg==
Content-type: application/octet-stream
X-goog-storage-class: MULTI_REGIONAL
```

Example of a composed GCS Object response headers:
```
HTTP/1.1 200 OK
X-goog-metageneration: 1
Content-location: https://storage.googleapis.com/storage/v1/b/anima-frank/o/coderfrank-composed?alt=media
Content-disposition: attachment
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Vary: Origin,X-Origin
Content-length: 0
Cache-control: no-cache, no-store, max-age=0, must-revalidate
Etag: CNr8/deVs+8CEAE=
X-goog-generation: 1615841237892698
Pragma: no-cache
X-guploader-uploadid: ABg5-UzkOS0YokMzs44P_reufnM5l2idHXaVhKKbsjyQr1OzvgqZtOUIEkahxuFplJfkcylpFQ_4GyEq98JbWocEvt4
Date: Mon, 15 Mar 2021 20:48:59 GMT
X-goog-hash: crc32c=AAAAAA==
Content-type: application/octet-stream
X-goog-storage-class: MULTI_REGIONAL
```

frankyn (changes requested):

src/file.ts
stephenplusplus: I wouldn’t mind tackling that in a follow-up PR, unless you feel strongly it should be included here. This change shouldn’t alter the behavior we currently have, as far as how that situation is handled.

I will add a test for if just one algorithm comes back.

stephenplusplus: Test added.

frankyn (approved):
Thanks @stephenplusplus, LGTM!

Pending @shaffeeullah approval as well.

shaffeeullah (approved):

PR #1421, opened 2021-03-12
fix(perf): Dont always need credentials when signing

### LINKED ISSUE #1386: getSignedUrl slow when using Workload Identity on GKE
Our application needs to sign objects in relatively large bursts. A single request from our browser application could result in 100s of storage objects to be returned by our server, which should then all be signed concurrently.

We're observing the signing to be slow for large batches _when running our application in GKE_. In contrast, on our local dev machines, signing is fast for large batches.

A notable difference between our production env (GKE) and our local environments is that GKE is set up with Workload Identitity, while local development uses a local service account key file (JSON).

In some cases we've seen errors like these, but in most cases signing is just slow (~1 second):
```
http://169.254.169.254/computeMetadata/v1/instance/service-accounts/?recursive=true network timeout at:
```

I'm pretty sure that `169.254.169.254` is the local IP for the GKE metadata server.

So my current hypothesis is that the metadata daemon in GKE is either hitting an actual resource limit, or it's being throttled somewhere. I also believe that this `@google-cloud/storage` module is not caching credential responses from the metadata service.

This is quite unfortunate, since we recently migrated from using Service Account keyfiles in our GKE Secrets, to using Workload Identity as it seems to be the recommended mode of authenticating our workloads.

I wonder if there are intrinsic limitations to Workload Identity that prevent this library from caching the credential response. If so, would there be a way for us to take over the process of fetching the credentials using Workload Identity, so we can enforce our own caching policy here?

#### Environment details

  - OS: `FROM node:12.16.3-alpine3.11` (docker base image)
  - Node.js version: 12.16.3
  - npm version: n/a
  - `@google-cloud/storage` version: `5.5.0`

#### Steps to reproduce

  1. Spin up a GKE cluster with Workload Identity enabled
  2. Fire 100s of concurrent object signing requests; `storage.bucket('...').file('...').getSignedUrl(...)`

nhumrich:
When getting signed urls, the credentials are always loaded
only so that the client email can be used in the signing blob.
The Google Auth client caches the email on the first call, which never
changed. When using GCE/GKE for authentication, calling
`getCredentials()` makes a call to the metadata service.
This call is often not needed because the email is already known.
In the case of signing many files at once, this can really degrade the
metadata service and cause timeout errors.
This change simply uses the cached client email if it exists, before
making a remote call

Fixes #1386

### DIFF
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -277,8 +277,14 @@ export class URLSigner {
     const credentialScope = `${datestamp}/auto/storage/goog4_request`;
 
     const sign = async () => {
-      const credentials = await this.authClient.getCredentials();
-      const credential = `${credentials.client_email}/${credentialScope}`;
+      let client_email;
+      if ((this.authClient as any).jsonContent?.client_email == null) { // eslint-disable-line
+        const credentials = await this.authClient.getCredentials();
+        client_email = credentials.client_email;
+      } else {
+        client_email = (this.authClient as any).jsonContent.client_email; // eslint-disable-line
+      }
+      const credential = `${client_email}/${credentialScope}`;
       const dateISO = dateFormat.format(
         config.accessibleAt ? config.accessibleAt : new Date(),
         'YYYYMMDD[T]HHmmss[Z]',


src/signer.ts
```diff
@@ -277,8 +277,14 @@ export class URLSigner {
     const credentialScope = `${datestamp}/auto/storage/goog4_request`;
 
     const sign = async () => {
-      const credentials = await this.authClient.getCredentials();
-      const credential = `${credentials.client_email}/${credentialScope}`;
+      let client_email;
+      if ((this.authClient as any).jsonContent?.client_email == null) { // eslint-disable-line
```
stephenplusplus, line 281: I don’t believe we should access those internal properties from this library. “getCredentials()” should be returning the cached credentials, rather than re-fetching them each time. If it’s not, we should look into adding that behavior over there.

stephenplusplus (changes requested):

src/signer.ts
nhumrich: Yes, you are right. That is a better fix. Looking into this, is actually appears that this _was_ fixed in google auth library upstream. So all I needed to do was update the library

PR #1426, opened 2021-03-23
fix: update CopyOptions type to include cacheControl, contentType and contentEncoding

### LINKED ISSUE #1422: Update `CopyOptions` type to include cacheControl, contentType and contentEncoding
# Use Case
I'd like to copy a file to a destination bucket, and specify a new `cacheControl` metadata value in an atomic way. Currently, the only documented way to set the `cacheControl` metadata field is through the `File.setMetadata()` method, which results in a separate network call and is not atomic. 

# Current Behaviour
While not documented, it is currently possible to copy and update the cache control atomically by doing the following: `oldFile.copy(newFile, {cacheControl: CACHE_CONTROL} as any)`.  **Would it be possible to update the `CopyOptions` type to include `cacheControl`, `contentEncoding` and `contentType`? If I set one of the properties, the GCS service clobbers all the other  ones, so I'd need to explicitly specify them all in the `copy` call.** 

# stephenplusplus pr
I am aware of the existence of this PR: https://github.com/googleapis/nodejs-storage/pull/1406 . To clarify, the `metadata` field in `CopyOptions` does *not* alter the actual `cacheControl` field, but rather, some user defined metadata map on the Object. 

Screen Shot 2021-03-17 at 3 45 47 PM

# Other client libraries
It looks like the c++ client library does support this behaviour as well. It'd be nice for the nodejs library to achieve parity with the other client libraries.

Reference Links: 
https://googleapis.dev/cpp/google-cloud-storage/0.4.0/structgoogle_1_1cloud_1_1storage_1_1v0_1_1WithObjectMetadata.html
https://googleapis.dev/cpp/google-cloud-storage/0.4.0/classgoogle_1_1cloud_1_1storage_1_1v0_1_1Client.html#aeb5908b36999c1a3a357220e0f8e59a4

#### Steps to reproduce

  1.  Call `oldFile.copy(newFile, {cacheControl: CACHE_CONTROL} as any)`. This will update the object's cache control policy, even though it is not documented.

shaffeeullah:
As explained in https://github.com/googleapis/nodejs-storage/issues/1422, `File.copy` already allows users to update these fields. It was missing documentation. This adds that documentation by adding those fields to the type & adds a system test for it.

Fixes: https://github.com/googleapis/nodejs-storage/issues/1422 🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -304,6 +304,9 @@ export interface FileOptions {
 }
 
 export interface CopyOptions {
+  cacheControl?: string;
+  contentEncoding?: string;
+  contentType?: string;
   destinationKmsKeyName?: string;
   metadata?: Metadata;
   predefinedAcl?: string;
@@ -857,6 +860,9 @@ class File extends ServiceObject<File> {
    * @typedef {object} CopyOptions Configuration options for File#copy(). See an
    *     [Object
    * resource](https://cloud.google.com/storage/docs/json_api/v1/objects#resource).
+   * @property {string} [cacheControl] The cacheControl setting for the new file.
+   * @property {string} [contentEncoding] The contentEncoding setting for the new file.
+   * @property {string} [contentType] The contentType setting for the new file.
    * @property {string} [destinationKmsKeyName] Resource name of the Cloud
    *     KMS key, of the form
    *     `projects/my-project/locations/location/keyRings/my-kr/cryptoKeys/my-key`,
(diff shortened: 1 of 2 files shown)


JustinBeckwith (approved):
nit: since this impacts types, I would classify it as a `fix`.  We probably would want to include this in a changelog, because it has an impact on how the user would use the library.

PR #1428, opened 2021-03-23
fix: don't fail if ~/.config doesn't exist

### LINKED ISSUE #1420: Resumable Upload fails if ~/.config does not exist
#### Environment details

  - OS: Linux
  - Node.js version: 10.16.2
  - npm version: 6.9.0
  - `@google-cloud/storage` version: 5.8.0

#### Steps to reproduce

  1. Try to do a resumable upload in an environment (e.g. minimal container) where ~/.config does not exist
  2. It fails

This is a rehash of a portion of #909:
> Second, the check for write access to the configDir fails if the directory does not exist, leading to a ResumableUploadError.

This was never fixed.

Example error log (sanitized) from our application:

```
[2021-03-10T22:30:49.609Z] ERROR: xyzzy/GoogleStorage/27322 on xyzzy-0: destination stream error for file upload: xyzzy-0/2021-03-10/xyzzy-0_2021-03-10T22:29:33Z-xyzzy_0.xyzzy
    ResumableUploadError: A resumable upload could not be performed. The directory, /home/xyzzy/.config, is not writable. You may try another upload, this time setting `options.resumable` to `false`.
        at fs.access.err (/usr/lib/node_modules/@xyzzy/xyzzy/node_modules/@google-cloud/storage/build/src/file.js:1327:39)
        at FSReqWrap.args [as oncomplete] (fs.js:140:20)
```

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1420 🦕


src/file.ts
```diff
@@ -1802,7 +1802,9 @@ class File extends ServiceObject<File> {
       const configDir = xdgBasedir.config || os.tmpdir();
 
       fs.access(configDir, fs.constants.W_OK, err => {
-        if (err) {
+        if (err && fs.existsSync(configDir)) {
```
JustinBeckwith, line 1805: Generally you want to avoid calling `Sync` methods inside of production level code paths.  They block the node.js event loop, and file system access is slow.  If we wanted to keep down this path, I think we'd have to make 2 distinct calls to `fs.access` with varying flags:  one with `R_OK` and one with `W_OK`.  

That having been said ...   the pattern of checking for a given directory or file before starting a write is generally frowned upon.  This is specifically called out in the `fs` docs:

> Do not use `fs.access()` to check for the accessibility of a file before calling `fs.open()`, `fs.readFile()` or `fs.writeFile()`. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

From here:
https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback

Should we instead handle errors that come back directly from `startSimpleUpload_`, and give a good error message based on the resulting stream?

JustinBeckwith (changes requested):

src/file.ts
JustinBeckwith: Adding to the first point - if we decide to keep going down the pre-check route, it's possible you can do a single `fs.stat` and do bit math to play with the resulting mode - I'm out of my depth here though (@bcoe may know more)

src/file.ts
```diff
@@ -1802,25 +1804,32 @@ class File extends ServiceObject<File> {
...
-            return;
+        const maybeCreateFolder = async () => {
+          if (err) {
+            try {
+              await mkDirAsync(configDir, {mode: 0o0700});
+            } catch (mkDirErr) {
+              if (options.resumable) {
+                const error = new ResumableUploadError(
+                  [
+                    'A resumable upload could not be performed. The directory,',
```
stephenplusplus, line 1815: I think there are two causes of failures now:

- The directory doesn't exist, and can't be created
- The directory exists, but cannot be written to

Should we generalize the error message or maybe do a separate message for each?

src/file.ts
```diff
@@ -1802,25 +1804,32 @@ class File extends ServiceObject<File> {
...
-          // User didn't care, resumable or not. Fall back to simple upload.
-          this.startSimpleUpload_(fileWriteStream, options);
-          return;
-        }
-
-        this.startResumableUpload_(fileWriteStream, options);
+          this.startResumableUpload_(fileWriteStream, options);
+        };
+        maybeCreateFolder().catch(e => {
+          throw new Error(e);
```
shaffeeullah, line 1831: This line is not covered by code coverage, but I'm not sure how to get to this case. This error shouldn't get thrown. Any ideas for how to write a test for this are welcome.

stephenplusplus: Maybe related, do we need the function “maybeCreateFolder”? Can the code inside simply execute when the fs.access callback runs?

src/file.ts
shaffeeullah: Great callout! I've created a separate issue to track this [here](https://github.com/googleapis/nodejs-storage/issues/1431).

src/file.ts
shaffeeullah: Good point. Fixed!

frankyn (approved):

src/file.ts
```diff
@@ -1801,26 +1801,30 @@ class File extends ServiceObject<File> {
       // https://github.com/yeoman/configstore/blob/f09f067e50e6a636cfc648a6fc36a522062bd49d/index.js#L11
       const configDir = xdgBasedir.config || os.tmpdir();
 
-      fs.access(configDir, fs.constants.W_OK, err => {
```
stephenplusplus, line 1804: I attempted a bit of a code re-sort and added comments for clarity. Hopefully it's not too much, and it's helpful for future developers. Feel free to take only pieces if there are any parts that you think are useful:

```js
// The logic below attempts to mimic the resumable upload library,
// gcs-resumable-upload. That library requires a writable configuration
// directory in order to work. If we wait for that library to discover any
// issues, we've already started a resumable upload which is difficult to back
// out of. We want to catch any errors first, so we can choose a simple, non-
// resumable upload instead.

// Same as configstore (used by gcs-resumable-upload):
// https://github.com/yeoman/configstore/blob/f09f067e50e6a636cfc648a6fc36a522062bd49d/index.js#L11
const configDir = xdgBasedir.config || os.tmpdir();

fs.access(configDir, fs.constants.W_OK, accessErr => {
  if (!accessErr) {
    // A configuration directory exists, and it's writable. gcs-resumable-upload
    // should have everything it needs to work.
    this.startResumableUpload_(fileWriteStream, options);
    return;
  };

  // The configuration directory is either not writable, or it doesn't exist.
  // gcs-resumable-upload will attempt to create it for the user, but we'll try
  // it now to confirm that it won't have any issues. That way, if we catch the
  // issue before we start the resumable upload, we can instead start a simple
  // upload.
  fs.mkdir(configDir, {mode: 0o0700}, err => {
    if (!err) {
      // We successfully created a configuration directory that
      // gcs-resumable-upload will use.
      this.startResumableUpload_(fileWriteStream, options);
      return;
    }

    if (options.resumable) {
      // The user wanted a resumable upload, but we couldn't create a
      // configuration directory, which means gcs-resumable-upload will fail.
      const error = new ResumableUploadError(
        [
          'A resumable upload could not be performed. The directory,',
          `${configDir}, is not writable. You may try another upload,`,
          'this time setting `options.resumable` to `false`.',
        ].join(' ')
      );
      stream.destroy(error);
    } else {
      // The user didn't care, resumable or not. Fall back to simple upload.
      this.startSimpleUpload_(fileWriteStream, options);
    }
  });
});
```

test/file.ts
```diff
@@ -1892,6 +1902,28 @@ describe('File', () => {
...
 
+    describe('should succeed if resumable requested but config dir doesnt exist', () => {
+      const configDir = xdgBasedirCached.config;
+
+      before(() => {
+        rimraf.sync(configDir);
+      });
+
+      it('should succeed if resumable requested but config dir doesnt exist', done => {
+        assert.strictEqual(fs.existsSync(configDir), false);
```
shaffeeullah, line 1913: `fs.access(configDir, fs.constants.W_OK, accessErr => {` is not throwing an error for this test and I'm really confused as to why. @stephenplusplus any ideas?

stephenplusplus: I believe that was because we were stubbing `fs.access` earlier in the test file to always pass. Checking into this caused me to get in the editor and write a bit of code as I worked through the issue, and I ended up with a few tests to cover the changes in the PR. I'm sorry I keep doing this. It seems to work out that when I'm investigating an issue, I end up writing a bunch of code to test my theories out. So, here is what I ended up with: https://gist.github.com/stephenplusplus/dccf237542b7a702ebff90f03f2a557a. Don't worry about using any of it if you wanted to go a different way with the tests.

shaffeeullah: No need to apologize! Thanks so much for doing this!!! Super super helpful :) Really appreciate it

shaffeeullah: Good catch noting that `fs.access` was being stubbed. I totally missed that.

JustinBeckwith (approved):

package.json
```diff
@@ -111,6 +112,7 @@
     "node-fetch": "^2.2.0",
     "normalize-newline": "^3.0.0",
     "proxyquire": "^2.1.3",
+    "rimraf": "^3.0.2",
```
stephenplusplus, line 115: I think we can remove this dependency now.

stephenplusplus (changes requested):

stephenplusplus (approved):

stephenplusplus:
Thank you! LGTM!

PR #1461, opened 2021-05-04
docs: Notification create documentation was listed as exists

shaffeeullah:
Documentation for `notification.exists` is listed [twice](https://googleapis.dev/nodejs/storage/latest/Notification.html#exists). 
image

This is because the first one should be `notification.create`.


danielduhh (approved):
🙏🏿

frankyn (approved):

PR #1474, opened 2021-05-27
feat: customize retry behavior implementation

shaffeeullah:
Custom retry strategy behavior on the storage side. Will merge all storage retry PRs into the `shaffeeullah/retryStrategy` branch and then merge that branch into master at the end. This avoids one giant PR.

- Note: `autoRetry` and `maxRetries` implementation currently relies on `nodejs-common` defaults. This PR also pulls control over these defaults into the `nodejs-storage` library.


src/storage.ts
```diff
@@ -371,6 +374,13 @@ export class Storage extends Service {
    * @property {boolean} [autoRetry=true] Automatically retry requests if the
    *     response is related to rate limits or certain intermittent server
    * errors. We will exponentially backoff subsequent requests by default.
+   * @property {number} [backoffMultiplier = 2] Dictates the rate of exponential backoff.
```
tritone, line 377: I think it might be clearer if you add a little explanation/example about how these parameters interact with one another. For example with the default settings, up to how many retries will happen and how much will they be spaced out? You could also link to something explaining backoff if that's easier.

src/storage.ts
```diff
@@ -371,6 +374,13 @@ export class Storage extends Service {
...
+   * @property {number} [retryDeadline = 600] The length of time to keep retrying in seconds.
+   *     The last sleep period will be shortened as necessary, so that the last retry runs
+   *     at deadline (and not considerably beyond it).
+   * @property {number} [maxRetryTimeout = 64] The maximum time to delay in seconds.
```
tritone, line 381: I wouldn't call this a timeout exactly I think-- I've heard the term "envelope" or "window" for this (spacing between retries).

src/storage.ts
```diff
@@ -46,6 +46,9 @@ export interface CreateBucketQuery {
 }
 
 export interface StorageOptions extends ServiceOptions {
+  backoffMultiplier?: number;
```
tritone, line 49: Have we considered factoring the retry options into some sort of sub-config? Not sure if this is a thing in Node. In Go for example the various mathy configs are all in https://pkg.go.dev/github.com/googleapis/gax-go/v2#Backoff

src/storage.ts
shaffeeullah: changed this to `totalTimeout` to match the retry customization variable of the google gax library

tritone: I was confused but it looks like you actually used `MaxRetryDelay` from there for this param, not `TotalTimeout`

shaffeeullah: whoops yeah my previous naming was super confusing. does it look right to you now?

tritone: Yep much better!

src/storage.ts
```diff
@@ -393,6 +405,12 @@ export class Storage extends Service {
    * @param {StorageOptions} [options] Configuration options.
    */
   constructor(options: StorageOptions = {}) {
+    const AUTO_RETRY_DEFAULT = true;
+    const MAX_RETRY_DEFAULT = 3;
```
tritone, line 409: Was this constant being used already? 3 seems way too low.

src/storage.ts
shaffeeullah: Yay thanks for the feedback!

src/storage.ts
shaffeeullah: Yeah this is the current default

src/storage.ts
```diff
@@ -393,6 +420,12 @@ export class Storage extends Service {
    * @param {StorageOptions} [options] Configuration options.
    */
   constructor(options: StorageOptions = {}) {
+    const AUTO_RETRY_DEFAULT = true;
```
ddelgrosso1, line 423: Would it make sense to move these constants to the class level along with a comment for consistency with other files? For example https://github.com/googleapis/nodejs-storage/blob/fbb90c2b5e01c971c1aa0bbc54ddabb2df89673d/src/bucket.ts#L381

shaffeeullah: fixed! good callout

test/index.ts
```diff
@@ -176,7 +176,87 @@ describe('Storage', () => {
...
+        retryOptions: {autoRetry},
+      });
+      const calledWith = storage.calledWith_[0];
+      assert.strictEqual(calledWith.retryOptions.autoRetry, autoRetry);
+    });
+
+    it('should throw if autoRetry is defined twice', () => {
+      const autoRetry = 10;
+      assert.throws(() => {
+        const storage = new Storage({
```
ddelgrosso1, line 195: Since this is testing that the constructor throws an exception could the assignment to the const be removed to cleanup the linter warning? Same goes for L285.

shaffeeullah: whoops, totally missed the linter warnings. fixed

ddelgrosso1 (approved):
LGTM

PR #1476, opened 2021-05-27
feat: implemented retry function

shaffeeullah:


src/storage.ts
```diff
@@ -413,10 +414,42 @@ export class Storage extends Service {
...
+          return true;
+        }
+
+        if (err.errors) {
+          for (const e of err.errors) {
+            const reason = e.reason;
+            if (
+              reason === 'rateLimitExceeded' ||
+              reason === 'userRateLimitExceeded' ||
+              (reason && reason.includes('EAI_AGAIN')) ||
```
shaffeeullah, line 436: @frankyn The 3 above errors were already being retried via common, but are not listed in the retry guidance document. Should they continue to be retried? Or should I remove these?

src/storage.ts
```diff
@@ -413,10 +414,42 @@ export class Storage extends Service {
...
+              reason === 'Connection Reset By Peer' ||
+              reason === 'Unexpected Connection Closure'
```
shaffeeullah, line 438: @frankyn What would be a good way to force these failures to see if this is the correct string to be looking for?

frankyn: Emulator has Reset connection and Unexpected connection closure support. We can talk through this in our meeting later today.

src/storage.ts
frankyn: Chime in your doc

src/storage.ts
```diff
@@ -413,10 +418,42 @@ export class Storage extends Service {
...
     const baseUrl = EMULATOR_HOST || `${options.apiEndpoint}/storage/v1`;
 
+    /**
+     * Returns true if the API request should be retried, given the error that was
+     * given the first time the request was attempted.
+     *
+     * @param {error} err - The API error to check if it is appropriate to retry.
+     * @return {boolean} True if the API request should be retried, false otherwise.
+     */
+    const retryFunction = function (err?: ApiError) {
```
shaffeeullah, line 428: TODO: make this configurable

JustinBeckwith (commented):
would like us to be super careful here.  A few thoughts!

There is already logic to retry requests built into nodejs-common, where most of these calls go through.  It's here:  https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L513

Can we use a single implementation?

I also want to dig in on consistency across clients.  If this is implemented one way in Storage, and another in PubSub, we're going to run into trouble.  Is there a design doc folks who do nodejs can jointly work on and chat about this?

JustinBeckwith:
@shaffeeullah and I took this offline, and I am onboard and happy 👍

src/storage.ts
```diff
@@ -413,10 +418,42 @@ export class Storage extends Service {
...
+      if (err) {
+        if ([408, 429, 500, 502, 503, 504].indexOf(err.code!) !== -1) {
+          return true;
+        }
+
+        if (err.errors) {
+          for (const e of err.errors) {
+            const reason = e.reason;
+            if (
+              reason === 'rateLimitExceeded' ||
```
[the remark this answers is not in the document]
shaffeeullah: This will be removed. Spoke with frank about this offline, this error won't be thrown by GCS

src/storage.ts
```diff
@@ -218,6 +218,36 @@ const TOTAL_TIMEOUT_DEFAULT = 600;
...
+ * @param {error} err - The API error to check if it is appropriate to retry.
+ * @return {boolean} True if the API request should be retried, false otherwise.
+ */
+const RETRYABLE_ERR_FN_DEFAULT = function (err?: ApiError) {
+  if (err) {
+    if ([408, 429, 500, 502, 503, 504].indexOf(err.code!) !== -1) {
+      return true;
+    }
+
+    if (err.errors) {
```
ddelgrosso1, line 235: Is there a situation where ```err !== undefined``` but ```err.errors === undefined``` (or really anything non-enumerable)?

shaffeeullah: I dont believe so. If there were, the bug would have already surfaced [here](https://github.com/googleapis/nodejs-common/blob/56764dc58634d151f27b67869d38a7ef97708b6c/src/util.ts#L513)

ddelgrosso1: I think it would be safe to remove this ```if``` statement in that case.

shaffeeullah: Sorry I misread your original question. It is possible for err.errors to be undefined. For example, in this test it is undefined. https://github.com/googleapis/nodejs-storage/blob/cec5e15780a0fb726247f4c281391157bdaa1cda/test/index.ts#L307

ddelgrosso1: I see it now: https://github.com/googleapis/nodejs-common/blob/4a1b1ab2bdedb92e4c5b102f04b005c252887c63/src/util.ts#L228. Thanks for clarifying!

ddelgrosso1 (approved):
LGTM

PR #1493, opened 2021-06-23
feat: customize retry behavior implementation (#1474)

shaffeeullah:
Retry customization implementation. 

Adds the RetryOptions interface which allows users to configure retryDelayMultiplier, totalTimeout, maxRetryDelay, autoRetry, maxRetries, and retryableErrorFn


ddelgrosso1 (approved):
LGTM

src/storage.ts
```diff
@@ -163,6 +177,77 @@ export type GetHmacKeysResponse = [HmacKey[]];
...
+  if (err) {
+    if ([408, 429, 500, 502, 503, 504].indexOf(err.code!) !== -1) {
+      return true;
+    }
+
+    if (err.errors) {
+      for (const e of err.errors) {
+        const reason = e.reason?.toLowerCase();
+        if (
+          (reason && reason.includes('eai_again')) ||
```
danielduhh, line 239: What does this `eai_again` mean? Maybe add a comment?

src/storage.ts
```diff
@@ -163,6 +177,77 @@ export type GetHmacKeysResponse = [HmacKey[]];
...
+          reason === 'connection reset by peer' ||
```
danielduhh, line 240: Is it same to assume the text here won't change?

shaffeeullah: There's nothing stopping the GCS team from changing it. Similar logic has been in the released version of nodejs-common (https://github.com/googleapis/nodejs-common/blob/513374ca1bce94ba720106ab78c776a0e61c0358/src/util.ts#L549) and I'm not sure of another way to check for this error. This logic was designed to be similar to common.

src/storage.ts
shaffeeullah: It's a DNS lookup error. Added a comment.

danielduhh (approved):
🚢

PR #1503, opened 2021-07-13
feat: remove gaxios dependency

shaffeeullah:
Removes dependency on gaxios by refactoring `File.isPublic`. 

This 
1. Allows the library to capitalize on the existing retry logic in `nodejs-common`
2. Removes a dependency
3. Makes the `isPublic` function consistent with most other functions in the library

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1256,10 +1256,14 @@ class File extends ServiceObject<File> {
         query.userProject = options.userProject;
       }
 
+      interface Headers {
+        [index: string]: any;
+      }
+
       const headers = {
         'Accept-Encoding': 'gzip',
         'Cache-Control': 'no-store',
-      } as any;
+      } as Headers;
 
       if (rangeRequest) {
         const start = typeof options.start === 'number' ? options.start : '0';
--- a/src/file.ts
+++ b/src/file.ts
@@ -3016,7 +3016,8 @@ class File extends ServiceObject<File> {
         } else {
           callback!(null, true);
         }
-      });
+      }
+    );
   }
 
   makePrivate(
--- a/src/file.ts
+++ b/src/file.ts
@@ -1257,7 +1257,7 @@ class File extends ServiceObject<File> {
       }
 
       interface Headers {
-        [index: string]: any;
+        [index: string]: string;
       }
 
       const headers = {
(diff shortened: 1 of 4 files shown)


src/file.ts
```diff
@@ -2996,20 +2995,29 @@ class File extends ServiceObject<File> {
...
-    gaxiosRequest({
-      method: 'HEAD',
-      url: `http://${
-        this.bucket.name
-      }.storage.googleapis.com/${encodeURIComponent(this.name)}`,
-    }).then(
-      () => callback!(null, true),
-      (err: GaxiosError) => {
-        if (err.code === '403') {
+    const callbackFunction = function (err: Error | ApiError | null) {
```
ddelgrosso1, line 2998: From a style standpoint it doesn't appear we define functions in this manner anywhere else in this file (and others). I would suggest we either define it as an arrow function or move it inline to util.makeRequest as an arrow function.

src/file.ts
```diff
@@ -1260,7 +1259,7 @@ class File extends ServiceObject<File> {
       const headers = {
         'Accept-Encoding': 'gzip',
         'Cache-Control': 'no-store',
-      } as Headers;
+      } as any;
```
ddelgrosso1, line 1262: Maybe we should define an interface / type instead of using any here to avoid adding additional linter warnings.

src/file.ts
```diff
@@ -1257,6 +1256,10 @@ class File extends ServiceObject<File> {
         query.userProject = options.userProject;
       }
 
+      interface Headers {
+        [index: string]: any;
```
ddelgrosso1, line 1260: Could this be `[index: string]: string;`?

JustinBeckwith (commented):
This isn't really going to do much for you.  `gaxios` is a required dependency of `google-auth-library`, and removing it here doesn't actually remove it from the transitive dependency graph.  We want to move *away* from the `request` style libraries, not closer towards them.

shaffeeullah:
> This isn't really going to do much for you. `gaxios` is a required dependency of `google-auth-library`, and removing it here doesn't actually remove it from the transitive dependency graph. We want to move _away_ from the `request` style libraries, not closer towards them.

Good to know! We'd like to move over the isPublic function to common anyway to be consistent with the rest of the library and also to leverage the retry functionality we recently implemented in common.

src/file.ts
shaffeeullah: Ah yes. I just copied this from the Gaxios file. Making that change now

ddelgrosso1 (approved):

PR #1504, opened 2021-07-14
feat: applied customization to multipart filesave

shaffeeullah:

### DIFF
--- a/test/file.ts
+++ b/test/file.ts
@@ -242,8 +242,8 @@ describe('File', () => {
         maxRetryDelay: 60,
         retryableErrorFn: (err: HTTPError) => {
           return err.code === 500;
-        }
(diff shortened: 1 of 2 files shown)


test/file.ts
```diff
@@ -48,6 +48,7 @@ import {
   GetSignedUrlConfig,
   GenerateSignedPostPolicyV2Options,
   GenerateSignedPostPolicyV2Callback,
+  Storage,
```
ddelgrosso1, line 51: Is this import needed?

shaffeeullah: Good catch. It must've autoimported when I was testing something and I missed it

ddelgrosso1 (approved):
LGTM

PR #1511, opened 2021-07-22
feat: retries for resumable bucket.upload and file.save

ddelgrosso1:

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -3737,7 +3737,7 @@ class Bucket extends ServiceObject {
               .pipe(writable)
               .on('error', err => {
                 if (
-                  isMultipart &&
+                  (isMultipart || options.resumable) &&
                   this.storage.retryOptions.autoRetry &&
                   this.storage.retryOptions.retryableErrorFn!(err)
                 ) {
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -3725,7 +3725,6 @@ class Bucket extends ServiceObject {
     callback?: UploadCallback
   ): Promise<UploadResponse> | void {
     const upload = () => {
-      const isMultipart = options.resumable === false;
       const returnValue = retry(
         async (bail: (err: Error) => void) => {
           await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


src/bucket.ts
```diff
@@ -3737,7 +3737,7 @@ class Bucket extends ServiceObject {
               .pipe(writable)
               .on('error', err => {
                 if (
-                  isMultipart &&
+                  (isMultipart || options.resumable) &&
```
shaffeeullah, line 3740: This will always evaluate to true, right? Because each upload is either multipart or resumable. I think this should be checking for (isMultipart || err.message) or something like that. We want to retry all multipart uploads and then resumable uploads only where it is a certain error (because otherwise gcs-resumable-upload will handle it)

test/file.ts
```diff
@@ -4118,7 +4118,7 @@ describe('File', () => {
           await file.save(DATA, options);
           throw Error('unreachable');
         } catch (e) {
-          assert.strictEqual(e.message, 'first error');
+          assert.strictEqual(e.message, 'unreachable');
```
shaffeeullah, line 4121: We shouldn't test this like this. If we're checking that it retries, we shouldn't throw the "unreachable" error (because it will be reachable) and instead assert the retryCount

src/bucket.ts
ddelgrosso1: In that case maybe we don't need this check at all. Certain errors are being explicitly handled down at the gcs-resumable-upload level [here](https://github.com/googleapis/gcs-resumable-upload/blob/362a65aa0d067c540faae922796aaee8d33258e6/src/index.ts#L654). This basically tells Gaxios to let gcs-resumable-upload handle everything. However, during the URL [creation](https://github.com/googleapis/gcs-resumable-upload/blob/362a65aa0d067c540faae922796aaee8d33258e6/src/index.ts#L626) only certain codes are handled at this the same level and the rest bubble up. I think we should handle anything that bubbles up and meets our retryableFn criteria at this level. What do you think?

shaffeeullah: If we try to retry a 500 error downstream and it never works out, it will be bubbled up here and then we will send it back downstream. This creates exponentially more retries than `maxRetries`.

ddelgrosso1: It actually doesn't cause an exponential increase in retries because the error that occurs when retries are exhausted is not being checked for in the retry function (it only has an error message and no error code). As a result the code at this level does not send it back downstream. Everything appears to get handled correctly. I will circle up with you to show you what I mean in case I missed anything obvious.

tritone (approved):

shaffeeullah (approved):

PR #1514, opened 2021-07-28
fix: don't retry create HMAC key

shaffeeullah:
Aligns behavior with guidance posted here: https://cloud.google.com/storage/docs/retry-strategy

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -30,8 +30,7 @@ import {Bucket} from '../src';
 import {GetFilesOptions} from '../src/bucket';
 import sinon = require('sinon');
 import {HmacKey} from '../src/hmacKey';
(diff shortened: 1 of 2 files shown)


test/index.ts
```diff
@@ -601,21 +610,8 @@ describe('Storage', () => {
       );
     });
 
-    it('should make request with method options as query parameter', async () => {
```
shaffeeullah, line 604: the logic in this test is duplicated by the logic in the "should make correct API request" test, so this test can be safely removed.

ddelgrosso1 (approved):

shaffeeullah:
This logic can be found here https://github.com/googleapis/nodejs-storage/pull/1517

PR #1517, opened 2021-08-02
fix: don't retry non-idempotent functions

shaffeeullah:
Don't retry createHmacKey and createNotification since they're not idempotent

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -950,6 +950,7 @@ export class Storage extends Service {
         method: 'POST',
         uri: `/projects/${projectId}/hmacKeys`,
         qs: query,
+        maxRetries: 0,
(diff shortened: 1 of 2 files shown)


src/storage.ts
```diff
@@ -950,6 +950,7 @@ export class Storage extends Service {
         method: 'POST',
         uri: `/projects/${projectId}/hmacKeys`,
         qs: query,
+        maxRetries: 0,
```
tritone, line 953: Worth adding a comment noting that this is on purpose because of idempotency?

tritone (commented):
Nice approach!

ddelgrosso1 (approved):

PR #1524, opened 2021-08-06
docs: add missing arguments to StorageOptions documentation

ddelgrosso1:

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -485,7 +485,7 @@ export class Storage extends Service {
    * @property {string} [userAgent] The value to be prepended to the User-Agent
    *     header in API requests.
    * @property {object} [authClient] GoogleAuth client to reuse instead of creating a new one.
-   * @property {number} [timeout] The amount of time in milliseconds to wait for a http request before timing out.
+   * @property {number} [timeout] The amount of time in milliseconds to wait per http request before timing out.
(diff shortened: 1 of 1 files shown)


src/storage.ts
```diff
@@ -484,6 +484,10 @@ export class Storage extends Service {
    *     error should be retried and false otherwise.
    * @property {string} [userAgent] The value to be prepended to the User-Agent
    *     header in API requests.
+   * @property {object} [authClient] GoogleAuth client to reuse instead of creating a new one.
+   * @property {number} [timeout] The amount of time in milliseconds to wait for a http request before timing out.
```
tritone, line 488: Is this the total timeout (covering all potential retries in a library call) or per-request?

shaffeeullah:
Where are these parameters actually used? I don't see any tests verifying that they work as expected.

src/storage.ts
shaffeeullah: No, this is different.  `totalTimeout` is in `retryOptions`.

ddelgrosso1: Per request, I'll add that caveat to the sentence.

ddelgrosso1:
@shaffeeullah the ones I added are passed into common and primarily used at that level.

shaffeeullah (approved):

PR #1531, opened 2021-08-09
samples: added create bucket notification sample

shaffeeullah:


shaffeeullah:
The test is failing because the service account does not have permission to execute this operation. I am working on resolving this (because I don't have permissions to change the permissions of the service account). 

@bcoe will need your help again here. Docs test is failing bc of OwlBot commit that adds a link off master for the samples included in this PR.

PR #1533, opened 2021-08-10
samples: implemented streaming file download sample

shaffeeullah:
Implemented streaming download sample based on normal download sample: https://github.com/googleapis/nodejs-storage/blob/master/samples/downloadFile.js

This sample is for [this page]( https://cloud.google.com/storage/docs/streaming#storage-stream-download-object-nodejs).

Docs check will fail because of OwlBot commit linking to the sample that exists in this PR. @bcoe will use his admin privileges to merge this PR (thanks, Ben!!)


samples/streamFileDownload.js
```diff
@@ -0,0 +1,68 @@
...
+  const storage = new Storage();
+
+  async function streamFileDownload() {
+    // The example below demonstrates how we can reference a remote file, then
+    // pipe its contents to a local file. This is effectively creating a local
+    // backup of your remote data.
+    await storage
+      .bucket(bucketName)
+      .file(fileName)
+      .createReadStream()
```
tritone, line 57: I would maybe separate out a variable for the stream created at this step, and add a comment that you can do whatever you want to do with the stream (not just pipe it into a file necessarily).

shaffeeullah: Do you think that might add too much complexity to the sample? There is tons of nodejs documentation about using streaming and things to do with streams. This sample just aims to show how to use streams with storage.

tritone: I think the main difference between a streaming download and a standard download is simply the fact that that the stream is something that can be used in different ways, right? Otherwise this will just end up looking identical to the standard download sample in some languages (e.g. Go).

tritone: The docs page says "Cloud Storage supports streaming transfers, which allow you to stream data to and from your Cloud Storage account without requiring that the data first be saved to a file."

This makes me think it might be better to have the sample stream the contents of the file to stdout or something like that.

shaffeeullah: I added the 'finish' function so it's a little different from the normal download sample. it now shows a stream function and how it can be used. i'm slightly opposed to separating out the stream into its own variable because that's not how we have any of the stream examples currently in our repo. This sample is consistent with all our other documentation.

ddelgrosso1 (approved):

rebecca-pete:
https://github.com/googleapis/nodejs-storage/blob/1272448fbf5918abd643293739b6b32b52a93878/samples/streamFileDownload.js#L41

Suggest replacing text with:

"The filename and file path where you want to download the file"

bcoe (approved):

PR #1534, opened 2021-08-10
samples: add sample for streaming file upload

ddelgrosso1:


shaffeeullah:
@ddelgrosso1 I don't see the actual sample file here.

samples/streamFileUpload.js
```diff
@@ -0,0 +1,73 @@
...
+  // Get a reference to the bucket
+  const myBucket = storage.bucket(bucketName);
+
+  // Create a reference to a file object
+  const file = myBucket.file(destFileName);
+
+  async function streamFileUpload() {
+    fs.createReadStream(filePath)
+      .pipe(file.createWriteStream())
+      .on('error', () => {
```
shaffeeullah, line 59: I don't think we should catch errors like this in the sample. What do you think? In this case it will fail silently. Maybe just remove this clause altogether?

shaffeeullah (approved):

samples/streamFileUpload.js
```diff
@@ -0,0 +1,70 @@
...
+  const fs = require('fs');
+
+  // Creates a client
+  const storage = new Storage();
+
+  // Get a reference to the bucket
+  const myBucket = storage.bucket(bucketName);
+
+  // Create a reference to a file object
+  const file = myBucket.file(destFileName);
```
tritone, line 54: Similarly to my comment on the download sample-- I think the point of this page is that you don't have to use a file for streaming. Maybe we can create a generic stream and show how it could be used to upload from in-memory buffer, or user input, etc? I think this would be more useful and interesting.

tritone (approved):

samples/streamFileUpload.js
ddelgrosso1: If I understand the concern, instead of reading from a file the preference would be to read from just any other type of stream? What about if the example showed creating a readable stream from a string and used that to stream to storage?

shaffeeullah: While I agree that showing those things are cool, I think reading a file is a core use case and should be the case shown in the main sample. If customers want other examples, we have those available in our documentation https://googleapis.dev/nodejs/storage/latest/File.html#createWriteStream

shaffeeullah: The c++ sample that already exists uploads from text https://cloud.google.com/storage/docs/streaming#storage-stream-upload-object-cpp
Maybe if we don't use a file we can use text instead to match c++

tritone: Discussed with @shaffeeullah offline-- I think copying C++ and writing from a readable stream from a string is fine. Let's make sure we have a comment to make it clear that the writes to the stream will continue to be sent until it is closed.

bcoe (approved):

PR #1543, opened 2021-08-19
feat: handle retries for conditionally idempotent bucket ops

ddelgrosso1:

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -57,7 +57,6 @@ import {
   URLSigner,
   Query,
 } from './signer';
-import {type} from 'os';
 
 interface SourceObject {
   name: string;
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -264,7 +264,7 @@ export enum BucketActionToHTTPMethod {
   list = 'GET',
 }
 
-export enum Methods {
+export enum AvailableServiceObjectMethods {
   setMetadata,
   delete,
 }
@@ -621,7 +621,7 @@ class Bucket extends ServiceObject {
   getFilesStream: Function;
   signer?: URLSigner;
 
-  originalRetryValue?: boolean;
+  instanceRetryValue?: boolean;
 
   constructor(storage: Storage, name: string, options?: BucketOptions) {
     options = options || {};
@@ -1016,7 +1016,7 @@ class Bucket extends ServiceObject {
 
     this.getFilesStream = paginator.streamify('getFiles');
 
-    this.originalRetryValue = storage.retryOptions.autoRetry;
+    this.instanceRetryValue = storage.retryOptions.autoRetry;
   }
 
   addLifecycleRule(
@@ -1244,16 +1244,12 @@ class Bucket extends ServiceObject {
 
     this.disableAutoRetryConditionallyIdempotent_(
       this.methods.setMetadata,
-      Methods.setMetadata
+      AvailableServiceObjectMethods.setMetadata
     );
 
     if (options.append === false) {
-      this.setMetadata(
-        {lifecycle: {rule: newLifecycleRules}},
-        callback
-      ).finally(() => {
-        this.storage.retryOptions.autoRetry = this.originalRetryValue;
-      });
+      this.setMetadata({lifecycle: {rule: newLifecycleRules}}, callback);
+      this.storage.retryOptions.autoRetry = this.instanceRetryValue;
       return;
     }
 
@@ -1276,10 +1272,9 @@ class Bucket extends ServiceObject {
           },
         },
         callback!
-      ).finally(() => {
-        this.storage.retryOptions.autoRetry = this.originalRetryValue;
-      });
+      );
     });
+    this.storage.retryOptions.autoRetry = this.instanceRetryValue;
   }
 
   combine(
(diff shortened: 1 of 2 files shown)


src/bucket.ts
```diff
@@ -616,6 +621,8 @@ class Bucket extends ServiceObject {
   getFilesStream: Function;
   signer?: URLSigner;
 
+  originalRetryValue?: boolean;
```
shaffeeullah, line 624: Maybe rename this to `instanceRetryValue`? Makes it more apparent that we want to override the value set at the Storage constructor level.

src/bucket.ts
```diff
@@ -264,6 +264,11 @@ export enum BucketActionToHTTPMethod {
   list = 'GET',
 }
 
+export enum Methods {
```
shaffeeullah, line 267: The name `Methods` seems vague/confusing to me if I was picking this up knowing nothing, but I'm not sure what a better option would be.

src/bucket.ts
```diff
@@ -3951,6 +4051,25 @@ class Bucket extends ServiceObject {
   getId(): string {
     return this.id!;
   }
+
+  disableAutoRetryConditionallyIdempotent_(
+    coreOpts: any,
```
shaffeeullah, line 4056: Is `any` the best option here? Can it be object?

src/bucket.ts
```diff
@@ -3951,6 +4051,25 @@ class Bucket extends ServiceObject {
   getId(): string {
     return this.id!;
   }
+
+  disableAutoRetryConditionallyIdempotent_(
```
shaffeeullah, line 4055: This function is really cool. Thanks for doing this!

src/bucket.ts
```diff
@@ -3814,6 +3908,12 @@ class Bucket extends ServiceObject {
       options
     );
 
+    // Do not retry if precondition option ifMetagenerationMatch is not set
+    let numberOfRetries = this.storage.retryOptions.maxRetries;
```
shaffeeullah, line 3912: Do we want to name this `maxRetries` for consistency? Or is it confusing to use the same name again? I guess it's not hard to understand, but sometimes fewer than this number of retries will occur if it succeeds on the first retry (for example)

JustinBeckwith (commented):
Could we add tests?

src/bucket.ts
```diff
@@ -1233,8 +1242,18 @@ class Bucket extends ServiceObject {
       return apiFormattedRule;
     });
 
+    this.disableAutoRetryConditionallyIdempotent_(
```
tritone, line 1245: I'm not any kind of node expert but it looks like you are disabling the retry during the setMetadata operation and then reenabling it afterwards. If so, does that pose any kind of concurrency/race condition issues with the client? Also, will user configuration choices override this logic?

tritone (approved):

src/bucket.ts
ddelgrosso1: There shouldn't be an issue with concurrency but I will make sure there isn't the possibility of a race condition. This is overriding the users configuration temporarily so actually the opposite of your second question.

src/bucket.ts
ddelgrosso1: Object won't work, I suppose I could mimic / clone the type from common if that is preferred?

src/bucket.ts
tritone: Gotcha. Does that mean that users can't configure the library to retry in spite of the operation being non-idempotent?

ddelgrosso1: The user can still configure the library to retry. However, for these conditionally idempotent operations if the correct precondition is not set, the library will override the user setting and turn off retries which is what the [spec](https://cloud.google.com/storage/docs/retry-strategy) notes.

src/bucket.ts
shaffeeullah: I think either that or disabling linting on this line. Object from common is what we've done in the past, so might be a good idea. It creates duplicate code though :( 

Your call on what you think is best.

test/bucket.ts
```diff
@@ -1532,6 +1581,16 @@ describe('Bucket', () => {
         done();
       });
     });
+
+    it('should disable autoRetry when ifMetagenerationMatch is undefined', done => {
+      bucket.setMetadata = () => {
+        assert.strictEqual(bucket.storage.retryOptions.autoRetry, false);
+      };
+
+      bucket.enableLogging({prefix: PREFIX}, (err: Error | null) => {
```
shaffeeullah, line 1590: can you remove `err` here to get rid of the linter warning?

test/bucket.ts
```diff
@@ -2131,6 +2199,14 @@ describe('Bucket', () => {
         done();
       });
     });
+
+    it('should disable autoRetry when ifMetagenerationMatch is undefined', done => {
+      bucket.setMetadata = (metadata: {}, options_: SetFileMetadataOptions) => {
```
shaffeeullah, line 2204: linter warnings also on `metadata` and `options`

test/bucket.ts
```diff
@@ -2553,6 +2681,18 @@ describe('Bucket', () => {
       });
     });
 
+    it('should disable autoRetry when ifMetagenerationMatch is undefined', done => {
+      const fakeFile = new FakeFile(bucket, 'file-name');
+      fakeFile.isSameFile = () => {
+        return true;
+      };
+      const options = {destination: fakeFile, metadata};
+      bucket.upload(filepath, options, (err: Error, file: FakeFile) => {
```
shaffeeullah, line 2690: linter warnings on `err` and `file`

shaffeeullah (approved):
Thanks, Denis!!

PR #1567, opened 2021-08-26
WIP: retry strategy runner code

shaffeeullah:
Not close to being done/working.


conformance-test/retryStrategy.ts
```diff
@@ -0,0 +1,151 @@
...
+    describe(`Scenario ${testCase.id}`, () => {
+      retryTestCases.retryCases.forEach((instructionSet: RetryCase) => {
+        const instructions = instructionSet.instructions;
+        //TODO set emulator based on instructions
+        testCase.methods.forEach(jsonMethod => {
+          const jsonMethodName = jsonMethod.name;
+          const jsonMethodResources = jsonMethod.resources;
+          const functionList = methodMap.get(jsonMethodName);
+          functionList?.forEach(storageMethod => {
+            function generateName(bucketOrFile: string) {
```
ddelgrosso1, line 91: If this scenario starts to get long it might be worth moving this function outside instead of nesting it in here. Also, maybe use template strings for the return to be consistent with other places.

conformance-test/retryStrategy.ts
```diff
@@ -0,0 +1,151 @@
...
+            });
+
+            it(`${storageMethod}`, async () => {
+              //if there are multiple cases, we're going to run into duplicate names here
+              const result = storageMethod(bucket);
+              // based on expectSuccess, make sure the right thing happens
+            });
+
+            after(() => {
+              return Promise.all([deleteAllBucketsAsync()]);
```
ddelgrosso1, line 114: Is there a need to rewrap the return from `deleteAllBucketsAsync()` here?

conformance-test/test-data/retryStrategyTestData.json
```diff
@@ -0,0 +1,262 @@
...
+          {
+            "instructions": []
+          }
+        ],
+        "methods": [],
+        "preconditionProvided": false,
+        "expectSuccess": false
+      }
+    ]
+  }
```
ddelgrosso1, line 262: Probably get caught by the linter but this file is missing a new line.

shaffeeullah: `v4SignedUrl.json` also doesn't have a new line. I think this is fine?

shaffeeullah: or is new line the correct thing to do

shaffeeullah:
@ddelgrosso1 Still WIP. Needs to be refactored for sure. I want to discuss the overall structure with you in our next sync before doing more. I'm not sure how I'm supposed to be using RESOURCES with the way that we structured the method mapping (and the way that I wrote the methods). Let's pair on this in our next sync.

ddelgrosso1:
@shaffeeullah sounds good to me. If I have some time today I will take a look again.

conformance-test/retryStrategy.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+});
+
+function excecuteScenario(testCase: RetryTestCase) {
+  testCase.retryCases.forEach((instructionSet: RetryCase) => {
+    configureTestBench(instructionSet.instructions);
+    testCase.methods.forEach(jsonMethod => {
+      const functionList = methodMap.get(jsonMethod.name);
+      functionList?.forEach(storageMethodString => {
+        const storageMethodObject = (global as any).storageMethodString;
```
ddelgrosso1, line 102: If `storageMethodString` is coming in as string type you may have to use bracket notation here instead.

shaffeeullah: using bracket notation throws errors: "Type string cannot be used as an index type"
I will run the code and debug it from there.

ddelgrosso1: I would just make sure that using a `.` works with a string. Maybe TypeScript handles it cleaner than vanilla JS.

ddelgrosso1: Actually, I think it might be because `global` is of type `any` and not `object`. If you have a way to test, I would just make sure dot notation works as expected and leave it that way if so.

shaffeeullah: I was thinking of testing it after we merge this PR? i'm sure i'll encounter other bugs along the way.

ddelgrosso1:
Structurally I think this is in a good spot to go into the staging branch. Perhaps maybe just fix up the linter problems (at least the error)?

shaffeeullah:
> Structurally, I think this is in a good spot to go into the staging branch. Perhaps maybe just fix up the linter problems (at least the error)?

Theoretically fixed the error. Waiting for linting to finish running. All the warnings saying that a function isn't used will continue to exist because the functions are used via variables. I will manually add an ignore to all of those. Let me know if you think this is the incorrect way to handle this situation.

ddelgrosso1 (approved):

PR #1569, opened 2021-08-27
feat: add method mapping json file

ddelgrosso1:


shaffeeullah (approved):
should `deleteResumableCache` be here? I just realized that it's still commented out in my libraryMethods file. Otherwise, looks good. I didn't check to see that the number of functions here matches the number of functions in the libraryMethod file. Hopefully if one is missing from either of these it'll surface itself when we're testing.

ddelgrosso1:
@shaffeeullah there were a couple functions that I could not figure out what actual endpoint they were testing, i.e. most of the HMAC related functions as well as `getServiceAccount`.

PR #1586, opened 2021-09-08
chore: Label deprecated options with TS style syntax

fhinkel:
Deprecated properties still show up with api-extractor. To get the same behavior
as with JSDocs, we need to mark deprected items `internal`. See
https://api-extractor.com/pages/tsdoc/tag_deprecated/


fhinkel:
See internally b/199306482

shaffeeullah (approved):

PR #1590, opened 2021-09-15
test: pull testbench docker image, kokoro presubmit conformance test …

ddelgrosso1:
…hooks


ddelgrosso1:
There is currently a `conformance-test.yaml` that appears to run the tests on every pull request. However, it seems like most other tests / workflows use the .cfg / .sh approach. Not sure which is better and if we should just remove the older yaml file?

shaffeeullah:
This looks good to me, but you should also request review from someone who actually knows how this is supposed to work.

conformance-test/testBenchUtil.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function runTestBenchDockerImage(): Promise<Buffer> {
+  return new Promise((resolve, reject) => {
+    try {
+      resolve(execSync(RUN_CMD));
+    } catch (err) {
+      reject(err);
+    }
+  });
+}
```
danielbankhead, line 46: [Optional] This block can be streamlined to:

```suggestion
export async function getTestBenchDockerImage(): Promise<Buffer> {
  return execSync(PULL_CMD);
}

export async function runTestBenchDockerImage(): Promise<Buffer> {
  return execSync(RUN_CMD);
}
```

Note `execSync` would still blocking the main thread - if you're interested in having this more non-blocking, [`exec`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) might be a good choice

danielbankhead (commented):
The TS looks good; not familiar enough with the Kokoro components yet to make a call

.kokoro/presubmit/node10/conformance-test.cfg
```diff
@@ -0,0 +1,4 @@
+env_vars: {
```
fhinkel, line 1: If you look at the internal job configs, you see that we only have node12 configured. This file should be here: https://github.com/googleapis/nodejs-storage/tree/main/.kokoro/presubmit/node12

conformance-test/testBenchUtil.ts
```diff
@@ -0,0 +1,34 @@
+/*!
```
bcoe, line 1: nit: I would be tempted to cll this `test-bench-util.ts`, this just tends to be the JavaScript convention.

bcoe (approved):
Nits about filenames aside, LGTM.

PR #1591, opened 2021-09-16
fix: fixed runtime errors. all the tests still fail, but they run now

shaffeeullah:


conformance-test/retryStrategy.ts
```diff
@@ -66,7 +66,9 @@ const jsonToNodeApiMapping = fs.readFileSync(
   'utf-8'
 );
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const methodMap: Map<String, String[]> = new Map(Object.entries(JSON.parse(jsonToNodeApiMapping)));
+const methodMap: Map<String, String[]> = new Map(
```
ddelgrosso1, line 69: Is there something missing from the PR? I could be misreading something but this seems like just a linting fix?

ddelgrosso1: Neve mind, looks like there was a glitch with github and I wasn't seeing all changes.

ddelgrosso1 (approved):

PR #1593, opened 2021-09-16
samples: added retry samples

shaffeeullah:
Samples for customizing retry configuration on the storage instance


samples/storageRetryOnce.js
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * This application demonstrates how to perform basic operations on buckets with
+ * the Google Cloud Storage API.
+ *
+ * For more information, see the README.md under /storage and the documentation
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main() {
+  // [START storage_retry_once]
```
tritone, line 24: In Go we won't give the option to configure a particular number of retries (only time-related knobs such as initial delay, multiplier, and total timeout). So we won't be able to do an equivalent to this.

samples/storageRetryConditionallyIdempotent.js
```diff
@@ -0,0 +1,42 @@
...
+function main() {
+  // [START storage_retry_conditionally_idempotent]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
+   */
+  // Imports the Google Cloud client library and idempotency strategy options
+  const {Storage, IdempotencyStrategy} = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage({
```
tritone, line 32: We should show calling a library function using this configuration-- in other languages you configure on a per-call basis so something directly equivalent to this doesn't make sense.

samples/storageRetryConditionallyIdempotent.js
```diff
@@ -0,0 +1,42 @@
...
+    retryOptions: {
```
tritone, line 33: I feel like this could use a bit more in-line commentary since the configs will be specific to the language.

tritone (commented):
Couple thoughts. What do you think about just combining this into one sample? I feel like that would probably be sufficient.

shaffeeullah:
@tritone Let me know what you think of this new sample. The idea here is each language could show off all their configuration options and how they can be set.

samples/storageRetryConditionallyIdempotent.js
cojenco: +1  Retries are configured on a per-call basis in python, so we wouldn't be able to demonstrate anything equivalent here. My concern is that we usually try to implement code samples that are performing the same operations, and in this case, it might be confusing to the users. Is there a way to implement a specific operation instead (i.e. list buckets)?

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+ * the Google Cloud Storage API.
+ *
+ * For more information, see the README.md under /storage and the documentation
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main() {
+  // [START storage_configure_retries]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
```
cbonnie, line 30: I see that this comment seems standard across our samples - which lines is it referring to?

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+   */
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage({
+    retryOptions: {
+      autoRetry: true, //If this is false, requests will not retry and the setting of the below parameters will not have any effect.
```
cbonnie, line 38: "and the parameters below will not affect the retry behavior"
Can also use "settings" instead of "behavior" if it makes more sense, but the doc currently use "behavior"

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+      retryDelayMultiplier: 3, //The multiplier by which to increase the delay time between the completion of failed requests, and the initiation of the subsequent retrying request.
+      totalTimeout: 500, //The total time, starting from when the initial request is sent, after which an error will be returned, regardless of the retrying attempts made meanwhile.
```
cbonnie, line 40: "// The total time between an initial request getting sent and its timeout. After timeout, an error will be returned regardless of any retry attempts made during this time period."

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+      maxRetryDelay: 60, //The maximum delay time between requests. When this value is reached, retryDelayMultiplier will no longer be used to increase delay time.
+      maxRetries: 2, //	Maximum number of automatic retries attempted before returning the error.
```
cbonnie, line 42: "// The maximum number..." to be consistent

samples/configureRetries.js
cbonnie: Just a nit - can we change the formatting of the comments so it doesn't extend too far in the code block and become difficult to scroll? 

For example:
```
// If this is false, requests will not retry and the
// parameters below will not affect the retry behavior.
autoRetry: true,
// The multiplier by which...
// ...
retryDelayMultiplier: 3
```
I'm happy to make these changes myself in a follow up PR so they aren't a blocker.

samples/configureRetries.js
tritone: It's just standard for the node samples, but usually there is a line with the project ID for example? Do we need that here?

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+      idempotencyStrategy: IdempotencyStrategy.RetryAlways, // Will respect other retry settings and attempt to retry conditionally idempotent operations.
```
tritone, line 43: I think this is maybe showing too many config options. Perhaps we could show one timing option as well as the idempotency strategy, and then give a link to the part of the node docs which describe all the configs? (The link could also live in a text box above the samples rather than in the sample code itself).

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+    },
+  });
+  console.log(
```
tritone, line 46: IMO it's not necessary to do all this printing.

samples/configureRetries.js
```diff
@@ -0,0 +1,61 @@
...
+  console.log(`Auto Retry: ${storage.retryOptions.autoRetry}`);
+  console.log(
+    `Retry delay multiplier: ${storage.retryOptions.retryDelayMultiplier}`
+  );
+  console.log(`Total timeout: ${storage.retryOptions.totalTimeout}`);
+  console.log(`Maximum retry delay: ${storage.retryOptions.maxRetryDelay}`);
+  console.log(`Maximum retries: ${storage.retryOptions.maxRetries}`);
+  console.log(
+    `Idempotency strategy: ${storage.retryOptions.idempotencyStrategy}`
+  );
```
tritone, line 58: Let's add a function call to a function that will use the retry strategy (e.g. buckets.patch, to choose something conditionally idempotent). We'll need to include this function call to show the configuration for Go and Python so it will be good to have them be parallel.

samples/configureRetries.js
shaffeeullah: added some lines to uncomment

ddelgrosso1 (approved):

samples/configureRetries.js
```diff
@@ -0,0 +1,84 @@
...
+      // below will not affect retry behavior.
+      autoRetry: true, 
+      // The multiplier by which to increase the delay time between the 
+      // completion of failed requests, and the initiation of the subsequent 
+      // retrying request.
+      retryDelayMultiplier: 3,
+      // The total time between an initial request getting sent and its timeout. 
+      // After timeout, an error will be returned regardless of any retry attempts 
+      // made during this time period.
+      totalTimeout: 500,
```
cojenco, line 54: LGTM! Just a question here,  in this case, `maxRetries: 5` would take precedence of `totalTimeout`, correct?

cojenco (approved):

samples/configureRetries.js
shaffeeullah: what do you mean by that? once the timeout is hit, it will stop retrying, regardless of the number of retries completed

cojenco: Right, but in this case if we do the calculation, 5 retries will only take around 100 seconds. Once the `maxRetries: 5` is exhausted, it will stop retrying, regardless of the `totalTimeout`?

shaffeeullah: probably correct. totalTimeout is not just the additive time of all the retries, but also includes the time the function takes to run. so if it's a long running function, all 5 might not be exhausted.

cojenco: Ahhhh I see, that makes sense. Thanks for the clarification!

tritone (approved):

samples/configureRetries.js
```diff
@@ -0,0 +1,88 @@
...
+
+  async function deleteFileWithCustomizedRetrySetting() {
+    await storage.bucket(bucketName).file(fileName).delete();
+    console.log(`File ${fileName} deleted with a customized retry strategy.`);
+  }
+
+  deleteFileWithCustomizedRetrySetting();
+  // [END storage_configure_retries]
+}
+main(...process.argv.slice(2));
```
ddelgrosso1, line 88: Perhaps add error handling for `unhandledRejection`.

ddelgrosso1 (approved):

samples/configureRetries.js
```diff
@@ -0,0 +1,92 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const {
+  IdempotencyStrategy,
+} = require('@google-cloud/storage/build/src/storage');
+
+/**
+ * This application demonstrates how to perform basic operations on buckets with
```
cbonnie, line 20: I would get rid of lines 19-25 if possible. it seems like copy from a template that sounds too generic to be helpful.

samples/configureRetries.js
```diff
@@ -0,0 +1,92 @@
...
+ * the Google Cloud Storage API.
+ *
+ * For more information, see the README.md under /storage and the documentation
```
cbonnie, line 23: Update "For more info..." sentence to: 

"See reference documentation for the options used below at https://googleapis.dev/nodejs/storage/latest/global.html#StorageOptions."

samples/configureRetries.js
```diff
@@ -0,0 +1,92 @@
...
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main(bucketName = 'my-bucket', fileName = 'test.txt') {
+  // [START storage_configure_retries]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
```
cbonnie, line 30: Is there a reason the lines below need to be commented out in the sample (i know it's a standard thing, but i'm wondering if it needs to be here in particular)? I'm not too familiar with node.js but I think a better change could be:

```
// The ID of your GCS bucket
const bucketName = `your-unique-bucket-name`;

// The ID of your GCS file
const fileName = `your-file-name`;
...
```

samples/configureRetries.js
```diff
@@ -0,0 +1,92 @@
...
+  // The ID of your GCS bucket
+  // const bucketName = 'your-unique-bucket-name';
+
+  // The ID of your GCS file
+  // const fileName = 'your-file-name';
+
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
+
+  // Creates a client
```
cbonnie, line 41: Again, not sure if this is feasible in node.js, but something close to this would be my suggestion: 

```
// Creates a client. 
const storage = new Storage({

    retryOptions: {
        ...
        autoRetry: true,
        ...
        // Customizes the retry behavior with a wait time multiplier of 3 (default=2)
        retryDelayMultiplier: 3,
        ...
        // Customizes the retry behavior with a run time deadline of 500 seconds
        // (default=120)
        totalTimeout: 500,
        ...
        // Customizes the retry behavior with 5 maximum retries before returning
        // an error (default=3)
```

PR #1601, opened 2021-09-21
feat: check that instructions were utilized in testbench

ddelgrosso1:

### DIFF
--- a/conformance-test/retryStrategy.ts
+++ b/conformance-test/retryStrategy.ts
@@ -81,9 +81,9 @@ const OPTIONS = {
   },
 };
 
-const EMULATOR_HOST =
-  process.env.STORAGE_EMULATOR_HOST || 'http://localhost:9000';
-const storage = new Storage({apiEndpoint: EMULATOR_HOST});
+const TESTBENCH_HOST =
+  process.env.STORAGE_EMULATOR_HOST || 'http://localhost:9000/';
+const storage = new Storage({apiEndpoint: TESTBENCH_HOST});
 
 describe('retry conformance testing', () => {
   before(async () => {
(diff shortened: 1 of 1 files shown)


conformance-test/retryStrategy.ts
```diff
@@ -19,12 +19,14 @@ import * as path from 'path';
 import * as uuid from 'uuid';
 import pLimit = require('p-limit');
 import * as assert from 'assert';
+import fetch from 'node-fetch';
```
shaffeeullah, line 22: does this introduce a dev dependency on node-fetch?

conformance-test/retryStrategy.ts
```diff
@@ -84,6 +85,10 @@ const OPTIONS = {
   },
 };
 
+const EMULATOR_HOST =
+  process.env.STORAGE_EMULATOR_HOST || 'http://localhost:9000';
```
shaffeeullah, line 89: does this need a trailing `/`?

conformance-test/retryStrategy.ts
```diff
@@ -84,6 +85,10 @@ const OPTIONS = {
   },
 };
 
+const EMULATOR_HOST =
```
shaffeeullah, line 88: should we call this `TESTBENCH_HOST`?

conformance-test/retryStrategy.ts
ddelgrosso1: We already had `node-fetch` listed as a dev dependency in the `package.json` which is why I went with it as opposed to the `https` lib in node.

shaffeeullah (approved):

PR #1614, opened 2021-09-24
samples: changed PAP unspecified to inherited

### LINKED ISSUE #1616: Change Public Access Prevention "unspecified" to "inherited"
GCS is changing the `unspecified` public access prevention value to `inherited`. 

This will require the following changes: 
1. Modify the values in the library. Constants/enum values for unspecified: these should be marked as deprecated. They can stay in the library if deletion would constitute a breaking change
2. Update the system tests. 
3. Create a new sample with region tag public_access_prevention_inherited. So we don't break any documentation links, please do not delete the unspecified sample or sample test from your repo at this point in time.

Please see nodejs example here: https://github.com/googleapis/nodejs-storage/pull/1614/files

Please wait for approval from myself or @tritone before merging or releasing this change so we can coordinate around release dates.  New inherited value should be rolled out to allowlisted testing projects this coming week; should be at 100% by 9/30

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1616


samples/setPublicAccessPreventionInherited.js
```diff
@@ -0,0 +1,49 @@
...
+  // Imports the Google Cloud client library
+  const {Storage} = require('@google-cloud/storage');
+
+  // Creates a client
+  const storage = new Storage();
+  async function setPublicAccessPreventionInherited() {
+    // Sets public access prevention to 'inherited' for the bucket
+    await storage.bucket(bucketName).setMetadata({
+      iamConfiguration: {
+        publicAccessPrevention: 'inherited',
```
tritone, line 39: I'm going to assume that you just took the unspecified sample and did s/unspecified/inherited. Let me know if there were any other changes...

samples/system-test/buckets.test.js
```diff
@@ -184,9 +184,9 @@ it("should get a bucket's public access prevention metadata", async () => {
   assert.ok(metadata.iamConfiguration.publicAccessPrevention);
 });
 
-it('should set public access prevention to unspecified', async () => {
```
tritone, line 187: The test for the unspecified sample should stay until we remove it from the docs.

system-test/storage.ts
```diff
@@ -881,21 +881,21 @@ describe('storage', () => {
...
       const publicAccessPreventionStatus =
         bucketMetadata.iamConfiguration.publicAccessPrevention;
       return assert.strictEqual(
         publicAccessPreventionStatus,
-        PUBLIC_ACCESS_PREVENTION_UNSPECIFIED
+        PUBLIC_ACCESS_PREVENTION_INHERITED
       );
     });
 
     it('makes public a bucket with unspecified public access prevention', async () => {
```
tritone, line 895: Still says unspecified here, check for additional occurences.

tritone (commented):
Couple small things, generally seems fine!

tritone (approved):
Thanks!

ddelgrosso1 (approved):

JustinBeckwith (approved):

samples/system-test/buckets.test.js
shaffeeullah: this test needs to be deleted because it is failing. (setting it to unspecified is returning inherited)

SidBlack (commented):
alles ok

PR #1617, opened 2021-09-25
fix: respect precondition settings from constructors

ddelgrosso1:

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1420,38 +1420,7 @@ class Bucket extends ServiceObject {
       maxRetries = 0;
     }
 
-    if (
-      options?.ifGenerationMatch ||
-      this.instancePreconditionOpts?.ifGenerationMatch
-    ) {
-      options.ifGenerationMatch =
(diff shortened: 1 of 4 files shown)


src/bucket.ts
```diff
@@ -1417,6 +1420,39 @@ class Bucket extends ServiceObject {
       maxRetries = 0;
     }
 
+    if (
```
JustinBeckwith, line 1423: If I'm understanding this right, could we accomplish the same thing with:

```js
if (this.instancePreconditionOpts) {
  Object.assign(options, this.instancePreconditionOpts);
}
```

This should just take every value on `instancePreconditionOpts` and smoosh it into `options`, if the values in `instancePreconditionOpts` are not undefined. 

```

JustinBeckwith (approved):

shaffeeullah (approved):
Thanks, Denis! PR to unblock failing samples test is here https://github.com/googleapis/nodejs-storage/pull/1620

PR #1634, opened 2021-10-01
fix: fix logic for buckets that do not have a generation

ddelgrosso1:


test/bucket.ts
```diff
@@ -3237,4 +3238,42 @@ describe('Bucket', () => {
       );
     });
   });
+  describe('disableAutoRetryConditionallyIdempotent_', () => {
```
shaffeeullah, line 3241: is there a test for making sure autoRetry stays true? we should cover at least one case of that. otherwise, looks good. thanks, denis!

shaffeeullah (approved):

test/bucket.ts
ddelgrosso1: Added!

PR #1664, opened 2021-10-13
chore: use this.request in isPublic instead of util.makeRequest

ddelgrosso1:


ddelgrosso1:
TLDR: util.makeRequest was not sending the headers and the interceptors property is private so they cannot be copied over. I switched to using this.request instead.

shaffeeullah (approved):
Code changes look good. maybe change the PR title to be `chore` instead of `fix`? This doesnt actually change the functionality at all and we wouldnt want to release for this change

frankyn (changes requested):

PR #1666, opened 2021-10-14
fix: do not use src precondition options in copy.

ddelgrosso1:

The PR addressees the issue that currently a copy operation either creates or overwrites a destination file. The issue was that the preconditions were being checked against the source file.


shaffeeullah (commented):
can you also please update the description of this PR with why we're doing this? so if someone is ever tracing the change in the future they can understand it.

shaffeeullah (approved):
Thanks, Denis!

PR #1702, opened 2021-11-04
fix: add scenario 3 conformance tests

ddelgrosso1:
test: disable retries when creating write stream as part of upload.

test: add scenario 3 tests

update nodejs-common to 3.8.1

test: disable autoRetry when calling createWriteStream from save

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -3756,6 +3756,9 @@ class File extends ServiceObject<File> {
     const returnValue = retry(
       async (bail: (err: Error) => void) => {
         await new Promise<void>((resolve, reject) => {
+          if (maxRetries === 0) {
+            this.storage.retryOptions.autoRetry = false;
+          }
           const writable = this.createWriteStream(options)
             .on('error', err => {
               if (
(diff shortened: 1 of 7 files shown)


shaffeeullah (approved):
I think this should be `fix` to trigger a release from the common upgrade. otherwise, looks good

src/file.ts
```diff
@@ -4081,7 +4084,13 @@ class File extends ServiceObject<File> {
  * that a callback is omitted.
  */
 promisifyAll(File, {
-  exclude: ['publicUrl', 'request', 'save', 'setEncryptionKey'],
+  exclude: [
+    'publicUrl',
+    'request',
+    'save',
+    'setEncryptionKey',
+    'shouldRetryBasedOnPreconditionAndIdempotencyStrat',
```
shaffeeullah, line 4092: might have to update unit tests because of this

ddelgrosso1: Yes, it broke a bunch of other tests as well which I am trying to figure out.

PR #1711, opened 2021-11-10
samples: upload without authentication

shaffeeullah:
This is the canonical sample for uploading data without authentication.
Tests to be written once we decide on the format of the sample.

Related customer issue: https://github.com/googleapis/gcs-resumable-upload/issues/201


samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+  // The ID of your GCS bucket
+  // const bucketName = 'your-unique-bucket-name';
+
+  // The contents that you want to upload
+  // const contents = 'these are my contents';
+
+  // The new ID for your GCS file
+  // const destFileName = 'your-new-file-name';
+
+  // Imports the Google Cloud client library
```
cbonnie, line 33: How about "Imports the Google Cloud Node.js client library"

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+  const fetch = require('node-fetch');
+
+  // Creates a client
+  const storage = new Storage();
+
+  async function uploadWithoutAuthentication() {
+    const file = storage.bucket(bucketName).file(destFileName);
+    let location; // endpoint to which we should upload the file
+
+    // Option 1: use file.createResumableUpload
```
cbonnie, line 45: // Option 1: Use file.createResumableUpload. 
// It returns an authenticated endpoint
// to which you can make requests without
// credentials. 

(however line breaking works)

samples/README.md
```diff
@@ -1855,6 +1856,23 @@ __Usage:__
 
 
 
+### Upload Without Auth
```
cbonnie, line 1859: ### Upload without authenticating

cbonnie: This is proper sentence case so it'll render correctly when added to docs. Also provides more detail as to what "auth" means.

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+    // file.createResumableUpload returns an authenticated endpoint
+    // to which we can make requests without credentials
+    [location] = await file.createResumableUpload(); //auth required
+
+    // Option 2: use signed URLs to manually start resumable upload
```
cbonnie, line 50: // Option 2: Use signed URLs to manually start resumable uploads.
// Authenticating is required to get the signed URL, but isn't 
// required to start the resumable upload.

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+    // no auth required
+    const resumableSession = await fetch(signedUrl, {
+      method: 'POST',
+      headers: {
+        'x-goog-resumable': 'start',
+      },
+    });
+    location = resumableSession.headers.location;
+
+    // passing the location to file.save removes the need to
```
cbonnie, line 70: // Passes the location to file.save so you
// don't need to authenticate this call.

(however the line breaking works)

samples/README.md
shaffeeullah: This is autogenerated by a bot and cannot be changed (without a ton of effort, like contacting the team that wrote the bot)

shaffeeullah: I think I can change the word "auth" to "authenticating" by renaming the file the sample is in. Do you think i should make that change?

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function main(
+  bucketName = 'my-bucket',
+  contents = 'these are my file contents',
+  destFileName = 'file.txt'
+) {
+  // [START storage_upload_without_authentication]
```
shaffeeullah, line 20: @cbonnie what do you think about this region tag name?

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,85 @@
...
+      validation: false,
+    });
+
+    console.log(`${destFileName} uploaded to ${bucketName}`);
+  }
+
+  uploadWithoutAuthentication().catch(console.error);
+  // [END storage_upload_without_authentication]
+}
+
```
ddelgrosso1, line 84: Maybe just add the standard error handling here, `process.on(....`

samples/uploadWithoutAuth.js
```diff
@@ -0,0 +1,88 @@
...
+  const fetch = require('node-fetch');
+
+  // Creates a client
+  const storage = new Storage();
+
+  async function uploadWithoutAuthentication() {
+    const file = storage.bucket(bucketName).file(destFileName);
+    let location; // endpoint to which we should upload the file
+
```
frankyn, line 44: I'm wondering if this sample should be split into two separate samples as to not overload, but it's readable so not completely against it.

frankyn (changes requested):

samples/uploadWithoutAuth.js
shaffeeullah: I think if we separate this into two samples, we run the risk of someone missing one of the samples and only learning one way it could be done. It also depends on how the page is structured though. @cbonnie do you have thoughts here?

samples/README.md
cbonnie: I think it'd be best to make that change. "Upload without auth" seems like it could imply many things (upload without authenticating, upload without OAuth, etc)

samples/uploadWithoutAuth.js
cbonnie: Looks fine to me. I don't think the region tag is visible to the user.

samples/uploadWithoutAuth.js
cbonnie: We can definitely split this up into two samples and make it clear (in writing) that there's two ways to do the task.

frankyn: You could nest them in the same region block: https://cloud.google.com/storage/docs/listing-objects#code-samples is an example

shaffeeullah: oh cool! i didnt realize that. thanks for this example. i will make this change

shaffeeullah: @frankyn @cbonnie please take another look

samples/uploadWithoutAuthenticationSignedUrl.js
```diff
@@ -0,0 +1,84 @@
...
+
+    // no auth required
+    const resumableSession = await fetch(signedUrl, {
+      method: 'POST',
+      headers: {
+        'x-goog-resumable': 'start',
+      },
+    });
+
+    // endpoint to which we should upload the file
```
cbonnie, line 63: "Endpoint"

cbonnie (approved):

frankyn (approved):

ddelgrosso1 (approved):

PR #1715, opened 2021-11-18
docs: change properties with function value to methods

fhinkel:
Internally, see b/206943288

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -629,7 +629,10 @@ class Bucket extends ServiceObject {
   acl: Acl;
   iam: Iam;
 
-  getFilesStream: (query?: GetFilesOptions) => Readable;
+  getFilesStream(query?: GetFilesOptions): Readable {
+    // placeholder body, overwritten in constructor
+    return new Readable();
+  }
   signer?: URLSigner;
 
   private instanceRetryValue?: boolean;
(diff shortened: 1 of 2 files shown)


src/bucket.ts
```diff
@@ -629,7 +629,10 @@ class Bucket extends ServiceObject {
   acl: Acl;
   iam: Iam;
 
-  getFilesStream: (query?: GetFilesOptions) => Readable;
+  getFilesStream(query?: GetFilesOptions): Readable {
+    // placeholder body, overwritten in constructor
```
bcoe, line 633: This seems reasonble to me, but I would open a ticket in parser to fix this issue, and reference the ticket # here.

fhinkel: You mean in api-documnter? Yeah, I can open an issue there.

bcoe (approved):

shaffeeullah (changes requested):
Does this PR have to be `fix`? Can it be `docs`?

fhinkel:
> Does this PR have to be `fix`? Can it be `docs`?

Sure!

shaffeeullah (approved):

src/bucket.ts
fhinkel: done: https://github.com/microsoft/rushstack/issues/3031

PR #1720, opened 2021-11-19
fix: stop File.download from truncating output file on failure

### LINKED ISSUE #1713: `File.download` truncates output file on failure
The following minimal example attempts to download an object that does not
exist. We expect it to do nothing, but instead it truncates the output file(!):

```javascript
async function main() {
  const gcs = require("@google-cloud/storage");
  const bucketName = "gcp-public-data-landsat";
  const f = new gcs.Storage().bucket(bucketName).file("enoent");
  try {
    await f.download({ destination: "/tmp/gcsoutput" });
  } catch (e) {
    console.error("Download failed:", e.message);
  }
}
main();
```

As expected, this fails because the object does not exist:

> Download failed: No such object: gcp-public-data-landsat/enoent

However, it also creates or truncates `/tmp/gcsoutput`, as if the object
actually had existed and been 0 bytes long.

This is surprising and violates atomicity guarantees: a download that fails to
start should have no side effects, and should certainly not truncate a file. In
my use case, creating an empty file is just as bad as truncating an existing
one.

By way of comparison, using `curl(1)` for the same task *does not* touch the
output file if download fails:

```
$ printf 'hey\n' >/tmp/curlout
$ curl -fsSL https://storage.googleapis.com/gcp-public-data-landsat/enoent -o /tmp/curlout
curl: (22) The requested URL returned error: 404 
$ cat /tmp/curlout
hey
```

Suggestion: in `File#download`, only call `fs.createWriteStream(destination)`
once OK response headers have been received from the server. (If the download
fails midway through, ending up with the truncated contents in the destination
makes sense.)

#### Environment details

  - OS: Debian (but shouldn't matter)
  - Node.js version: v16.8.0 (but shouldn't matter)
  - npm version: v7.21.0 (but shouldn't matter)
  - `@google-cloud/storage` version: v5.15.4

#### Steps to reproduce

 1. `printf 'hello\n' >/tmp/gcsoutput`
 2. Run the script listed above, which should fail to download and leave
    `/tmp/gcsoutput` unchanged.
 3. Note that `/tmp/gcsoutput` is now in fact empty.

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1713 🦕

### DIFF
--- a/test/file.ts
+++ b/test/file.ts
@@ -2715,6 +2715,37 @@ describe('File', () => {
         });
       });
 
+      it.only('file contents should remain unchanged if file nonexistent', done => {
+        tmp.setGracefulCleanup();
+        tmp.file(async (err, tmpFilePath) => {
+          console.log(tmpFilePath);
+          assert.ifError(err);
+
+          const fileContents = 'file contents that should remain unchanged';
+          fs.writeFile(tmpFilePath, fileContents, 'utf-8', (err) => {
+            assert.ifError(err);
+          });
+
+          const error = new Error('Error.');
+          Object.assign(fileReadStream, {
+            _read(this: Readable) {
+              process.nextTick(() => {
+                this.emit('error', error);
+              });
+            },
+          });
+
+          file.download({destination: tmpFilePath},  (err: Error) => {
+            assert.strictEqual(err, error);
+            fs.readFileSync(tmpFilePath, (err, tmpFileContents) => {
+              assert.ifError(err);
+              assert.strictEqual(fileContents, tmpFileContents.toString());
+              done();
+            });
+          });
+        });
+      });
+
       it('should execute callback with error', done => {
         tmp.setGracefulCleanup();
         tmp.file((err, tmpFilePath) => {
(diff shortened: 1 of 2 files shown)


ddelgrosso1 (approved):

test/file.ts
```diff
@@ -2715,6 +2715,61 @@ describe('File', () => {
...
 
+      it('should process the entire stream', done => {
+        tmp.setGracefulCleanup();
+        tmp.file(async (err, tmpFilePath) => {
+          assert.ifError(err);
+
+          const fileContents = 'abcdefghijklmnopqrstuvwxyz';
+
+          fileReadStream.on('resume', () => {
+            fileReadStream.emit('data', fileContents);
```
danielbankhead, line 2726: Minor: for future reviewers, we may want to leave a comment as to why we're explicitly emitting twice

danielbankhead (approved):
Looks good!

test/file.ts
shaffeeullah: Do you think the test title is sufficient? If not, we should change the test title to something more descriptive.

ddelgrosso1: I think the title is sufficient, describes what it is testing succinctly enough.

PR #1733, opened 2021-12-16
samples: Delete setPublicAccessPreventionUnspecified.js

shaffeeullah:


samples/setPublicAccessPreventionUnspecified.js
```diff
@@ -1,49 +0,0 @@
...
-   * TODO(developer): Uncomment the following lines before running the sample.
-   */
-  // The name of your GCS bucket
-  // const bucketName = 'Name of a bucket, e.g. my-bucket';
-  // Imports the Google Cloud client library
-  const {Storage} = require('@google-cloud/storage');
-
-  // Creates a client
-  const storage = new Storage();
-  async function setPublicAccessPreventionUnspecified() {
```
frankyn, line 35: There should be a test for this, could you remove it as well?

frankyn (changes requested):

samples/setPublicAccessPreventionUnspecified.js
shaffeeullah: no test. the test broke when the backend changed, so we removed it

frankyn: When the backend changed?

frankyn (approved):

PR #1736, opened 2021-12-22
chore(deps): gcs-resumable-upload migration

shaffeeullah:


system-test/kitchen.ts
```diff
@@ -0,0 +1,184 @@
...
+import * as assert from 'assert';
+import {describe, it, beforeEach} from 'mocha';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import {Readable} from 'stream';
+import {createURI, ErrorWithCode, upload} from '../src/gcs-resumable-upload';
+
+const bucketName = process.env.BUCKET_NAME || 'gcs-resumable-upload-test';
+const filePath = path.join(__dirname, '../../system-test/data/20MB.zip');
```
danielbankhead, line 24: If we can, let's remove this binary in favor of generating a temp file. See:
- https://nodejs.org/api/os.html#ostmpdir
- https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

If we want a 20MB file of random bytes:
- `crypto.randomBytes(1024 * 1024 * 20)`

This way we don't have to check in a test blob into the repo.

danielbankhead:
After this is merged, we should consider a refactor:
- to remove unnecessary deps like `pumpify`, `abort-controller`, `is-stream`, etc. in favor of native deps
- more tightly integrate `gcs-resumable-upload` stream handling with File (shared type definitions, removing things that are optional that File wouldn't use)

system-test/kitchen.ts
ddelgrosso1: Good call. I would agree as well, I'm not a huge fan of committing blobs.

system-test/kitchen.ts
```diff
@@ -0,0 +1,189 @@
...
+import * as crypto from 'crypto';
+import * as os from 'os';
+import {Readable} from 'stream';
+import {createURI, ErrorWithCode, upload} from '../src/gcs-resumable-upload';
+
+const bucketName = process.env.BUCKET_NAME || 'gcs-resumable-upload-test';
+tmp.setGracefulCleanup();
+const tmpFileContents = crypto.randomBytes(1024 * 1024 * 20);
+const filePath = path.join(os.tmpdir(), '20MB.zip');
+fs.createWriteStream(filePath).write(tmpFileContents);
```
danielbankhead, line 29: This stream should be closed to avoid open handle issues (https://nodejs.org/api/fs.html#filehandlecreatereadstreamoptions)

Perhaps write stream would be better easier? https://nodejs.org/api/fs.html#fswritesyncfd-buffer-offset-length-position

danielbankhead:
https://github.com/googleapis/gcs-resumable-upload v4.0.0 has been merged - should we include it in this PR?

danielbankhead (approved):

danielduhh:
late review here:
@danielbankhead do you mind adding issues for the follow up action items?
@shaffeeullah does the single index.ts file need to be nested in the gcs-resumable-media folder? We could just rename the file to resumable media and move under src/ with the rest of the package

PR #1745, opened 2022-01-10
fix: remove compodoc dev dependency

ddelgrosso1:

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "xdg-basedir": "^4.0.0"
   },
   "devDependencies": {
-    "@compodoc/compodoc": "^1.1.7",
     "@google-cloud/pubsub": "^2.0.0",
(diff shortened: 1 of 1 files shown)


ddelgrosso1:
This dependency was/is breaking CI builds for Node 15 due to only supporting 12, 14, 16.

shaffeeullah:
Is it not necessary?

ddelgrosso1:
@shaffeeullah that is what I was trying to see by removing it and running everything through CI. It was brought in as part of the GCS Resumable move but not sure what it was doing there or if we need it here.

shaffeeullah (approved):

ddelgrosso1:
This was used in the gcs-resumable-upload library to generate the documentation. It is not needed here as we are not publicly exposing those interfaces through storage but rather just consuming the code that was internalized here. I feel comfortable that this can be safely removed.

PR #1754, opened 2022-01-13
feat: Expose `chunkSize` param for `CreateResumableUploadOptions`

danielbankhead:


ddelgrosso1 (approved):

src/file.ts
```diff
@@ -190,6 +190,7 @@ export type PredefinedAcl =
   | 'publicRead';
 
 export interface CreateResumableUploadOptions {
+  chunkSize?: number;
```
tritone, line 193: Does this need its own documentation somewhere, or is it covered by https://github.com/googleapis/nodejs-storage/blob/437d40057f48ad7369c1743acf7578263ef56998/src/gcs-resumable-upload/index.ts#L119 ?

danielbankhead: Good call-out, added some more discoverable documentation

tritone (approved):

PR #1761, opened 2022-01-24
test: reduce number of operations and make names more unique

### LINKED ISSUE #1758: Version v0.11.0 of testbench enforces bucket name uniqueness and causes conformance tests to fail
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.

  - Search the issues already opened: https://github.com/googleapis/nodejs-storage/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS:
  - Node.js version:
  - npm version:
  - `@google-cloud/storage` version:

#### Steps to reproduce

  1. Unpin testbench version from v0.10.0 to latest (v0.11.0+)
  2. Run conformance tests

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

ddelgrosso1:

Fixes #1758  🦕

### DIFF
--- a/conformance-test/retryStrategy.ts
+++ b/conformance-test/retryStrategy.ts
@@ -64,20 +64,7 @@ const methodMap: Map<String, String[]> = new Map(
 
 const DURATION_SECONDS = 600; // 10 mins.
 
-const TESTS_PREFIX = `storage-retry-tests-${shortUUID()}-`;
-const FILE_OPTIONS = {
-  preconditionOpts: {
-    ifGenerationMatch: 0,
-    ifMetagenerationMatch: 1,
-  },
-};
-
-const BUCKET_OPTIONS = {
-  preconditionOpts: {
-    ifMetagenerationMatch: 1,
-  },
-};
-
+const TESTS_PREFIX = `storage.retry.tests.${shortUUID()}.`;
 const TESTBENCH_HOST =
   process.env.STORAGE_EMULATOR_HOST || 'http://localhost:9000/';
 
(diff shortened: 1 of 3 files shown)


conformance-test/retryStrategy.ts
```diff
@@ -64,20 +64,7 @@ const methodMap: Map<String, String[]> = new Map(
...
-  },
-};
-
-const BUCKET_OPTIONS = {
-  preconditionOpts: {
-    ifMetagenerationMatch: 1,
-  },
-};
-
+const TESTS_PREFIX = `storage.retry.tests.${shortUUID()}.`;
```
ddelgrosso1, line 67: Changed to using `.` due to https://cloud.google.com/storage/docs/naming-buckets (see section about using a period in names).

conformance-test/retryStrategy.ts
```diff
@@ -254,7 +235,7 @@ async function createFileForTest(
 }
 
 function generateName(storageMethodString: String, bucketOrFile: string) {
-  return `${TESTS_PREFIX}${storageMethodString.toLowerCase()}${bucketOrFile}`;
+  return `${TESTS_PREFIX}${storageMethodString.toLowerCase()}${bucketOrFile}.${shortUUID()}`;
```
ddelgrosso1, line 238: Adding an extra UUID on the end to enforce uniqueness. This eliminates the need to delete buckets at the end of each test and saves a few network calls.

conformance-test/test-bench-util.ts
```diff
@@ -21,7 +21,7 @@ const PORT = new URL(HOST).port;
 const CONTAINER_NAME = 'storage-testbench';
 const DEFAULT_IMAGE_NAME =
   'gcr.io/cloud-devrel-public-resources/storage-testbench';
-const DEFAULT_IMAGE_TAG = 'v0.10.0';
+const DEFAULT_IMAGE_TAG = 'v0.11.0';
```
ddelgrosso1, line 24: We will need to be more aware when test bench releases and make sure we update this accordingly after locally testing. I am hesitant to go back to using `latest` after it broke our tests and potentially blocked releases for a few days.

ddelgrosso1:
TLDR: This PR does 2 primary things: 1. makes bucket naming more unique as testbench as of v0.11.0 is enforcing naming uniqueness. 2. We essentially had a lot of duplicated (n^2ish) operations going on in part due to misunderstanding the sequencing of `before` / `beforeEach` with `describe` blocks. This has now been unwound and most everything is part of `beforeEach` with a new nested `describe` block to encapsulate the individual tests.

conformance-test/retryStrategy.ts
shaffeeullah: what is the benefit of using `.` over `-`?

conformance-test/retryStrategy.ts
shaffeeullah: shouldnt text prefix + storagemethodstring be globally unique? which case is this covering?

ddelgrosso1: Currently test prefix is defined as a constant at the file level. As a result every. test executing in this file utilizes the same prefix and because some tests are repeated in different scenarios the current setup does not guarantee uniqueness.

shaffeeullah (approved):

shaffeeullah:
@ddelgrosso1 any reason this is still a draft PR?

danielbankhead (approved):

PR #1764, opened 2022-01-26
fix(gcs-resumable-upload): Stop Duplicate Response Handlers on Retries

danielbankhead:
Porting fix from https://github.com/googleapis/gcs-resumable-upload/pull/502:

> Previously, `startUploading` would setup a 'response' event listener for passing responses to `responseHandler` - this is problematic in the event of a retry as the retry would create its own request (thus previous requests would trigger `responseHandler` as well).
> 
> Instead, we can accomplish the desired functionality by checking `makeRequestStream`'s return value.

### DIFF
--- a/src/gcs-resumable-upload/index.ts
+++ b/src/gcs-resumable-upload/index.ts
@@ -767,12 +767,6 @@ export class Upload extends Pumpify {
       },
     });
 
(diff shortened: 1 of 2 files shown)


ddelgrosso1 (approved):
Thanks for porting this over

PR #1768, opened 2022-01-27
fix: removed double semicolon

matteobaccan:
I have removed an empty statement behind a double semicolon

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

PR #1774, opened 2022-02-03
test: Replace cached test binary with generated file

danielbankhead:
Removes static `.gz` file and repaired `should upload a gzipped file and download it` test.

Should unblock #1768


system-test/storage.ts
```diff
@@ -2521,41 +2516,37 @@ describe('storage', () => {
...
-        file!.download(downloadCallback);
+      const contents: Buffer = await new Promise((resolve, reject) => {
+        return file.download((error, content) => {
+          if (error) return reject(error);
+          resolve(content);
+        });
       });
+
+      assert.strictEqual(contents.toString(), expectedContents);
+      file.delete();
```
shaffeeullah, line 2549: does this need to be awaited?

danielbankhead: solid catch

ddelgrosso1 (approved):

PR #1797, opened 2022-02-22
docs: Document `DownscopedClient` Usage

### LINKED ISSUE #1522: Document how to use a CredentialAccessBoundary/DownscopedToken
There is currently no documentation or example anywhere obvious that describes how you are supposed to use a `CredentialAccessBoundary` or `DownscopedClient `with `Storage` either in this library or in `google-auth-library`. It would be very helpful to have this information for issuing onetime tokens for accessing specific resources within a bucket. 

This seems like a relatively common need for projects that deal with file uploads/downloads and so would be worth explicitly documenting since it's not remotely close to obvious.

danielbankhead:
Await https://github.com/googleapis/nodejs-common/pull/733

Fixes #1522 🦕

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -521,7 +521,7 @@ export class Storage extends Service {
    *     retry a conditionally idempotent operation.
    * @property {string} [userAgent] The value to be prepended to the User-Agent
    *     header in API requests.
-   * @property {object} [authClient] GoogleAuth client to reuse instead of creating a new one.
+   * @property {object} [authClient] `AuthClient` or `GoogleAuth` client to reuse instead of creating a new one.
    * @property {number} [timeout] The amount of time in milliseconds to wait per http request before timing out.
    * @property {object[]} [interceptors_] Array of custom request interceptors to be returned in the order they were assigned.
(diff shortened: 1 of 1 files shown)


ddelgrosso1:
No objection to this PR but this looks like something we should probably further document in a sample at some point in the near future.

ddelgrosso1 (approved):

ddelgrosso1:
Added 'do not merge' until common is released.

danielbankhead:
> No objection to this PR but this looks like something we should probably further document in a sample at some point in the near future.

I think it's fair to add the sample to this PR; makes the resolution a bit more comprehensive.

danielbankhead:
Offline update: We'll need to sync with the rest of the team on rolling out DownscopedClient for Storage samples to the other langs. Will merge this now, pending a sample directly in this repo in the future. In the meantime, these will suffice for Node.js for now:
- https://github.com/googleapis/google-auth-library-nodejs#sample-usage-1
- https://github.com/googleapis/google-auth-library-nodejs/blob/main/samples/downscopedclient.js

PR #1812, opened 2022-03-10
samples: set client endpoint

### LINKED ISSUE #1781: Sample request: setting endpoint on storage client
We need samples added that describe how to set the endpoint for our storage clients.
Such as setting for localhost/emulator, or private/vpc, or for the soon to be added locational endpoints feature.

Sample tracker & Request:

https://docs.google.com/document/d/1aeL7jgs789s7xg9khTxGzGWkb4k-t6LUk-uqhhs1Vuw/edit#

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/1781 🦕


shaffeeullah:
@ddelgrosso1 @Breathtender will write tests after I get confirmation that this sample looks good to you in terms of structure / things i'm printing

src/storage.ts
```diff
@@ -540,7 +540,7 @@ export class Storage extends Service {
    * @property {number} [timeout] The amount of time in milliseconds to wait per http request before timing out.
    * @property {object[]} [interceptors_] Array of custom request interceptors to be returned in the order they were assigned.
    * @property {string} [apiEndpoint = storage.google.com] The API endpoint of the service used to make requests.
-   * @property {boolean} [useAuthWithCustomEndpoint] Controls whether or not to use authentication when using a custom endpoint.
+   * @property {boolean} [useAuthWithCustomEndpoint = false] Controls whether or not to use authentication when using a custom endpoint.
```
shaffeeullah, line 543: @ddelgrosso1 please confirm that this does actually default to false. (I couldn't find a default value that we set, and so an undefined boolean would evaluate to false. But please make sure I'm not missing anything)

ddelgrosso1: You are correct, this is defined as an optional boolean or undefined. By default it would be undefined which means we would not use auth with a custom endpoint.

Breathtender:
Thanks @shaffeeullah yes this looks good

shaffeeullah:
> Thanks @shaffeeullah yes this looks good

@Breathtender awesome. Just added tests. Please take another look

samples/setClientEndpoint.js
```diff
@@ -0,0 +1,48 @@
...
+ * the Google Cloud Storage API.
+ *
+ * For more information, see the README.md under /storage and the documentation
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main(apiEndpoint = 'https://storage.googleapis.com') {
+  // [START storage_set_client_endpoint]
+  /**
+   * TODO(developer): Uncomment the following lines before running the sample.
```
Breathtender, line 26: Wondering if this is unnecessary duplication since apiEndpoint is also set in the function call. But if this is the way the rest of the nodejs samples are doing it, then no worries, carry on.

samples/system-test/storage.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+'use strict';
+
+const {assert} = require('chai');
+const {it} = require('mocha');
+const cp = require('child_process');
+
+const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
+
+it('should intialize storage with a custom api enpoint', async () => {
```
Breathtender, line 23: nit: enpoint -> endpoint

unforced (approved):
Couple little things but otherwise LGTM

samples/setClientEndpoint.js
shaffeeullah: This is the way all our samples are structured. The region tag is below the function call

samples/system-test/storage.test.js
shaffeeullah: Good catch

samples/setClientEndpoint.js
Breathtender: Oh that makes sense, very good.

ddelgrosso1 (approved):
LGTM

PR #1813, opened 2022-03-14
test: add conformance scenario 7

ddelgrosso1:

### DIFF
--- a/conformance-test/libraryMethods.ts
+++ b/conformance-test/libraryMethods.ts
@@ -288,7 +288,7 @@ export async function rotateEncryptionKey(_bucket: Bucket, file: File) {
 }
 
 export async function saveResumable(_bucket: Bucket, file: File) {
-  const buf = await createTestBuffer(FILE_SIZE_BYTES);
+  const buf = createTestBuffer(FILE_SIZE_BYTES);
   await file.save(buf, {
     resumable: true,
     chunkSize: CHUNK_SIZE_BYTES,
--- a/conformance-test/libraryMethods.ts
+++ b/conformance-test/libraryMethods.ts
@@ -187,7 +187,7 @@ export async function bucketUploadResumable(bucket: Bucket) {
     delete bucket.instancePreconditionOpts.ifMetagenerationMatch;
     bucket.instancePreconditionOpts.ifGenerationMatch = 0;
   }
-  await createTestFileFromBuffer(
+  createTestFileFromBuffer(
     FILE_SIZE_BYTES,
     path.join(__dirname, 'test-data/tmp.dat')
   );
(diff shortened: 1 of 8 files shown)


ddelgrosso1:
Would one of you mind pulling this down and running it a few times? I was seeing some weirdness locally and wouldn't mind a fresh set of eyes.

conformance-test/libraryMethods.ts
```diff
@@ -180,15 +184,18 @@ export async function bucketSetStorageClass(bucket: Bucket) {
 
 export async function bucketUploadResumable(bucket: Bucket) {
   if (bucket.instancePreconditionOpts) {
+    delete bucket.instancePreconditionOpts.ifMetagenerationMatch;
```
danielbankhead, line 187: Perhaps a comment here - I'm not sure what happens if `ifMetagenerationMatch` isn't deleted

src/gcs-resumable-upload/index.ts
```diff
@@ -553,6 +553,19 @@ export class Upload extends Pumpify {
 
   protected async createURIAsync(): Promise<string> {
     const metadata = this.metadata;
+    const headers: gaxios.Headers = {};
```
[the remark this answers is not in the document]
ddelgrosso1: Woops forgot to actually assign it. Thanks.

ddelgrosso1:
I think I see what the problem might be. I am constantly overwriting the same file which might be causing parallel access issues. Fixing this now.

conformance-test/libraryMethods.ts
```diff
@@ -187,15 +192,14 @@ export async function bucketUploadResumable(bucket: Bucket) {
...
-  );
-  await bucket.upload(path.join(__dirname, 'test-data/tmp.dat'), {
+  const filePath = path.join(__dirname, `test-data/tmp-${uuid.v4()}.dat`);
+  createTestFileFromBuffer(FILE_SIZE_BYTES, filePath);
+  await bucket.upload(filePath, {
     resumable: true,
     chunkSize: CHUNK_SIZE_BYTES,
     metadata: {contentLength: FILE_SIZE_BYTES},
   });
+  deleteTestFile(filePath);
```
shaffeeullah, line 202: what happens if the test file is not deleted? do we delete buckets as well? (i dont recall that we do)

ddelgrosso1: This is cleaning up the local file not the file in GCS / emulator. I suppose it is unnecessary as everything will get wiped when the environment is torn down but I figured better to just clean it up anyway.

shaffeeullah:
will re-implement on a more up-to-date branch

PR #1839, opened 2022-03-31
chore(deps): Internalize common

ddelgrosso1:

### DIFF
--- a/src/nodejs-common/operation.ts
+++ b/src/nodejs-common/operation.ts
@@ -1,0 +1,164 @@
+/*!
+ * Copyright 2022 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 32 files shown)


shaffeeullah (approved):

src/nodejs-common/operation.ts
```diff
@@ -0,0 +1,164 @@
...
+  Metadata,
+  MetadataCallback,
+  ServiceObject,
+  ServiceObjectConfig,
+} from './service-object';
+import {ApiError} from './util';
+import {promisify} from 'util';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class Operation<T = any> extends ServiceObject<T> {
```
danielbankhead, line 26: This generic can be "tightened" in a future PR - we can pass an expected interface here

src/nodejs-common/service.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  protected async getProjectIdAsync(): Promise<string> {
+    const projectId = await this.authClient.getProjectId();
+    if (this.projectId === PROJECT_ID_TOKEN && projectId) {
+      this.projectId = projectId;
+    }
+    return this.projectId;
+  }
+
```
danielbankhead, line 174: In the future, for functionality we internalize, we can use a `Promise`-only interface to reduce the API surface

danielbankhead (approved):
LGTM, we should create a follow-up issue for further improvements

PR #1850, opened 2022-04-07
test: add tests for OLM MPU

### LINKED ISSUE #1834: Add support for AbortIncompleteMultipartUpload LifecycleAction
This feature adds the following capabilities to the existing GCS Lifecycle APIs:
- A new type of LifecycleAction -  `AbortIncompleteMultipartUpload` will be added. 

**If there are code changes other than documentation to implement this feature, please add the following system tests:**

- Validate that “AbortIncompleteMultipartUpload” lifecycle action can be set via Bucket Create, Patch and Update operations.
- Bucket Get and List operations should return the correct state of the lifecycle configuration on the bucket which includes the rules with the new action if present.

**If no code changes are required, verify that there are tests proving strings are properly propagated.**

Internal: see [go/gcs-dpe-olm-mpu](http://go/gcs-dpe-olm-mpu) for more info.

Please wait for approval from myself before merging or releasing this change so we can coordinate around release dates.

ddelgrosso1:

Fixes #1834  🦕

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -1296,6 +1296,8 @@ describe('storage', () => {
   });
 
   describe('bucket object lifecycle management', () => {
+    const CUSTOM_TIME_BEFORE = '2020-01-01';
+
     it('should add a rule', done => {
       bucket.addLifecycleRule(
         {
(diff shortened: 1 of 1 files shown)


shaffeeullah:
Do we need this since there are no library changes?

ddelgrosso1:
I guess that depends on how we want to handle these types of changes going forward. I know in our meeting there was some discussion about the value of tests like this. Was a decision reached about whether or not we want to continue adding tests for things like this?

shaffeeullah:
I don't think we want to handle changes like this. Let's make sure we're testing our library and not just testing GCS. I think this just adds bloat.

ddelgrosso1:
As this did not require a code change the tests are not warranted. Closing.

PR #1864, opened 2022-04-11
fix!: default to resumable with no threshold

shaffeeullah:


src/bucket.ts
```diff
@@ -362,8 +362,6 @@ export interface UploadOptions
   destination?: string | File;
   encryptionKey?: string | Buffer;
   kmsKeyName?: string;
-  resumable?: boolean;
-  timeout?: number;
```
ddelgrosso1, line 366: Was this accidentally removed? It is still documented below.

shaffeeullah: not accidentally removed and still documented below because it's still part of the interface. it's declared on `CreateResumableUploadOptions`. I removed this because it's duplicated

ddelgrosso1 (approved):

PR #1867, opened 2022-04-12
samples: Refactor Upload Directory Sample for Improved Reliability

danielbankhead:


samples/uploadDirectory.js
```diff
@@ -36,64 +36,44 @@ async function main(bucketName, directoryPath) {
   // Creates a client
   const storage = new Storage();
 
-  const fs = require('fs');
+  const fs = require('fs/promises');
```
ddelgrosso1, line 39: It looks like before node 14 you need to do `require('fs').promises`

samples/uploadDirectory.js
```diff
@@ -19,7 +19,7 @@
 //   description: Uploads full hierarchy of a local directory to a bucket.
 //   usage: node files.js upload-directory <bucketName> <directoryPath>
 
-async function main(bucketName, directoryPath) {
+function main(bucketName = 'my-bucket', directoryPath = '.') {
```
ddelgrosso1, line 22: Should these default to the old values just for consistency?

danielbankhead: I'm not seeing defaults before this change - I think we require these explicitly or it breaks.

ddelgrosso1: My mistake, more so meant the commented out examples.

samples/uploadDirectory.js
danielbankhead: Thanks, I ran into another issue w/ `require('fs').promises` where `eslint` believes its not available in 10 haha I just used `util.promisify` for now until

ddelgrosso1 (approved):

PR #1871, opened 2022-04-13
fix!: remove configstore

shaffeeullah:
Fixes: https://github.com/googleapis/gcs-resumable-upload/issues/347


src/gcs-resumable-upload.ts
```diff
@@ -887,30 +820,6 @@ export class Upload extends Pumpify {
...
-      // this resumable upload is unrecoverable (bad data or service error).
-      //  -
-      //  https://github.com/googleapis/gcs-resumable-upload/issues/15
-      //  -
-      //  https://github.com/googleapis/gcs-resumable-upload/pull/16#discussion_r80363774
-      if (resp && resp.status === TERMINATED_UPLOAD_STATUS_CODE) {
-        this.restart();
-        return;
-      }
-
```
danielbankhead, line 913: Why is this block getting removed?

danielbankhead (commented):
Such a therapeutic PR

src/gcs-resumable-upload.ts
shaffeeullah: We don't want to retry 404s. This was the case of a stale URI. `TERMINATED_UPLOAD_STATUS_CODE` is not a retryable error code.

danielbankhead:
I think `xdg-basedir` can be removed in this PR too

danielbankhead (approved):

ddelgrosso1 (approved):

rhodgkins:
Thanks @ddelgrosso1, out of interest how do resumable uploads work now? Are they only resumable for the lifetime of the nodejs process?

ddelgrosso1:
@rhodgkins apologies for the delay in responding. Yes that is correct they are only resumable during the lifetime of a process now. We realized during our investigation into whether or not configStore was needed that there existed the possibility of data corruption in the old methodology. ConfigStore was only being used to store the first 16 bytes of an uploaded file which is a poor way to determine if the files are the same. Coupled with the numerous issues with having to cache on disk we made the decision to remove configStore entirely. Now, during the lifetime of a process a resumable upload will resume from the spot of interruption.

PR #1878, opened 2022-04-19
fix: update typescript to 4.6.3 and fix resulting errors

ddelgrosso1:

### DIFF
--- a/system-test/storage.ts
+++ b/system-test/storage.ts
@@ -195,7 +195,9 @@ describe('storage', () => {
             /Could not load the default credentials/,
             /does not have storage\.objects\.create access/,
           ];
-          assert(allowedErrorMessages.some(msg => msg.test(e.message)));
+          assert(
+            allowedErrorMessages.some(msg => msg.test((e as Error).message))
+          );
         }
       });
     });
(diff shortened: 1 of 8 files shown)


system-test/storage.ts
```diff
@@ -3923,7 +3925,8 @@ describe('storage', () => {
       } else {
         return false;
       }
-    } catch (error) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (error: any) {
```
shaffeeullah, line 3929: should this be `any` or `Error`?

shaffeeullah (approved):
One question but otherwise looks good. I see there are linter warnings for two unused variables -- `os` in file.ts and `TERMINATED_UPLOAD_STATUS_CODE` in gcs-resumable-upload. Can you please delete those two fields? That's my bad. I removed the code that used them.

system-test/storage.ts
ddelgrosso1: So you can't actually declare it as an `Error` in the catch block, typescript will only take `unknown` or `any`. I can try one other thing to. try and figure out what type this is so it can be properly casted below.

PR #1905, opened 2022-05-02
build: Streamline `mocha` config

danielbankhead:
Adds support for native `mocha` config support - now we can use `npx mocha` without passing additional args/config

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "predocs": "npm run compile",
     "docs": "jsdoc -c .jsdoc.js",
     "system-test": "mocha system-test --timeout 600000 --exit",
-    "conformance-test": "mocha --parallel --require conformance-test/globalHooks.ts conformance-test/",
+    "conformance-test": "mocha --parallel --require ts-node/register --require conformance-test/globalHooks.ts conformance-test/",
     "test": "c8 mocha test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -26,13 +26,16 @@
     "google storage",
     "storage"
   ],
+  "mocha": {
```
ddelgrosso1, line 29: What does moving this here vs the command line provide?

ddelgrosso1, line 29: What does moving this here vs the command line provide?

danielbankhead: It allows 2 things:
- `npx mocha <path-to-test(s)>` without additional flags to type/remember
- Easier IDE integration - IDEs no long require additional config in order to run `mocha` tests (which can change over time)

danielbankhead:
Additional note/context: Mocha merges the config file (`.mocharc` - owned by owl-bot/synthtool) and `package.json#mocha` (owned by us): https://mochajs.org/#merging

package.json
danielbankhead: Add, if `mocha` is globally installed (`npm i -g mocha`) simply typing `mocha` in the CLI will run the default tests (same as `npx mocha`)

ddelgrosso1 (approved):

PR #1920, opened 2022-05-09
feat: add x-goog-api-client headers for retry metrics

ddelgrosso1:

### DIFF
--- a/src/gcs-resumable-upload.ts
+++ b/src/gcs-resumable-upload.ts
@@ -264,7 +264,11 @@ export class Upload extends Writable {
   contentLength: number | '*';
   retryOptions: RetryOptions;
   timeOfFirstRequest: number;
-  currentInvocationId = uuid.v4();
+  private currentInvocationId = {
+    chunk: uuid.v4(),
+    uri: uuid.v4(),
+    offset: uuid.v4(),
+  };
   private upstreamChunkBuffer: Buffer = Buffer.alloc(0);
   private chunkBufferEncoding?: BufferEncoding = undefined;
   private numChunksReadInRequest = 0;
@@ -532,10 +536,6 @@ export class Upload extends Writable {
 
   protected async createURIAsync(): Promise<string> {
     const metadata = this.metadata;
-    // Create a new invocation id if this is not a retry call
-    if (this.numRetries === 0) {
-      this.currentInvocationId = uuid.v4();
-    }
 
     // Check if headers already exist before creating new ones
     const reqOpts: GaxiosOptions = {
@@ -550,7 +550,7 @@ export class Upload extends Writable {
       ),
       data: metadata,
       headers: {
-        'x-goog-api-client': `gl-node/${process.versions.node} gccl/${packageJson.version} gccl-invocation-id/${this.currentInvocationId}`,
+        'x-goog-api-client': `gl-node/${process.versions.node} gccl/${packageJson.version} gccl-invocation-id/${this.currentInvocationId.uri}`,
       },
     };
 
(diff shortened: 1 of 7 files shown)


src/gcs-resumable-upload.ts
```diff
@@ -27,6 +27,8 @@ import {GoogleAuth, GoogleAuthOptions} from 'google-auth-library';
 import {Readable, Writable} from 'stream';
 import retry = require('async-retry');
 import {RetryOptions, PreconditionOptions} from './storage';
+import * as packageJson from '../package.json';
```
danielbankhead, line 30: I didn't catch this line during our pairing, but I believe this may break [`package.json#files`](https://github.com/googleapis/nodejs-storage/blob/cbe2590186a6ef6229a10b7e62e3c4a1800dccf2/package.json#L13-L16) for consumers (via `npm i @google-cloud/storage`) as `build/package.json` wouldn't be included [[npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files)].

We currently use this pattern for including `package.json` (a little messy): https://github.com/googleapis/nodejs-storage/blob/cbe2590186a6ef6229a10b7e62e3c4a1800dccf2/src/storage.ts#L617-L625

We have a few options:
- Perhaps we can add `build/package.json` to `package.json#files`. Drawback: some tooling may believe `build` in itself is a separate package (as `package.json`) and may misbehave.
- We can use the pattern found in `src/storage`. Drawback: messy-looking.

CC: @bcoe - as I'm working on an easy `package.json` import proposal for Node.js.

test/gcs-resumable-upload.ts
```diff
@@ -67,6 +67,8 @@ const RESUMABLE_INCOMPLETE_STATUS_CODE = 308;
 /** 256 KiB */
 const CHUNK_SIZE_MULTIPLE = 2 ** 18;
 const queryPath = '/?userProject=user-project-id';
+const X_GOOG_API_HEADER_REGEX =
+  /^gl-node\/[0-9]+\.[0-9]+\.[-.\w]+ gccl\/[0-9]+\.[0-9]+\.[-.\w]+ gccl-invocation-id\/[\w]{8}-[\w]{4}-[\w]{4}-[\w]{4}-[\w]{12}$/;
```
danielbankhead, line 71: The `gl-node/...` and `gccl/...` portions may not match for tagged semver preview/beta releases.

Here's a suggestion:

```suggestion
  /^gl-node\/(?<nodeVersion>[^W]+) gccl\/(?<gccl>[^W]+) gccl-invocation-id\/(?<gcclInvocationId>[^W]+)$/
```

Which will allow easy access to named groups for testing:
```ts
const match = X_GOOG_API_HEADER_REGEX.exec(...);

assert(match);

// match.groups.nodeVersion
// match.groups.gccl
// match.groups.gcclInvocationId
```

danielbankhead (changes requested):

test/gcs-resumable-upload.ts
danielbankhead: ^this suggestion is optional

src/gcs-resumable-upload.ts
```diff
@@ -541,7 +549,9 @@ export class Upload extends Writable {
         this.params
       ),
       data: metadata,
-      headers: {},
+      headers: {
+        'x-goog-api-client': `gl-node/${process.versions.node} gccl/${packageJson.version} gccl-invocation-id/${this.currentInvocationId.uri}`,
```
tritone, line 553: Just to clarify-- I spoke with @frankyn and he mentioned that each request in a resumable upload should have its own invocation id (unless it is a retry of a request that failed). Is that reflective of what's happening here?

(Looks like it is from the test cases, but wanted to double check)

danielbankhead (approved):

PR #1921, opened 2022-05-09
chore(deps): update promisify to 3.0

ddelgrosso1:

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,9 +70,9 @@
     "mime-types": "^2.0.8",
     "p-limit": "^3.0.1",
     "pumpify": "^2.0.0",
-    "retry-request": "^4.2.2",
+    "retry-request": "^5.0.0",
     "stream-events": "^1.0.4",
-    "teeny-request": "^7.1.3"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -104,15 +104,15 @@
     "jsdoc-fresh": "^1.0.1",
     "jsdoc-region-tag": "^1.0.2",
     "linkinator": "^2.0.0",
-    "mocha": "^9.2.2",
+    "mocha": "^10.0.0",
```
danielbankhead, line 107: I believe this will break for Node 12 - perhaps this is a typo that isn't caught yet upstream? https://github.com/mochajs/mocha/blob/master/package.json#L47

ddelgrosso1: We did this update for `teeny-request` all the tests passed and nothing complained. I assume because it is a dev dependency.

ddelgrosso1: After some discussion I am reverting `mocha` back to 9.2.2 so it does not bite us in the future.

danielbankhead (approved):

PR #1925, opened 2022-05-12
refactor(deps): Remove get-stream in favor of async iterators

### LINKED ISSUE #1909: refactor(deps): Remove `get-stream` dependency
Remove `get-stream` dependency - it shouldn't be difficult with modern constructs to [consume streams succinctly](https://nodejs.org/api/stream.html#consuming-readable-streams-with-async-iterators) in Node.js.

ddelgrosso1:

Fixes #1909  🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -4104,6 +4104,9 @@ class File extends ServiceObject<File> {
   }
 
   private async getBufferFromReadable(readable: Readable): Promise<Buffer> {
+    readable.once('error', e => {
+      throw e;
+    });
     let buf = Buffer.alloc(0);
     for await (const chunk of readable) {
       buf = Buffer.concat([buf, chunk]);
--- a/src/file.ts
+++ b/src/file.ts
@@ -24,7 +24,6 @@ import {
 import {promisifyAll} from '@google-cloud/promisify';
 
 import compressible = require('compressible');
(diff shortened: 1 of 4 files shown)


src/file.ts
```diff
@@ -4104,6 +4102,18 @@ class File extends ServiceObject<File> {
       this.storage.retryOptions.autoRetry = false;
     }
   }
+
+  private async getBufferFromReadable(readable: Readable): Promise<Buffer> {
+    readable.once('error', e => {
```
ddelgrosso1, line 4107: There seems to be some difference in how node < 16 handles errors with streams / async iterators. Test cases were failing with Node 12 / 14 until I explicitly handled a stream error. However, node 16 handled it with just the iterator.

ddelgrosso1: Broke other things... investigating a better way.

danielbankhead: Yea, this block handled the error a bit differently - the throw wouldn't populate in `getBufferFromReadable`, but would have a separate context as it is in a callback.

I actually don't think the catch is necessary - `for await (const chunk of readable) {` should populate the throw. E.g. https://nodejs.org/api/stream.html#readablesymbolasynciterator

danielbankhead (approved):

PR #1936, opened 2022-05-17
fix: use path.join and __dirname for require package.json

### LINKED ISSUE #1934: Could not resolve "../package.json" on v5.20.1
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

library issue

Not that I can find

No

#### Environment details

  - OS: macOS
  - Node.js version: 14 and 16
  - npm version: 6.14.16 and 8.5.5
  - `@google-cloud/storage` version: v5.20.1

#### Steps to reproduce

  1. Install `@google-cloud/storage` and use it as a dependency in your project
  2. Package your project with esbuild
  3. See that it fails with the following errors:
```
✘ [ERROR] Could not resolve "../package.json"

    node_modules/@google-cloud/storage/build/src/storage.js:384:34:
      384 │             packageJson = require('../package.json');
          ╵                                   ~~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "../../package.json"

    node_modules/@google-cloud/storage/build/src/nodejs-common/util.js:37:26:
      37 │     packageJson = require('../../package.json');
         ╵                           ~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "../package.json"

    node_modules/@google-cloud/storage/build/src/gcs-resumable-upload.js:37:26:
      37 │     packageJson = require('../package.json');
         ╵                           ~~~~~~~~~~~~~~~~~
```

I am using `firebase-admin` which has `@google-cloud/storage` as a dependency. When I pin `@google-cloud/storage` to `5.19.4` I can package my app fine with esbuild. When it is updated to `v5.20.1` it starts to fail with the above error.

ddelgrosso1:

Fixes #1934  🦕

### DIFF
--- a/src/gcs-resumable-upload.ts
+++ b/src/gcs-resumable-upload.ts
@@ -28,6 +28,7 @@ import {Readable, Writable} from 'stream';
 import retry = require('async-retry');
 import {RetryOptions, PreconditionOptions} from './storage';
 import * as uuid from 'uuid';
+import path = require('path');
 
 const NOT_FOUND_STATUS_CODE = 404;
 const TERMINATED_UPLOAD_STATUS_CODE = 410;
@@ -36,10 +37,10 @@ const DEFAULT_API_ENDPOINT_REGEX = /.*\.googleapis\.com/;
 let packageJson: ReturnType<JSON['parse']> = {};
 try {
   // if requiring from 'build' (default)
-  packageJson = require('../../package.json');
+  packageJson = require(path.join(__dirname, '../../package.json'));
 } catch (e) {
   // if requiring directly from TypeScript context
-  packageJson = require('../package.json');
+  packageJson = require(path.join(__dirname, '../package.json'));
 }
 
 export const PROTOCOL_REGEX = /^(\w*):\/\//;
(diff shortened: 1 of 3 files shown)


shaffeeullah (approved):

timdp:
@ddelgrosso1 @shaffeeullah This is unnecessary. `require()` is always relative to the file itself.

Additionally, I don't think it has anything to do with #1934, which it's supposed to close. The issue there is that the import is conditional (through a `try`).

faradaytrs:
@timdp You are right because it didn't fix the problem, now it immediately fails in runtime.

timdp:
It would do, yes. In the case of esbuild, I imagine it leaves the `require()` untouched and then tries to find `package.json` relative to the working directory of the bundle rather than the source file. Both are invalid.

Like I said on the original report, I would attempt to avoid the fallthrough altogether by making sure the source file is always at the same depth, whether or not it's gone through a build step.

PR #1951, opened 2022-05-19
feat!: Use `HashStreamValidator` for Data Validation 

### LINKED ISSUE #1910: refactor!(deps): Internalize crc32c Hashing Functionality
Remove `hash-stream-validation` dependency in favor of an internalized version of this functionality.

**BREAKING**: Drop 'magic import' of `fast-crc32c` module in favor of allowing customers to pass an optional `crc32c` hashing interface.
- This will provide a more customizable, reliable experience while allowing improved ES module & bundling support
- Update documentation to reflect this change; remove references to `fast-crc32c`

danielbankhead:
Fixes #1910  🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -27,8 +27,6 @@ import compressible = require('compressible');
 import * as crypto from 'crypto';
 import * as extend from 'extend';
 import * as fs from 'fs';
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const hashStreamValidation = require('hash-stream-validation');
 import * as mime from 'mime';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const pumpify = require('pumpify');
@@ -68,6 +66,9 @@ import {
   unicodeJSONStringify,
   formatAsUTCISO,
 } from './util';
+import {CRC32CValidatorGenerator} from './crc32c';
+import {HashStreamValidator} from './hash-stream-validator';
+
 import retry = require('async-retry');
 
 export type GetExpirationDateResponse = [Date];
@@ -288,11 +289,12 @@ export const STORAGE_POST_POLICY_BASE_URL = 'https://storage.googleapis.com';
 const GS_URL_REGEXP = /^gs:\/\/([a-z0-9_.-]+)\/(.+)$/;
 
 export interface FileOptions {
+  crc32cGenerator?: CRC32CValidatorGenerator;
   encryptionKey?: string | Buffer;
   generation?: number | string;
   kmsKeyName?: string;
-  userProject?: string;
   preconditionOpts?: PreconditionOptions;
+  userProject?: string;
 }
 
 export interface CopyOptions {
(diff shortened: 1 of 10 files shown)


src/file.ts
```diff
@@ -874,6 +876,9 @@ class File extends ServiceObject<File> {
       pathPrefix: '/acl',
     });
 
+    this.crc32cGenerator =
+      options.crc32cGenerator || this.bucket.crc32cGenerator;
```
ddelgrosso1, line 880: Is there a reason to look to the bucket if one is not passed as an option? Should there be a default within the File class?

src/file.ts
```diff
@@ -1896,10 +1893,7 @@ class File extends ServiceObject<File> {
       let failed = crc32c || md5;
 
       if (crc32c && metadata.crc32c) {
-        // We must remove the first four bytes from the returned checksum.
-        // http://stackoverflow.com/questions/25096737/
-        //   base64-encoding-of-crc32c-long-value
-        failed = !validateStream.test('crc32c', metadata.crc32c.substr(4));
+        failed = !validateStream.test('crc32c', metadata.crc32c);
```
ddelgrosso1, line 1896: Any reason you know of that this was previously removing the first 4 bytes? I think you mentioned it was only checking those bytes?

src/hash-stream-validator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      this.#crc32cHash = crc32cGenerator();
+    }
+
+    if (this.md5Enabled) {
+      this.#md5Hash = createHash('md5');
+    }
+  }
+
+  _flush(callback: () => void) {
```
ddelgrosso1, line 57: Might be good to also just mark these functions with `private`.

danielbankhead: Wanted to, but TS doesn’t like it as the parent class has it `public`

src/file.ts
danielbankhead: Yea… unfortunately we weren’t checking the complete hash before - just the last byte

src/file.ts
danielbankhead: It appears all `File`s must have a bucket (looking at the construction) and all buckets have a generator (whether custom or the default) - this way customers can set things at a storage, bucket, or per-object level

ddelgrosso1 (approved):

src/file.ts
danielbankhead: In short, invalid implementation - see the link from the removed comment:
- http://stackoverflow.com/questions/25096737/base64-encoding-of-crc32c-long-value

This would result in inaccurate validation. E.g.:
Before: https://github.com/googleapis/nodejs-storage/blob/main/test/file.ts#L2373
After: https://github.com/googleapis/nodejs-storage/blob/06e92ac57694ba259379c13383a9f5bd7ed56021/test/file.ts#L2194

The perf for validation is the same - the `update` method is the sensitive portion for performance

PR #1969, opened 2022-05-25
fix: capture and throw on non-existent files

### LINKED ISSUE #1968: v6.0.0 throws exceptions off main thread during upload

This is a bug in @google-cloud/storage client library at v6.0.0.

I could not find an instance.

I'm a Googler.

#### Environment details

  - OS: Reproduces all on OSes
  - Node.js version: v16.14.0
  - npm version: 8.10.0
  - `@google-cloud/storage` version: 6.0.0

#### Steps to reproduce

When uploading an object at a non-existent filepath:

```node
await storage.bucket('bucket').upload('non-existent-file', { destination: 'foo', resumable: true, gzip: true })
```

The error/rejection occurs off the main thread. In the 5.20.5, this issue does not exist.

The easiest way to reproduce this is with a test in mocha (which is how we found it when tests started failing):

```node
try {
  await storage.bucket('bucket').upload('non-existent-file', { destination: 'foo', resumable: true, gzip: true })
} catch(err: unknown) {
  expect(err).to.match('ENOENT')
}
```

Mocha will complain about an exception even though the exception is clearly caught. If you follow the stacktrace, the error is coming from a background stream:

```text
Error: ENOENT: no such file or directory, open 'non-existent-file'
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'non-existent-file'
}
```

The exact same tests with v5.20.5 succeeds without modification.

danielbankhead:
Fixes #1968  🦕

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -62,7 +62,7 @@ import {
   URLSigner,
   Query,
 } from './signer';
-import {Readable} from 'stream';
+import {pipeline, Readable} from 'stream';
 import {CRC32CValidatorGenerator} from './crc32c';
 
 interface SourceObject {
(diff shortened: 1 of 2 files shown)


sethvargo (approved):

sethvargo:
Thanks!

ddelgrosso1 (approved):

PR #1993, opened 2022-06-27
feat: add functionality for passing preconditions at the function level

shaffeeullah:
Adds functionality to pass preconditions at the function level. Also includes a few fixes such as:
- ensuring non-idempotent functions don't retry
- applying preconditions to the correct object if there are two objects involved
Internal ticket: b/239141167


ddelgrosso1:
Since we have more or less worked together on this I think most of the changes here make sense to me. As we discussed it would be nice to get some reviews from others who weren't directly involved.

conformance-test/conformanceCommon.ts
```diff
@@ -121,29 +134,22 @@ export function executeScenario(testCase: RetryTestCase) {
...
-                  file,
-                  notification,
-                  storage,
-                  hmacKey
-                );
-                throw Error(`${storageMethodString} was supposed to throw.`);
-              } catch (e) {
-                assert.notStrictEqual(e, undefined);
-              }
+              await assert.rejects(async () => {
```
ddelgrosso1, line 150: You do not need an async arrow function here, you can simply pass the function we are expecting to reject it does not need to be awaited.

danielbankhead (commented):
Looks good, just 2 things

src/iam.ts
```diff
@@ -344,10 +345,12 @@ class Iam {
       SetPolicyCallback
     >(optionsOrCallback, callback);
 
+    const maxRetries = 0; // We don't support ETag
```
frankyn, line 348: Etag is in the policy object though, does the retry logic check for existance in the JSON structure?

src/iam.ts
```diff
@@ -45,7 +46,7 @@ export interface GetPolicyCallback {
  * @param {string} [userProject] The ID of the project which will be
  *     billed for the request.
  */
-export interface SetPolicyOptions {
+export interface SetPolicyOptions extends PreconditionOptions {
```
frankyn, line 49: I might not understand well enough here, but setIamPolicy doesn't have preconditions: https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy

frankyn (commented):
Two comments, IIUC you're extending existing options with `PreconditionOptions` for all operations.

src/storage.ts
```diff
@@ -261,19 +261,30 @@ const IDEMPOTENCY_STRATEGY_DEFAULT = IdempotencyStrategy.RetryConditional;
...
+    );
+  };
+
   if (err) {
     if ([408, 429, 500, 502, 503, 504].indexOf(err.code!) !== -1) {
       return true;
     }
 
+    if (typeof err.code === 'string') {
+      const reason = (err.code as string).toLowerCase();
```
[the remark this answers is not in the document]
shaffeeullah: without this line it throws errors: `Property 'toLowerCase' does not exist on type 'never'.`

danielbankhead (approved):

shaffeeullah:
@frankyn can you please take another look at this when you have a sec?

frankyn (commented):
I had one more question, otherwise LGTM.

src/nodejs-common/util.ts
```diff
@@ -875,6 +875,7 @@ export class Util {
 
     if (typeof reqOpts.maxRetries === 'number') {
       options.retries = reqOpts.maxRetries;
+      options.noResponseRetries = reqOpts.maxRetries;
```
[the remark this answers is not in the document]
shaffeeullah: `reqOpts.maxRetries` overrides maxRetryValue. It's used in cases where we need to override the user (like forcing no retries because of idempotency)

frankyn (approved):

PR #1995, opened 2022-06-28
fix: correctly handle an empty file in download function

### LINKED ISSUE #1991: Download method hangs on zero byte files
#### Environment details

  - OS: Mac OS 12.4
  - Node.js version: 14.18.2
  - npm version: 6.14.15
  - `@google-cloud/storage` version: 6.1.0

#### Steps to reproduce

  1.  Create an empty file (e.g. `touch emptyfile.txt`) and upload it to cloud storage.
  2. Download the file using the code below

```
const storage = new Storage({ retryOptions: { autoRetry: false } });
storage
  .bucket(BUCKET_NAME)
  .file(emptyfile)
  .download({ destination: destFilename })
  .then(() => console.log("Success"))
  .catch(() => console.error("Error"));
```

When the target file has a size of zero bytes, the promise gets stuck in the pending state. It never resolves or throws an error. If using the method as async rather than a promise then the next line will never get called.

From what I can tell, changing the `autoRetry` options don't seem to make any difference.

Thanks,

Pete

ddelgrosso1:

Fixes #1991  🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2043,14 +2043,27 @@ class File extends ServiceObject<File> {
     delete options.destination;
 
     const fileStream = this.createReadStream(options);
-
+    let receivedData = false;
     if (destination) {
-      fileStream.on('error', callback).once('data', data => {
-        // We know that the file exists the server - now we can truncate/write to a file
-        const writable = fs.createWriteStream(destination);
-        writable.write(data);
-        fileStream.pipe(writable).on('error', callback).on('finish', callback);
-      });
+      fileStream
+        .on('error', callback)
+        .once('data', data => {
+          // We know that the file exists the server - now we can truncate/write to a file
+          receivedData = true;
+          const writable = fs.createWriteStream(destination);
+          writable.write(data);
+          fileStream
+            .pipe(writable)
+            .on('error', callback)
+            .on('finish', callback);
+        })
+        .on('end', () => {
+          // In the case of an empty file no data will be received before the end event fires
+          if (!receivedData) {
+            fs.openSync(destination, 'w');
+            callback(null, Buffer.alloc(0));
+          }
+        });
     } else {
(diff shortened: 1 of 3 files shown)


test/file.ts
```diff
@@ -55,6 +55,7 @@ import {
 } from '../src/file';
 import {ExceptionMessages, IdempotencyStrategy} from '../src/storage';
 import {formatAsUTCISO} from '../src/util';
+import {timeStamp} from 'console';
```
shaffeeullah, line 58: remove unused variable

shaffeeullah (approved):
looks good other than the extra line. thanks, denis!!

test/file.ts
ddelgrosso1: When tab autocomplete goes wrong...

src/file.ts
```diff
@@ -2043,14 +2043,27 @@ class File extends ServiceObject<File> {
...
-
+    let receivedData = false;
     if (destination) {
-      fileStream.on('error', callback).once('data', data => {
-        // We know that the file exists the server - now we can truncate/write to a file
-        const writable = fs.createWriteStream(destination);
-        writable.write(data);
-        fileStream.pipe(writable).on('error', callback).on('finish', callback);
-      });
+      fileStream
```
danielbankhead, line 2048: [Optional]: Now we're on Node 12+ we can use `for await (const chunk of readable)` to make this tinier: https://nodejs.org/api/stream.html#readablesymbolasynciterator

danielbankhead (approved):
LGTM, one optional comment

src/file.ts
ddelgrosso1: Since this isn't within an async function currently I'm going to leave it alone for the time being until we modernize our function signatures.

PR #2000, opened 2022-06-30
fix: force setMetadata calls to use promise version, invoke callbacks manually

ddelgrosso1:

Rationale for this change: 

`setMetadata` is a conditionally idempotent operation. Currently if no precondition is provided we flip a flag to turn off retries. However, since `setMetadata` is promisified we are seeing behavior where the flag is getting flipped back before the operation completes. This is causing incorrect behavior. I am changing the internal calls to wait for the promise to finish before flipping the flag back.

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1301,8 +1301,12 @@ class Bucket extends ServiceObject {
     );
 
     if (options.append === false) {
-      this.setMetadata({lifecycle: {rule: newLifecycleRules}}, callback);
-      this.storage.retryOptions.autoRetry = this.instanceRetryValue;
+      this.setMetadata({lifecycle: {rule: newLifecycleRules}})
+        .then(resp => callback!(null, resp))
+        .catch(callback!)
+        .finally(() => {
+          this.storage.retryOptions.autoRetry = this.instanceRetryValue;
+        });
       return;
     }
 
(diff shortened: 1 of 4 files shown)


src/bucket.ts
```diff
@@ -1301,8 +1301,12 @@ class Bucket extends ServiceObject {
     );
 
     if (options.append === false) {
-      this.setMetadata({lifecycle: {rule: newLifecycleRules}}, callback);
-      this.storage.retryOptions.autoRetry = this.instanceRetryValue;
+      this.setMetadata({lifecycle: {rule: newLifecycleRules}})
+        .then(resp => callback!(null, resp))
```
shaffeeullah, line 1305: will callbacks always be provided? (guessing yes, but just wanted to double check this)

test/file.ts
```diff
@@ -3654,22 +3654,6 @@ describe('File', () => {
       });
     });
 
-    it('should execute callback with error & API response', done => {
```
shaffeeullah, line 3657: For future reference, spoke with @ddelgrosso1 and determined that this case does not exist and that it would be safe to remove this test.

shaffeeullah (commented):
Can you also add a description to this PR to explain why this change is needed? :)

ddelgrosso1:
Marking do not merge until I fix this up and run a bunch of tests to make sure this doesn't break current behavior.

ddelgrosso1:
Been running some tests locally and from what I can tell things look the same as they were. @danielbankhead would you mind giving this PR a look as well?

danielbankhead (approved):
LGTM; maybe in a future refactor we can isolate the method-level retry count from the instance's retry count (perhaps via an optional param or property)

shaffeeullah (approved):

PR #2009, opened 2022-07-29
fix: implement setMetadata in HmacKey and fix associated tests

ddelgrosso1:

### DIFF
--- a/conformance-test/libraryMethods.ts
+++ b/conformance-test/libraryMethods.ts
@@ -680,7 +680,7 @@ export async function deleteHMAC(options: ConformanceTestOptions) {
   const metadata = {
     state: 'INACTIVE',
   };
-  options.hmacKey!.setMetadata(metadata);
+  await options.hmacKey!.setMetadata(metadata);
   await options.hmacKey!.delete();
 }
 
--- a/src/hmacKey.ts
+++ b/src/hmacKey.ts
@@ -20,7 +20,7 @@ import {
   SetMetadataResponse,
 } from './nodejs-common';
 import {SetMetadataOptions} from './nodejs-common/service-object';
-import {Storage} from './storage';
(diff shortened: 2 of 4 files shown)


conformance-test/libraryMethods.ts
```diff
@@ -680,8 +680,8 @@ export async function deleteHMAC(options: ConformanceTestOptions) {
   const metadata = {
     state: 'INACTIVE',
   };
-  options.hmacKey!.setMetadata(metadata);
-  await options.hmacKey!.delete();
+  await hmacKey.setMetadata(metadata);
```
shaffeeullah, line 683: does this not run into the problem where the instructions get used here?

src/hmacKey.ts
```diff
@@ -339,5 +355,57 @@ export class HmacKey extends ServiceObject<HmacKeyMetadata | undefined> {
...
+    metadata: Metadata,
+    options: SetMetadataOptions,
+    callback: MetadataCallback
+  ): void;
+  setMetadata(
+    metadata: Metadata,
+    optionsOrCallback: SetMetadataOptions | MetadataCallback,
+    cb?: MetadataCallback
+  ): Promise<SetMetadataResponse> | void {
+    // ETag preconditions are not currently supported so retries should be disabled
```
shaffeeullah, line 387: unless the retry strategy is `RetryAlways`, right? i made this mistake too in IAM, I will fix it now

conformance-test/libraryMethods.ts
ddelgrosso1: It does not.  I spoke with Cathy and verified in testing that only the delete call will utilize instruction count.

shaffeeullah: oh right because the instructions are on the method specified. this makes sense.

shaffeeullah (approved):
thanks, denis!

PR #2014, opened 2022-08-02
fix(refactor): Add a call from file.delete to the parent class delete

ddelgrosso1:

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -4025,7 +4025,8 @@ class File extends ServiceObject<File> {
       (typeof coreOpts === 'object' &&
         coreOpts?.reqOpts?.qs?.ifGenerationMatch === undefined &&
         localPreconditionOptions?.ifGenerationMatch === undefined &&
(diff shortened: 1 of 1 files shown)


src/file.ts
```diff
@@ -73,7 +73,12 @@ import {HashStreamValidator} from './hash-stream-validator';
 import {URL} from 'url';
 
 import retry = require('async-retry');
-import {SetMetadataOptions} from './nodejs-common/service-object';
+import {
+  DeleteCallback,
+  DeleteOptions,
+  SetMetadataOptions,
+} from './nodejs-common/service-object';
+import * as r from 'teeny-request';
```
shaffeeullah, line 81: what is this import being used for?

ddelgrosso1: It uses the type in the return https://github.com/googleapis/nodejs-storage/pull/2014/files#diff-bc9705d0f7a567399044dfc66ccc82d4d9aa1cff116842a0094d54e463c9ecbcR2056

shaffeeullah (approved):

PR #2022, opened 2022-08-08
refactor(deps): remove stream-events dependency

### LINKED ISSUE #1916: refactor(deps): Remove `stream-events` dependency
The usage of `stream-events` can be internalized using native [streams](https://nodejs.org/api/stream.html) and [events](https://nodejs.org/api/events.html) - such as emitting `writing` on `_write` method (e.g. https://github.com/googleapis/nodejs-storage/pull/1914/commits/f666842ada519c5428a39bac12cefd01d4fcd91b#diff-eee587ca8a340f7ab2aed59173accad2fb1d776f2c176ebda172d1ea1e239f81R390-R396)

ddelgrosso1:

Fixes #1916  🦕

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import * as querystring from 'querystring';
+import {PassThrough} from 'stream';
 
 export function normalize<T = {}, U = Function>(
   optionsOrCallback?: T | U,
@@ -165,4 +166,33 @@ export function formatAsUTCISO(
   }
 
   return resultString;
+}
+
+export class PassThroughShim extends PassThrough {
+  private shouldEmitReading = true;
+  private shouldEmitWriting = true;
+
+  constructor() {
+    super();
+  }
+
+  _read(size: number): void {
+    if (this.shouldEmitReading) {
+      this.emit('reading');
+      this.shouldEmitReading = false;
+    }
+    super._read(size);
+  }
+
+  _write(
+    chunk: never,
+    encoding: BufferEncoding,
+    callback: (error?: Error | null | undefined) => void
+  ): void {
+    if (this.shouldEmitWriting) {
+      this.emit('writing');
+      this.shouldEmitWriting = false;
+    }
+    super._write(chunk, encoding, callback);
+  }
 }
--- a/src/util.ts
+++ b/src/util.ts
@@ -206,6 +206,6 @@ export class PassThroughShim extends PassThrough {
       this.emit('writing');
       this.shouldEmitWriting = false;
     }
-    super._final(callback);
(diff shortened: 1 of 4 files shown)


shaffeeullah (approved):
Looks straightforward enough! @danielbankhead pls take a look as well though

src/util.ts
```diff
@@ -166,3 +167,32 @@ export function formatAsUTCISO(
...
   return resultString;
 }
+
+export class PassThroughShim extends PassThrough {
+  private shouldEmitReading = true;
+  private shouldEmitWriting = true;
+
+  constructor() {
+    super();
+  }
```
danielbankhead, line 177: [Optional] No need for the constructor here - the [default constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor) covers this.

src/util.ts
```diff
@@ -166,3 +167,32 @@ export function formatAsUTCISO(
...
+
+  _read(size: number): void {
+    if (this.shouldEmitReading) {
+      this.emit('reading');
+      this.shouldEmitReading = false;
+    }
+    super._read(size);
+  }
+
+  _write(
```
danielbankhead, line 187: You'll need a `_final` handler in addition to this `_write` handler:
- https://nodejs.org/api/stream.html#writable_finalcallback

danielbankhead (commented):
Looks good so far

src/util.ts
```diff
@@ -166,3 +167,32 @@ export function formatAsUTCISO(
 
   return resultString;
 }
+
+export class PassThroughShim extends PassThrough {
+  private shouldEmitReading = true;
+  private shouldEmitWriting = true;
```
danielbankhead, line 173: Looking at this further, 'reading' and 'writing' are called on every `_read` and `_write` event - to be fully backwards-compatible this class should do the same

ddelgrosso1: Ah good catch, I misread the stream-events / stubs code. Will fix.

ddelgrosso1: Actually I do believe it is only called once: https://github.com/stephenplusplus/stubs/blob/580b4d8b3a34c1193d5ec10e0db0d6ea4b4c0ec5/index.js#L30 it looks like after the number of calls is exhausted we revert to the cached function and stream-events would stop emitting reading / writing.

src/util.ts
ddelgrosso1: Added.

src/util.ts
ddelgrosso1: Removed

danielbankhead (approved):

PR #2023, opened 2022-08-08
fix: revert skip validation

### LINKED ISSUE #709: CONTENT_DOWNLOAD_MISMATCH with successful file download

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/709


src/file.ts
```diff
@@ -1536,23 +1536,17 @@ class File extends ServiceObject<File> {
...
-        // If object is stored compressed (having
-        // metadata.contentEncoding === 'gzip') and was served decompressed,
-        // then skip checksum validation because the remote checksum is computed
-        // against the compressed version of the object.
-        if (!isServedCompressed) {
+        // Did the server decompress object before serving it back to us?
+        const serverDecompressed =
+          headers['x-goog-stored-content-encoding'] === 'gzip' && !isCompressed;
+
+        if (serverDecompressed) {
```
frankyn, line 1543: Shoot, I think you need to remove this code block because it's making a secondary get metadata request. This conditional is meant to go around the validation below.

frankyn (changes requested):

test/file.ts
```diff
@@ -1395,19 +1395,6 @@ describe('File', () => {
         };
       }
 
-      describe('server decompression', () => {
-        it('should skip validation if file was stored compressed', done => {
-          file.metadata.crc32c = '.invalid.';
```
frankyn, line 1400: i think this is still needed no?

frankyn (commented):
One more comment.

system-test/storage.ts
```diff
@@ -2516,33 +2516,6 @@ describe('storage', () => {
...
-              undefined
-            );
-          })
-          .pipe(fs.createWriteStream(tmpFilePath))
-          .on('error', reject)
-          .on('finish', resolve);
-      });
-
-      await file.delete();
-    });
```
danielbankhead, line 2544: I think the opposite of this test should exist

shaffeeullah: this test is actually still needed. i had misunderstood the implementation

shaffeeullah:
@frankyn i think this system test is failing because it's not returning the `'x-goog-stored-content-encoding'` header. is that supposed to happen automatically?

danielbankhead (changes requested):
I think we need some integration tests to associate with the new behavior

src/file.ts
```diff
@@ -1536,31 +1536,16 @@ class File extends ServiceObject<File> {
...
+        const serverDecompressed =
+          headers['x-goog-stored-content-encoding'] === 'gzip' && !isCompressed;
 
         // If we're doing validation, assume the worst-- a data integrity
         // mismatch. If not, these tests won't be performed, and we can assume
         // the best.
         let failed = crc32c || md5;
 
-        if (validateStream) {
+        if (validateStream && !serverDecompressed) {
```
frankyn, line 1548: Would add a comment detailing why we care about `!serverDecompressed`:

> We must check if the server decompressed the data on serve because hash validation is not possible in this case.

frankyn (commented):
One last nit, @shaffeeullah, this will be blocked until the header is no longer allowlist required.

Otherwise, this LGTM! Thank you!

shaffeeullah:
> I think we need some integration tests to associate with the new behavior

There are already integration tests associated with this new boolean.

False case: https://github.com/googleapis/nodejs-storage/blob/0845525a9aebd94da92882ace4cd29bf125641a2/system-test/storage.ts#L2181

True case: https://github.com/googleapis/nodejs-storage/blob/0845525a9aebd94da92882ace4cd29bf125641a2/system-test/storage.ts#L2148

shaffeeullah:
> One last nit, @shaffeeullah, this will be blocked until the header is no longer allowlist required.

What is the timeline for that?

src/file.ts
```diff
@@ -1494,7 +1495,16 @@ class File extends ServiceObject<File> {
...
+        isCompressed = headers['content-encoding'] === 'gzip';
+
+        // The object is safe to validate if:
+        // 1. It was stored gzip and returned to us gzip OR
+        // 2. It was never stored as gzip
+        safeToValidate =
+          (headers['x-goog-stored-content-encoding'] === 'gzip' &&
+            isCompressed) ||
+          headers['x-goog-stored-content-encoding'] === 'identity';
+
```
danielbankhead, line 1507: Which tests validate the following use cases?
- `headers['x-goog-stored-content-encoding'] === 'gzip' && headers['content-encoding'] === 'gzip'`
- `headers['x-goog-stored-content-encoding'] === 'gzip' && headers['content-encoding'] !== 'gzip'`
- `headers['x-goog-stored-content-encoding'] !== 'gzip' && headers['x-goog-stored-content-encoding'] === 'identity'`

shaffeeullah: Good callout. I was looking at it more like tests for `true` `safeToValidate` and tests for `false` safeToValidate`. I'll add additional tests for the specific situations.

shaffeeullah: @danielbankhead added tests for cases 1 & 2, case 3 is covered by existing tests

src/file.ts
```diff
@@ -1562,31 +1572,13 @@ class File extends ServiceObject<File> {
...
-
         // If we're doing validation, assume the worst-- a data integrity
         // mismatch. If not, these tests won't be performed, and we can assume
         // the best.
-        let failed = crc32c || md5;
-
-        if (validateStream) {
+        // We must check if the server decompressed the data on serve because hash
+        // validation is not possible in this case.
+        let failed = (crc32c || md5) && safeToValidate;
```
frankyn, line 1580: I think i got confused with `failed` variable; what does it do?

frankyn (commented):
One more comment; otherwise LGTM, we can't merge this until the feature is available in production to everyone unless y'all are planning making a release until then.

src/file.ts
shaffeeullah: failed defaults to true (if `crc32c || md5`) and then later gets changed to `false` if it passes. (see below lines.) It's not true that it failed if it wasnt `safeToValidate` to begin with, so we need to make sure that's `true` before assuming `failed`

frankyn (approved):

danielbankhead (approved):

shaffeeullah:
@frankyn says this is launched, clear to merge

PR #2029, opened 2022-08-11
fix: Remove `pumpify`

### LINKED ISSUE #1915: refactor(deps): Remove `pumpify` dependency
Native [streams](https://nodejs.org/api/stream.html) should suffice.

danielbankhead:
Fixes #1915 🦕

As a follow-up, we should remove [`duplexify`](https://github.com/mafintosh/duplexify), which is a breaking change as it is exposed as a return type in a few places. In many places, a `Transform` stream makes more sense and is much cleaner.

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -77,7 +77,6 @@ import {
   SetMetadataOptions,
 } from './nodejs-common/service-object';
 import * as r from 'teeny-request';
-import {ReadableStream} from 'stream/web';
 
 export type GetExpirationDateResponse = [Date];
 export interface GetExpirationDateCallback {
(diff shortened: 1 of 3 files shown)


src/file.ts
```diff
@@ -1948,101 +1950,92 @@ class File extends ServiceObject<File> {
...
     }
 
-    // Collect data as it comes in to store in a hash. This is compared to the
-    // checksum value on the returned metadata from the API.
-    const validateStream = new HashStreamValidator({
+    /**
+     * A callback for determining when the underlying pipeline is complete.
+     * It's possible the pipeline callback could error before the write stream
+     * calls `final` so by default this will destroy the write stream unless the
+     * write stream sets this callback via it's `final` handler.
```
shaffeeullah, line 1957: nit: `its`

shaffeeullah (approved):
one small nit, but otherwise looks good! please wait for @ddelgrosso1 to take a look as well. thanks for doing this!!

ddelgrosso1:
Overall nothing stands out to me as troublesome. Perhaps we should move this to a fix and/or feature. We have handled quite a few things during this fixit week and it might be nice to do a release.

PR #2031, opened 2022-08-11
samples: update bucketMetadata

### LINKED ISSUE #2033: samples: update bucketMetadata
Thanks!

The bucketMetadata.js sample incorrectly prints bucket metadata because it does not account for nested Objects. In particular, the `cors`, `iamConfiguration` and `lifecycleRules` properties have their values printed as `Array` or `Object` rather than the contents of the Array/Object. 
I updated to print the metadata as a JSON string instead.

thaaaarun:
The bucketMetadata.js sample was incorrectly printing bucket metadata because it didn't account for nested Objects. I updated to print the metadata as a JSON string instead.


shaffeeullah (approved):
🎉 Thanks, Tharun!

PR #2046, opened 2022-08-18
samples: add preconditions to storage.objects.rewrite, storage.objects.insert, storage.objects.patch, & storage.objects.delete samples

shaffeeullah:


samples/composeFile.js
```diff
@@ -51,7 +52,17 @@ function main(
...
-    await bucket.combine(sources, destinationFileName);
+    // Optional:
+    // Set a generation-match precondition to avoid potential race conditions
+    // and data corruptions. The request to upload is aborted if the object's
+    // generation number does not match your precondition. For a destination
+    // object that does not yet exist, set the ifGenerationMatch precondition to 0
+    // If the destination object already exists in your bucket, set instead a
+    // generation-match precondition using its generation number.
+    const combineOptions = {
+      ifGenerationMatch: generationMatchPrecondition,
```
ddelgrosso1, line 63: Nit: even though it is called out in the comment it might make sense to name this variable something like `destinationGenerationMatchPrecondition` just so it is abundantly clear.

samples/copyFile.js
```diff
@@ -49,11 +50,26 @@ function main(
...
+    // Optional:
+    // Set a generation-match precondition to avoid potential race conditions
+    // and data corruptions. The request to upload is aborted if the object's
+    // generation number does not match your precondition. For a destination
+    // object that does not yet exist, set the ifGenerationMatch precondition to 0
+    // If the destination object already exists in your bucket, set instead a
+    // generation-match precondition using its generation number.
+    const copyOptions = {
+      preconditionOpts: {
+        ifGenerationMatch: generationMatchPrecondition,
```
ddelgrosso1, line 64: Same comment regarding the name.

samples/copyOldVersionOfFile.js
```diff
@@ -53,12 +54,26 @@ function main(
...
+    // Optional:
+    // Set a generation-match precondition to avoid potential race conditions
+    // and data corruptions. The request to upload is aborted if the object's
+    // generation number does not match your precondition. For a destination
+    // object that does not yet exist, set the ifGenerationMatch precondition to 0
+    // If the destination object already exists in your bucket, set instead a
+    // generation-match precondition using its generation number.
+    const copyOptions = {
+      preconditionOpts: {
+        ifGenerationMatch: generationMatchPrecondition,
```
ddelgrosso1, line 67: Probably should have called this out above but same comment for all the places that the precondition applies to the destination.

shaffeeullah: updated this one and move. Good callout.

ddelgrosso1 (approved):
lgtm

PR #2056, opened 2022-09-01
fix: truncated `createReadStream` through early `end` event

### LINKED ISSUE #2044: v6.4.1 seems to introduce a bug with gzipped files
#### Environment details

  - OS: Cloud Run
  - Node.js version: 18.3.0 (ALPINE: 18.3-alpine3.15 Docker image)
  - npm version: 8.12.1
  - `@google-cloud/storage` version: 6.4.1

#### Steps to reproduce

I have a log file processing pipeline on Cloud Run/Node which is pretty well monitored and managed via CI/CD (GitHub -> Cloud Build -> Cloud Run). Dependabot raised a PR for 6.4.1 of this lib which passed all my tests and ran fine overnight on pre-live so I merged it which deploys to live. Immediately, the number of log files being rejected by the pipeline jumped and I saw that the Node app was logging errors:
`gzip error: unexpected end of file (code: Z_BUF_ERROR)` 
The processing time per file also jumped massively (by an order of magnitude or 2).
Once I reverted this version of the lib, the errors disappeared and processing time returned to normal.

The Node app Does this:
1. Receives a "new GCS file" push from Eventarc (HTTP lib is [Fastify](https://www.npmjs.com/package/fastify))
2. Streams the new file through the processing pipeline, using [gunzip-maybe](https://www.npmjs.com/package/gunzip-maybe)
3. Streams resultant data into BigQuery

So I think the usage of this lib is relatively simple - i am only using a few methods (via async/await):
* setMetadata()
* exists()
* bucket()
* file()
* delete()
* createReadStream() - probably the most likely cuplrit here, i guess

Please let me know if you have more Qs. Sorry this isn't a simple set of steps, I thought ^ was more practical in this case.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

### LINKED ISSUE #2055: Files occasionally being truncated when downloading
#### Issue
We're seeing occasionally that files are being truncated when they are downloaded without the storage library throwing an error.
What we're trying to do is concurrently pull down a set of files from a location to then process them. We also exclude any 0 byte files when we download. We create multiple pubsub subscriptions when doing this as the node version doesn't support concurrent messaging in pubsub in the way we need it to.
We're seeing this happening around 5% of the time we download files.

#### Environment details

  - OS: Windows
  - Node.js version: 16.16.0
  - npm version: 8.11.0
  - `@google-cloud/storage` version: 6.2.0

#### Steps to reproduce

Difficult to explain so including a snippet of what we're doing here. We're using storage alongside the pubsub library so including the code for that, as we have a suspicion that might be causing some of the issues.

```
const pubsub = new PubSub({ projectId });
const storage = new Storage({ projectId });
const subscriptions = [...Array(5).keys()].map(() => pubsub.subscription(validReleaseSub, {
  ackDeadline: 600,
  flowControl: {
    allowExcessMessages: false,
    maxMessages: 5
  }
}));

for(const subscription of subscriptions) {
  subscription.on('message', async (message) => {
   const bucket = storage.bucket(message.bucket);
   const [files] = await bucket.getFiles({ prefix: `${message.location}/` });

   await Promise.all(
     files.filter(file => file.metadata.size != 0).map(async file => {
      const destination = file.name.replace(message.location, destinationPath);
      await file.download({destination});
    });
   );

  // process the files
  });
}
```

danielbankhead:
Fixes #2044 🦕
Fixes #2055 🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1417,9 +1417,7 @@ class File extends ServiceObject<File> {
     // Authenticate the request, then pipe the remote API request to the stream
     // returned to the user.
     const makeRequest = () => {
-      const query = {
-        alt: 'media',
-      } as FileQuery;
+      const query: FileQuery = {alt: 'media'};
 
       if (this.generation) {
         query.generation = this.generation;
@@ -1460,8 +1458,7 @@ class File extends ServiceObject<File> {
         })
         .on('response', res => {
           throughStream.emit('response', res);
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          util.handleResp(null, res, null, onResponse as any);
+          util.handleResp(null, res, null, onResponse);
         })
         .resume();
 
@@ -1525,7 +1522,7 @@ class File extends ServiceObject<File> {
         const handoffStream = new PassThrough({
           final: async cb => {
             // Preserving `onComplete`'s ability to
-            // close `throughStream` before pipeline
+            // destroy `throughStream` before pipeline
             // attempts to.
             await onComplete(null);
             cb();
@@ -1558,7 +1555,6 @@ class File extends ServiceObject<File> {
         }
 
         if (rangeRequest || !shouldRunValidation) {
-          throughStream.end();
           return;
         }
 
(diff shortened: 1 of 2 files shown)


shaffeeullah (approved):
LGTM - thanks Dan!

PR #2062, opened 2022-09-08
feat: add multiple lifecycle rules

### LINKED ISSUE #2060: addLifecycleRule cannot be called concurrently
Starting calls to `addLifecycleRule` concurrently will likely result in subsequent calls overwriting metadata from earlier calls and losing some of the added rules ([src](https://github.com/googleapis/nodejs-storage/blob/5007542a33bc1cc1f50357101adbf216af7f35fb/src/bucket.ts#L1386-L1396)).

This is surprising behavior based on the interface and docs/comments. If the APIs don't support this, I think the documentation and examples could be improved; the interface could also accept an array as a recommended way to add multiple rules.

#### Environment details

  - OS: mac, linux
  - Node.js version: 16
  - npm version: 8.5
  - `@google-cloud/storage` version: 6.4.1

#### Steps to reproduce

  1. Create a bucket
  2. Given two rule objects r1 and r2, `await Promise.all([bucket.addLifecycleRule(r1), bucket.addLifecycleRule(r2)])`

Observed: there may only be one rule created.

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/2060

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1194,7 +1194,7 @@ class Bucket extends ServiceObject {
    * See {@link https://cloud.google.com/storage/docs/lifecycle| Object Lifecycle Management}
    * See {@link https://cloud.google.com/storage/docs/json_api/v1/buckets/patch| Buckets: patch API Documentation}
    *
-   * @param {LifecycleRule | [LifecycleRule]} rule The new lifecycle rule or rules to be added to objects
+   * @param {LifecycleRule|[LifecycleRule]} rule The new lifecycle rule or rules to be added to objects
    *     in this bucket.
    * @param {string|object} rule.action The action to be taken upon matching of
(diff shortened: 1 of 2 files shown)


src/bucket.ts
```diff
@@ -1339,7 +1343,11 @@ class Bucket extends ServiceObject {
 
     options = options || {};
 
-    const newLifecycleRules = arrify(rule).map(rule => {
+    if (!Array.isArray(rule)) {
+      rule = arrify(rule);
+    }
+
+    const newLifecycleRules = rule.map(rule => {
```
danielbankhead, line 1350: I think `rules` might be clearer here (and we can drop `arrify`):

```suggestion
    const rules = Array.isArray(rule) ? rule : [rule];

    const newLifecycleRules = rules.map(rule => {
```

danielbankhead (approved):
LGTM, 1 optional suggestion

PR #2070, opened 2022-09-20
fix: correct STORAGE_EMULATOR_HOST handling (#2069, #1314)

### LINKED ISSUE #1314: baseurl is not quite right when using STORAGE_EMULATOR_HOST
I think this [line](https://github.com/googleapis/nodejs-storage/blob/d854c2609eb87886a010e726daa5f1339ee9ea38/src/storage.ts#L404) should read:

``` js
    const baseUrl = `${EMULATOR_HOST}/storage/v1` || `${options.apiEndpoint}/storage/v1`;
```

I guess the intend was to have configurable path, not only the host, but it breaks either bucket api or the object api.

Currently, with the default url, we get 
``` js
config = {
 apiEndpoint: 'https://storage.googleapis.com',
 baseUrl: 'https://storage.googleapis.com/storage/v1',
 //...
}
```
while with a `EMULATOR_HOST= https://exmaple.com` we get 
``` js
config = {
 apiEndpoint: 'https://exmaple.com',
 baseUrl: 'https://exmaple.com',
 //...
}
```

As a result, when EMULATOR_HOST is set with `https://exmaple.com` I can create buckets (the url starts with `/storage/v1`), but can't upload/download file (url starts with `/upload/storage/v1`). If I set EMULATOR_HOST to https://exmaple.com/v1/storage/ I have the inverse problem (can't create bucket, but can upload/download)

### LINKED ISSUE #2069: STORAGE_EMULATOR_HOST not handled correctly
This is a rehash of #1314, which was somehow "fixed" by adding comments internal to the implementation.

#### Environment details

  - OS: Linux
  - Node.js version: 14.x
  - npm version: doesn't matter
  - `@google-cloud/storage` version: 5.8.5

#### Steps to reproduce

  1. Set STORAGE_EMULATOR_HOST to http://localhost:NNN
  2. Try to use the library

As described in #1314, the distinction in the Storage constructor between `apiEndpoint` and `baseUrl` is not handled correctly when this environment variable is used. There is no possible value for this environment variable that results in a fully functioning client.

The proposed fix in that issue also seems to work just fine, so will submit a PR

mgabeler-lee-6rs:
credit to @jpambrun for identifying the fix

Test assertions needed to change slightly since they were asserting the
presence of functionality that doesn't actually work, specifically allowing
one to change the base url to not have `/storage/v1`. The code fundamentally
does not support this at all, and is full of places that are hard coded to
always include that path element. The code _does_ support having a base path
_before_ that however, and so that is what I updated the tests to check.

Fixes #1314
Fixes #2069

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -666,7 +666,7 @@ export class Storage extends Service {
     options = Object.assign({}, options, {apiEndpoint});
 
     // Note: EMULATOR_HOST is an experimental configuration variable. Use apiEndpoint instead.
-    const baseUrl = (EMULATOR_HOST || options.apiEndpoint) + '/storage/v1';
+    const baseUrl = new URL('/storage/v1', EMULATOR_HOST || options.apiEndpoint).href;
 
     const config = {
       apiEndpoint: options.apiEndpoint!,
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -665,8 +665,12 @@ export class Storage extends Service {
 
     options = Object.assign({}, options, {apiEndpoint});
 
-    // Note: EMULATOR_HOST is an experimental configuration variable. Use apiEndpoint instead.
-    const baseUrl = new URL('/storage/v1', EMULATOR_HOST || options.apiEndpoint).href;
+    // Note: EMULATOR_HOST, if present and not overridden, has been applied to
+    // `options` at this point. Also, this uses string concatenation because the
+    // endpoint may contain a base path, and any trailing slash on that will
+    // have been removed, so using the two-arg URL constructor for relative path
+    // resolution won't work.
+    const baseUrl = new URL(options.apiEndpoint + '/storage/v1').href;
 
     const config = {
       apiEndpoint: options.apiEndpoint!,
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -666,7 +666,7 @@ export class Storage extends Service {
     options = Object.assign({}, options, {apiEndpoint});
 
     // Note: EMULATOR_HOST is an experimental configuration variable. Use apiEndpoint instead.
-    const baseUrl = EMULATOR_HOST || `${options.apiEndpoint}/storage/v1`;
+    const baseUrl = (EMULATOR_HOST || options.apiEndpoint) + '/storage/v1';
 
     const config = {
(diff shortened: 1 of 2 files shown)


src/storage.ts
```diff
@@ -666,7 +666,7 @@ export class Storage extends Service {
     options = Object.assign({}, options, {apiEndpoint});
 
     // Note: EMULATOR_HOST is an experimental configuration variable. Use apiEndpoint instead.
-    const baseUrl = EMULATOR_HOST || `${options.apiEndpoint}/storage/v1`;
+    const baseUrl = (EMULATOR_HOST || options.apiEndpoint) + '/storage/v1';
```
danielbankhead, line 669:
```suggestion
    const baseUrl = new URL('/storage/v1', EMULATOR_HOST || options.apiEndpoint).href;
```

danielbankhead (approved):
Looks good - one minor, optional suggestion.

We have a code freeze for the next few weeks - will merge this after 🙂

src/storage.ts
mgabeler-lee-6rs: Hmm, I think this broke the tests, looking for a better way...

mgabeler-lee-6rs: Couple things going wrong when this suggestion is applied:
1. Passing `/storage/v1` as the first arg discards any path info from the input, which the tests assert must be retained
2. This maybe could be improved by passing `./storage/v1` instead, but that still doesn't work due to the calls to `Storage.sanitizeEndpoint` above, which trim any trailing slash, which makes that relative path resolution still not work right.
3. The tests assert that a URL w/o a protocol should have the default `https` protocol inferred, but that doesn't work when passing through the `new URL()` constructor, because it reports the base url is invalid

I think I have a simple fix for this, but it runs into a weird issue where some tests are asserting odd mixes of what the test title says, and the _opposite_ of what it says.

I've pushed a change that applies that "simple fix" and makes the tests more internally consistent, but I wonder if that inconsistency was somehow intentional for some pre-existing use cases? I've left some comments in the code on the odd bits.

danielbankhead (approved):

src/storage.ts
danielbankhead: Thanks for looking into this

PR #2074, opened 2022-09-21
fix: Determine `Content-Length` Before Attempting Multi-chunk Upload

### LINKED ISSUE #2073: Multi-chunk Upload Without `contentLength` Can Hang When `fileSize % chunkSize != 0`
A multi-chunk resumable upload hangs when the `chunkSize` is not evenly divisible by the file size when a `contentLength` is not provided.

ACs:
- Fix upload logic for this scenario
- Fix unit test(s)
- Add integration tests for the following scenarios:
  - `fileSize % chunkSize != 0`
  - `fileSize < chunkSize`

danielbankhead:
Fixes #2073  🦕

### DIFF
--- a/src/resumable-upload.ts
+++ b/src/resumable-upload.ts
@@ -725,9 +725,8 @@ export class Upload extends Writable {
 
       const bytesToUpload = value!.chunk.byteLength;
 
-      // Important: we want to know if the upstream has ended and we shouldn't expect any more
-      // before unshifting data back into the queue. This way we will know if this is the last
-      // request or not.
+      // Important: we want to know if the upstream has ended and the queue is empty before
+      // unshifting data back into the queue. This way we will know if this is the last request or not.
       const isLastChunkOfUpload = !(await this.waitForNextChunk());
 
       // Important: put the data back in the queue for the actual upload iterator
--- a/src/resumable-upload.ts
+++ b/src/resumable-upload.ts
@@ -720,21 +720,34 @@ export class Upload extends Writable {
     // If using multiple chunk upload, set appropriate header
     if (multiChunkMode) {
       // We need to know how much data is available upstream to set the `Content-Range` header.
-      const result = await this.upstreamIterator(
-        expectedUploadSize,
-        true
-      ).next();
+      const oneChunkIterator = this.upstreamIterator(expectedUploadSize, true);
+      const {value} = await oneChunkIterator.next();
 
-      const bytesToUpload = result.value!.chunk.byteLength;
-      // Important: put the data back in the queue for the actual upload
-      this.unshiftChunkBuffer(result.value!.chunk);
+      const bytesToUpload = value!.chunk.byteLength;
+
+      // Important: we want to know if the upstream has ended and we shouldn't expect any more
+      // before unshifting data back into the queue. This way we will know if this is the last
+      // request or not.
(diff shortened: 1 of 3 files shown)


src/resumable-upload.ts
```diff
@@ -718,13 +718,37 @@ export class Upload extends Writable {
...
+      // `- 1` as the ending byte is inclusive in the request.
+      const endingByte = bytesToUpload + this.numBytesWritten - 1;
+
+      // `Content-Length` for multiple chunk uploads is the size of the chunk,
+      // not the overall object
+      headers['Content-Length'] = bytesToUpload;
       headers[
         'Content-Range'
-      ] = `bytes ${this.offset}-${endingByte}/${this.contentLength}`;
+      ] = `bytes ${this.offset}-${endingByte}/${totalObjectSize}`;
```
frankyn, line 751: this def looks like a bug in that the content-length may not match the totalObjectSize; I'm guessing this is how this issue came up?

frankyn: Alternatively, you don't need to supply `totalObjectSize` and could instead use `*` and only use `totalObjectSize` once it's known to support streams of unknown byte length.

frankyn (changes requested):

src/resumable-upload.ts
danielbankhead: Unfortunately it errors when we use `*` on the final chunk in a multiple chunk transfer (see this commit https://github.com/googleapis/nodejs-storage/pull/2074/commits/53aa25dc69dd0304cdd8b8ae0bdd2dd6bf4d0fed). 

Which makes sense for 2 reasons:
- the final chunk could  be smaller than the minimum chunk size threshold ([256 KiB](https://cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload))
- it's unclear how the server would know this is the last chunk and we're done uploading

frankyn: Why not buffer up to multiple of 256kib?

danielbankhead: We do, using the user-provided `chunkSize` property. The last chunk in the upload maybe be less than that (e.g. uploading a 257KiB file in 256KiB chunks).

frankyn: interesting, I would expect the following:

Chunk 1: 256KiB
Chunk 2 (Also Finalize) 1KiB

Does this not work as expected or am I missing context?

danielbankhead: Unfortunately that does not work. The server doesn't know the 2nd chunk is the last chunk with `*` as the end range for multi-chunk uploads (single chunk upload is fine with `*`).

danielbankhead: See this error:
Screen Shot 2022-09-21 at 5 02 45 PM

frankyn: ah, I was saying you need to swap the `*` with the complete object size on the final chunk otherwise you'll get that error as expected.

frankyn (approved):
Discussed the implementation in detail with @danielbankhead.

The PR LGTM, it introduces not only unit tests but also system tests which validate critical multi chunk use cases which were not being tested before against the live service.

Complexity of resumable media API with streams caused the following use cases to not be captured: https://github.com/googleapis/nodejs-storage/pull/2074/commits/aa803746f63c0461b1c6fcbad690fb64554b94b5
The tests that failed in this commit now pass as expected.

PR #2079, opened 2022-09-30
refactor(deps): remove arrify

### LINKED ISSUE #2065: refactor(deps): Remove `arrify` dependency in favor of native implementation
In most places we use `arrify` to uniform logic for `type` and `type[]`. Example:

https://github.com/googleapis/nodejs-storage/blob/fbe2deb72bd98db54a03cf228a360a871ba1915b/src/bucket.ts#L2129-L2134

Using `Array.isArray` alone should suffice rather than using this dependency. Example:

https://github.com/googleapis/nodejs-storage/blob/fbe2deb72bd98db54a03cf228a360a871ba1915b/src/bucket.ts#L1346-L1348

shaffeeullah:
Fixes https://github.com/googleapis/nodejs-storage/issues/2065

### DIFF
--- a/src/acl.ts
+++ b/src/acl.ts
@@ -18,7 +18,6 @@ import {
   Metadata,
 } from './nodejs-common';
 import {promisifyAll} from '@google-cloud/promisify';
-import arrify = require('arrify');
 
 export interface AclOptions {
   pathPrefix: string;
@@ -742,7 +741,7 @@ class Acl extends AclRoleAccessorMethods {
         let results;
 
         if (resp.items) {
-          results = arrify(resp.items).map(this.makeAclObject_);
+          results = resp.items.map(this.makeAclObject_);
         } else {
           results = this.makeAclObject_(resp);
         }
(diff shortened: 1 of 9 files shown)


src/acl.ts
```diff
@@ -742,7 +741,7 @@ class Acl extends AclRoleAccessorMethods {
         let results;
 
         if (resp.items) {
-          results = arrify(resp.items).map(this.makeAclObject_);
+          results = resp.items.map(this.makeAclObject_);
```
ddelgrosso1, line 744: Just want to check that resp.items is always an array?

shaffeeullah: yep. an array or undefined. we already check for undefined above.

test/index.ts
```diff
@@ -63,7 +62,6 @@ const fakePaginator = {
         return;
       }
 
-      methods = arrify(methods);
```
ddelgrosso1, line 66: If I read this correctly it will turn methods into an array of arrays, doesn't look like this change had any impact but just want to double check.

shaffeeullah: it won't change it into an array of arrays. `arrify(["hello"]) = ["hello"]` https://github.com/sindresorhus/arrify

ddelgrosso1 (approved):

PR #2094, opened 2022-10-24
fix: remove async from final function which causes double invocation …

### LINKED ISSUE #2082: ERR_MULTIPLE_CALLBACK on file.download()
yes

  - Search the issues already opened: https://github.com/googleapis/nodejs-storage/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
no

no

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: Win10 21H2
  - Node.js version: 16.13.2
  - npm version: 8.1.2
  - `@google-cloud/storage` version: 6.5.2

#### Steps to reproduce

  Try to download a file:
```
  try {
    await file.download({
      destination: destination
    })
  } catch(e) {
    console.log(e);
  }
```

As the result catching error ERR_MULTIPLE_CALLBACK

ddelgrosso1:
…of callback

Fixes #2082 🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1534,9 +1534,11 @@ class File extends ServiceObject<File> {
             // Preserving `onComplete`'s ability to
             // destroy `throughStream` before pipeline
             // attempts to.
-            onComplete(null).then(() => {
-              cb();
-            });
+            onComplete(null)
+              .then(() => {
+                cb();
+              })
+              .catch(cb);
           },
         });
 
--- a/src/file.ts
+++ b/src/file.ts
@@ -1530,12 +1530,13 @@ class File extends ServiceObject<File> {
         }
 
         const handoffStream = new PassThrough({
-          final: async cb => {
+          final: cb => {
             // Preserving `onComplete`'s ability to
             // destroy `throughStream` before pipeline
             // attempts to.
-            await onComplete(null);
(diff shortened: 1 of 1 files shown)


ddelgrosso1:
I believe this is related to https://github.com/nodejs/node/issues/39535 which can cause double invocation of callbacks under the same situation we had in the code here.

danielbankhead (approved):

src/file.ts
```diff
@@ -1530,12 +1530,15 @@ class File extends ServiceObject<File> {
...
             // Preserving `onComplete`'s ability to
             // destroy `throughStream` before pipeline
             // attempts to.
-            await onComplete(null);
-            cb();
+            onComplete(null)
+              .then(() => {
+                cb();
+              })
+              .catch(cb);
```
julienw, line 1541: couldn't this cause a double invocation if `cb` throws?

I think I would have used something such as:
```
.then(
  () => cb(),
  () => cb()
)
```

Also I'm not entirely sure how the previous code would end up with a double invocation, but I may miss some context.

PR #2140, opened 2023-02-02
fix: correctly handle if a user has no permissions when calling iam.t…

### LINKED ISSUE #2139: bucket.iam.testPermissions can kill process
#### Environment details

  - OS: MacOS
  - Node.js version: `18.12`
  - `@google-cloud/storage` version: `6.9.0`

#### Steps to reproduce

1. Create a service account that has no access to a storage bucket (ex. `no-access-bucket`)
    - It may need access to a different bucket in the same project, unsure
2. Run the following:
    ```ts
     // Make sure to use correct service account
    const bucket = storage.bucket("no-access-bucket");
    await bucket.iam.testPermissions([
      "storage.objects.get",
      "storage.objects.list",
      "storage.objects.create",
      "storage.objects.delete",
    ]);
    ```
3. Receive an error (cannot call `.indexOf` of undefined)
    - In my case, the error was _NOT_ successfully caught by a try/catch block
    - It actually terminated my whole remix dev server (this could be a remix quirk, or unrelated mistake of mine 😅)
    
    
I solved it by making the http request directly with this snippet:
```ts
    const [responseBody]: [{ permissions?: string[] }, any] =
      await bucket.request({
        qs: {
          permissions: requiredPermissions,
        },
        uri: "/iam/testPermissions",
        useQuerystring: true,
      });
```

The issue appears to be that `testPermissions` expects the http response json to _always_ include a `permissions` array. However, the behaviour I observed is that the response does not include the `permissions` key at all when the array would otherwise be empty.

That means line `477` will throw at the `.indexOf`:

https://github.com/googleapis/nodejs-storage/blob/9741a7a2acb2d4b75020d79e34a7d661ab866c63/src/iam.ts#L461-L483

ddelgrosso1:
…estPermissions

Fixes #2139  🦕

### DIFF
--- a/src/iam.ts
+++ b/src/iam.ts
@@ -470,7 +470,9 @@ class Iam {
           return;
         }
 
-        const availablePermissions = resp.permissions;
+        const availablePermissions = Array.isArray(resp.permissions)
+          ? resp.permissions
+          : [];
 
         const permissionsHash = permissionsArray.reduce(
           (acc: {[index: string]: boolean}, permission) => {
--- a/test/iam.ts
+++ b/test/iam.ts
@@ -240,6 +240,28 @@ describe('storage/iam', () => {
       );
     });
 
+    it('should return false for supplied permissions if user has no permissions', done => {
+      const permissions = ['storage.bucket.list', 'storage.bucket.consume'];
+      const apiResponse = {permissions: undefined};
+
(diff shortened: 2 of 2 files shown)


shaffeeullah (approved):
thanks, denis!!

qw-in (approved):
Thanks for the quick fix :thumbsup:

PR #2153, opened 2023-02-17
fix: refactor createReadStream to remove unnecessary stream

ddelgrosso1:

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -1915,7 +1915,6 @@ class File extends ServiceObject<File> {
       crc32c,
       md5,
       crc32cGenerator: this.crc32cGenerator,
-      updateHashesOnly: true,
     });
 
     const fileWriteStream = duplexify();
--- a/src/file.ts
+++ b/src/file.ts
@@ -394,7 +394,7 @@ export interface SetStorageClassCallback {
   (err?: Error | null, apiResponse?: Metadata): void;
 }
 
-class RequestError extends Error {
+export class RequestError extends Error {
   code?: string;
   errors?: Error[];
 }
@@ -1383,14 +1383,14 @@ class File extends ServiceObject<File> {
       typeof options.start === 'number' || typeof options.end === 'number';
     const tailRequest = options.end! < 0;
 
-    let validateStream: HashStreamValidator | undefined = undefined;
+    let validateStream: HashStreamValidator = new HashStreamValidator();
 
     const throughStream = new PassThroughShim();
+    const hashes: {crc32c?: string; md5?: string} = {};
 
     let isCompressed = true;
     let crc32c = true;
     let md5 = false;
-    let safeToValidate = true;
 
     if (typeof options.validation === 'string') {
       const value = options.validation.toLowerCase().trim();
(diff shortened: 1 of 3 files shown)


ddelgrosso1:
Summary of things change: 1. De-nest all the callbacks in `createReadStream` so it is a bit easier to read 2. Move validation logic into `HashStreamValidator` 3. Remove `HandoffStream` which was only functioning to delay validation until all data was available.

shaffeeullah (approved):
thanks, denis!!

src/hash-stream-validator.ts
```diff
@@ -20,27 +20,31 @@ import {
...
 } from './crc32c';
+import {FileExceptionMessages, RequestError} from './file';
 
 interface HashStreamValidatorOptions {
   crc32c: boolean;
   md5: boolean;
   crc32cGenerator: CRC32CValidatorGenerator;
+  crc32cExpectedValue?: string;
+  md5ExpectedValue?: string;
+  updateHashesOnly?: boolean;
```
danielbankhead, line 31:
```suggestion
  /** Enables CRC32C calculation. To validate a provided value use `crc32cExpected`. */
  crc32c: boolean;
  /** Enables MD5 calculation. To validate a provided value use `md5Expected`. */
  md5: boolean;
  /** Set a custom CRC32C generator */
  crc32cGenerator: CRC32CValidatorGenerator;
  /** Sets the expected CRC32C value to verify once all data has been consumed. Also sets the `crc32c` option to `true` */
  crc32cExpected?: string;
  /** Sets the expected MD5 value to verify once all data has been consumed. Also sets the `md5` option to `true` */
  md5Expected?: string;
```

src/file.ts
```diff
@@ -1415,6 +1415,97 @@ class File extends ServiceObject<File> {
...
+        this.getBufferFromReadable(rawResponseStream).then(body => {
+          err.message = body.toString('utf8');
+          throughStream.destroy(err);
+        });
+
+        return;
+      }
+
+      const headers = rawResponseStream.toJSON().headers;
+      isCompressed = headers['content-encoding'] === 'gzip';
```
danielbankhead, line 1449:
```suggestion
      const isCompressed = headers['content-encoding'] === 'gzip';
      const hashes: {crc32c?: string; md5?: string} = {};
```

src/file.ts
```diff
@@ -1386,11 +1386,11 @@ class File extends ServiceObject<File> {
     let validateStream: HashStreamValidator | undefined = undefined;
 
     const throughStream = new PassThroughShim();
+    const hashes: {crc32c?: string; md5?: string} = {};
 
     let isCompressed = true;
```
danielbankhead, line 1391:
```suggestion
```

src/file.ts
```diff
@@ -1386,11 +1386,11 @@ class File extends ServiceObject<File> {
     let validateStream: HashStreamValidator | undefined = undefined;
 
     const throughStream = new PassThroughShim();
+    const hashes: {crc32c?: string; md5?: string} = {};
```
danielbankhead, line 1389:
```suggestion
```

src/file.ts
```diff
@@ -1415,6 +1415,97 @@ class File extends ServiceObject<File> {
...
+              hashes[hashType as 'crc32c' | 'md5'] = hashValue;
+            });
+        }
+
+        validateStream = new HashStreamValidator({
+          crc32c,
+          md5,
+          crc32cGenerator: this.crc32cGenerator,
+          crc32cExpectedValue: hashes.crc32c,
+          md5ExpectedValue: hashes.md5,
```
danielbankhead, line 1480:
```suggestion
          crc32cGenerator: this.crc32cGenerator,
          crc32cExpectedValue: hashes.crc32c,
          md5ExpectedValue: hashes.md5,
```

danielbankhead (commented):
Added some suggestions, mostly for the hashstream validator

ddelgrosso1:
@danielbankhead I accepted most of the suggestions but had to fix up a few things. We cannot deduce whether or not crc32c or md5 checks are enabled purely by the presence of a hash. The server returns these values in the headers and the user may not want the validation so the boolean has to be set from user land. Additionally, `createWriteStream` uses the HashStreamValidator to calculate hashes when uploading. For this reason I had to bring back the `updateHashesOnly` flag so that it may be passed in the appropriate place so that validation is not performed.

src/hash-stream-validator.ts
```diff
@@ -54,12 +68,41 @@ class HashStreamValidator extends Transform {
...
+    // validation is not possible in this case.
+    let failed = this.crc32cEnabled || this.md5Enabled;
+
+    if (this.crc32cEnabled && this.crc32cExpected) {
+      failed = !this.test('crc32c', this.crc32cExpected);
+    }
+
+    if (this.md5Enabled && this.md5Expected) {
+      failed = !this.test('md5', this.md5Expected);
+    }
```
danielbankhead, line 94: I think I see the confusion - I think this block should only be checked in the presence of `this.crc32cExpected` or `this.md5Expected` (otherwise, if either aren't available then this would always fail).

danielbankhead (approved):
LGTM, the `_flush` suggestion is optional

PR #2161, opened 2023-03-07
test: add ranged read test, performance test code cleanups

ddelgrosso1:

### DIFF
--- a/internal-tooling/performanceUtils.ts
+++ b/internal-tooling/performanceUtils.ts
@@ -267,7 +267,7 @@ export async function* convertToCloudMonitoringFormat(
     generation="",\
     upload_id="",\
     retry_count="",\
-    status_code=""} ${throughput}`;
+    status_code="${curResult.status}"} ${throughput}`;
   }
 }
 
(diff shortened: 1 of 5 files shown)


internal-tooling/performanceUtils.ts
```diff
@@ -17,16 +17,37 @@
...
 import {Bucket, Storage, TransferManager} from '../src';
 
 export const BLOCK_SIZE_IN_BYTES = 1024;
-export const DEFAULT_SMALL_FILE_SIZE_BYTES = 5120;
-export const DEFAULT_LARGE_FILE_SIZE_BYTES = 2.147e9;
 export const NODE_DEFAULT_HIGHWATER_MARK_BYTES = 16384;
 export const DEFAULT_DIRECTORY_PROBABILITY = 0.1;
-export const DEFAULT_PROJECT_ID = 'GCS_NODE_PERFORMANCE_METRICS';
 export const DEFAULT_NUMBER_OF_OBJECTS = 1000;
+export const enum OUTPUT_FORMATS {
```
danielbankhead, line 27: I see these already existed, but maybe we want an [`as const`](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) object rather than [`const enum`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums).

danielbankhead (approved):
lgtm - one minor comment

internal-tooling/performanceUtils.ts
ddelgrosso1: Done

PR #2185, opened 2023-04-26
perf: Improve Multiple Chunk Upload Performance

danielbankhead:
`chunkSize` uploads should be much faster.  🦕

### DIFF
--- a/src/resumable-upload.ts
+++ b/src/resumable-upload.ts
@@ -484,11 +484,12 @@ export class Upload extends Writable {
       }
 
       yield bufToYield;
-    }
 
-    // notify upstream we've read from the buffer and we're able to consume
-    // more so it can potentially send more data down.
-    process.nextTick(() => this.emit('readFromChunkBuffer'));
+      // Notify upstream we've read from the buffer and we're able to consume
+      // more. It can also potentially send more data down as we're currently
+      // iterating.
+      this.emit('readFromChunkBuffer');
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


danielduhh (commented):
This looks like a significant change to the resumable write path. I suggest a design with some background information & comprehensive performance tests before merging this

src/resumable-upload.ts
```diff
@@ -260,20 +260,23 @@ export class Upload extends Writable {
     uri: uuid.v4(),
     offset: uuid.v4(),
   };
-  private upstreamChunkBuffer: Buffer = Buffer.alloc(0);
-  private chunkBufferEncoding?: BufferEncoding = undefined;
+  /**
+   * A cache of buffers written to this instance, ready for consuming
+   */
+  private writeBuffers: Buffer[] = [];
+  private bufferEncoding?: BufferEncoding = undefined;
```
ddelgrosso1, line 267: I think you mentioned removing this in our discussion but does encoding actually ever get used for anything?

danielbankhead: Not really; only for incoming buffers that are encoded to strings - which we decode here:

https://github.com/googleapis/nodejs-storage/blob/efb99b8aad6dbb8cf698cfcabc6a4d1fc3bbefd8/src/resumable-upload.ts#L398

ddelgrosso1:
Overall I think this looks good and since we walked through the code I understand the why/how. I think once the supporting document is a bit more fleshed out with some metrics, I am good with this getting merged / released.

ddelgrosso1 (approved):
Going to approve this pending the results of perf tests.

PR #2190, opened 2023-05-03
test: update retry conf test data

tritone:
Copies the retry_test json from the latest in the conformance test repo. Rename retryStrategyTest to retryTest for consistency, and add 408 errors for upload tests. S8 is still not implemented in node so that part is skipped.


ddelgrosso1 (approved):

tritone:
I had to drop 1 408 case where it happens during the middle of an upload. I believe this is a testbench issue; filing an issue on that repo and I will add in that case when it's ready.

ddelgrosso1:
@tritone thanks for the fixes / updates 😆

PR #2192, opened 2023-05-04
feat: MPU for transfer manager

ddelgrosso1:

### DIFF
--- a/internal-tooling/performTransferManagerTest.ts
+++ b/internal-tooling/performTransferManagerTest.ts
@@ -51,7 +51,7 @@ const argv = yargs(process.argv.slice(2))
  * to the parent thread.
  */
 async function main() {
-  let result: TestResult | undefined = undefined;
+  let results: TestResult[] = [];
 
   ({bucket, transferManager} = await performanceTestSetup(
     argv.project!,
@@ -60,18 +60,18 @@ async function main() {
 
   switch (argv.test_type) {
     case PERFORMANCE_TEST_TYPES.TRANSFER_MANAGER_UPLOAD_MANY_FILES:
-      result = await performUploadManyFilesTest();
+      results = await performUploadManyFilesTest();
       break;
     case PERFORMANCE_TEST_TYPES.TRANSFER_MANAGER_DOWNLOAD_MANY_FILES:
-      result = await performDownloadManyFilesTest();
+      results = await performDownloadManyFilesTest();
       break;
     case PERFORMANCE_TEST_TYPES.TRANSFER_MANAGER_CHUNKED_FILE_DOWNLOAD:
-      result = await performDownloadFileInChunksTest();
+      results = await performChunkUploadDownloadTest();
       break;
     default:
       break;
   }
-  parentPort?.postMessage(result);
+  parentPort?.postMessage(results);
   await performTestCleanup();
 }
 
@@ -87,7 +87,7 @@ async function performTestCleanup() {
  *
  * @returns {Promise<TestResult>} A promise that resolves containing information about the test results.
  */
-async function performUploadManyFilesTest(): Promise<TestResult> {
+async function performUploadManyFilesTest(): Promise<TestResult[]> {
   const fileSizeRange = getLowHighFileSize(argv.object_size);
   const creationInfo = generateRandomDirectoryStructure(
     argv.num_objects,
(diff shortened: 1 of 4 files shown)


src/transfer-manager.ts
```diff
@@ -369,6 +575,98 @@ export class TransferManager {
...
+        await mpuHelper.initiateUpload();
+      }
+      const startOrResumptionByte = mpuHelper.partsMap!.size * chunkSize;
+      const readStream = createReadStream(filePath, {
+        highWaterMark: chunkSize,
+        start: startOrResumptionByte,
+      });
+      for await (const curChunk of readStream) {
+        if (promises.length >= maxQueueSize) {
+          await Promise.all(promises);
```
danielduhh, line 652: what does this do? Execute all the parts? Should this get returned to the caller?

src/transfer-manager.ts
```diff
@@ -369,6 +575,98 @@ export class TransferManager {
...
+          promises = [];
+        }
+        promises.push(
+          limit(() => mpuHelper.uploadPart(partNumber++, curChunk))
+        );
+      }
+      await Promise.all(promises);
+      await mpuHelper.completeUpload();
```
danielduhh, line 660: What's the return statement here?

src/transfer-manager.ts
```diff
@@ -69,6 +84,197 @@ export interface DownloadFileInChunksOptions {
...
+    this.partsMap = partsMap;
+  }
+}
+/**
+ * Class representing an implementation of MPU in the XML API. This class is not meant for public usage.
+ *
+ * @private
+ * @experimental
+ */
+class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
```
danielduhh, line 142: should this go in a separate file?

internal-tooling/performTransferManagerTest.ts
```diff
@@ -179,15 +179,17 @@ async function performDownloadManyFilesTest(): Promise<TestResult> {
...
 }
 
 /**
- * Performs a test where a large file is downloaded as chunks in parallel.
+ * Performs a test where a large file is uploaded and downloaded as chunks in parallel.
  *
  * @returns {Promise<TestResult>} A promise that resolves containing information about the test results.
  */
-async function performDownloadFileInChunksTest(): Promise<TestResult> {
+async function performChunkUploadDownloadTest(): Promise<TestResult[]> {
```
danielduhh, line 191: Whats the reasoning for combining these? Simplicity?

danielduhh (commented):
Very exciting! Just had some initial thoughts/questions...

src/transfer-manager.ts
```diff
@@ -19,8 +19,12 @@ import {DownloadOptions, DownloadResponse, File} from './file';
 import * as pLimit from 'p-limit';
 import * as path from 'path';
 import * as extend from 'extend';
-import {promises as fsp} from 'fs';
+import {createReadStream, promises as fsp} from 'fs';
 import {CRC32C} from './crc32c';
+import {GoogleAuth} from 'google-auth-library';
+import {XMLParser, XMLBuilder} from 'fast-xml-parser';
```
danielduhh, line 25: Add to package JSON? It would be interesting to see the level of effort for adding this to one of our shared libs (gaxios?).

src/transfer-manager.ts
ddelgrosso1: This is similar to the changes I made for `deleteFiles`. `p-limit` only constrains the number of running promises, not the number of promises being held in memory. I added a queue here such that if a user is uploading a large file we aren't attempting to hold all those chunks in memory at once. By default we will only hold the same number of chunks in memory as the number of executing promises.

src/transfer-manager.ts
ddelgrosso1: The return is of type `Promise<void>` in the success case there is no return value to the user. If something goes wrong we throw an exception with the error, uploadId, and mapping of completed parts that were uploaded. This allows the user to call again supplying the uploadId and parts so that we may resume from where the failure occurred.

src/transfer-manager.ts
ddelgrosso1: I can move this to a separate file the only problem is I will have to export it which means a user can potentially come along and import it and use it directly. I was attempting to avoid that situation. Depends what people think and whether or not we are ok with saying "don't use this directly".

internal-tooling/performTransferManagerTest.ts
ddelgrosso1: Workload 8 is supposed to be upload and download test so I combined them.

src/transfer-manager.ts
ddelgrosso1: I added it: https://github.com/googleapis/nodejs-storage/pull/2192/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R62. I'm sure we could write something ourselves. I figured since this was a one off usage of the XML API it probably wasn't worth doing ourselves.

danielduhh (approved):
looks good. One question on returning the response of complete upload

src/transfer-manager.ts
ddelgrosso1: I change it to return the full response just to be inline with the raw API.

src/transfer-manager.ts
ddelgrosso1: Added a comment to explain the logic.

src/transfer-manager.ts
ddelgrosso1: Now returns `Promise<GaxiosResponse | undefined>`. The GaxiosResponse is the full API response.

PR #2199, opened 2023-05-18
fix: allow files in directories to be downloaded onto local machine

### LINKED ISSUE #2200: transferManager.downloadManyFiles not writing files locally
#### Issue summary
I hesitated to split this into 2 issues because they are rather tightly coupled

1. Issue 1: When `prefix` is not set, `transferManager.downloadManyFiles` does not actually write files locally. This is because `destination` is `undefined`.
2. Related issue 2: When `prefix` is set, `file.download(...)` attempts to write the file, but is unable to recursively create directory structure on local machine.

#### Proposed draft PR
https://github.com/googleapis/nodejs-storage/pull/2199

#### Standard questions
1) Is this a client library issue or a product issue? - client library
2) Did someone already solve this? - it hasn't been asked yet in the Issue or Stack overflow
3) Do you have a support contract? - At work yes, but this is a personal contribution outside work

#### Environment details

  - OS: macOS Catalina
  - Node.js version: v16.15.0
  - npm version: 8.5.5
  - `@google-cloud/storage` version: 6.10.1

#### Steps to reproduce

Issue 1:
  1. Implement code example for 'downloadManyFiles' without `prefix`
  ```
  async function downloadDirectory({ bucketId, sourceDirectory }) {
    const transferManager = new TransferManager(storage.bucket(bucketId));
    const options = {};

    async function downloadFolderWithTransferManager() {
      await transferManager.downloadManyFiles(sourceDirectory, options);
    }
    const res = await downloadFolderWithTransferManager();
}
  ```
  2. Running the function returns no error. However, no files are written locally, and a well-placed console.warn shows that this is because the `destination` is undefined. The draft PR contains a suggestion to always generate the destination.
  
Screen Shot 2023-05-18 at 12 53 20 AM
  

Issue 2:
  1. Implement same example for 'downloadManyFiles', except we now set `prefix`
  ```
  async function downloadDirectory({ bucketId, sourceDirectory }) {
    const transferManager = new TransferManager(storage.bucket(bucketId));

    const options = {
      prefix: "hello",
    };
    async function downloadFolderWithTransferManager() {
      await transferManager.downloadManyFiles(sourceDirectory, options);
    }
    const res = await downloadFolderWithTransferManager();
}
  ```
  2. Running the function returns 
  ```
  ENOENT: no such file or directory, open 'hello/test-folder/'
  ```
  
Screen Shot 2023-05-18 at 12 38 18 AM
 
  This is due to the file stream at https://github.com/googleapis/nodejs-storage/blob/fe9e3a461356a7fa6e698d364174d4dc84aaa12b/src/file.ts#L2108-L2111 attempting to save a directory object locally. It appears that google storage api returns empty directory objects according to google docs here https://cloud.google.com/knowledge/kb/listing-only-objects-using-gsutil-and-not-directories-000004568. The draft PR proposes to filter out these empty directory objects from the final write.

Thanks, it is my first issue and fork of the repo! Open to feedback. I've already signed the CLA, but kept the PR as draft until the codeowner reviews the issue/intention of the existing design.

hochoy:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-storage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary) (n/a)

Fixes #2200 🦕


hochoy:
@ddelgrosso1 , I tried running `samples-test` and `system-test` but it appears those require an active project in order to do an e2e/live test against real GCP infrastructure. 

Is that accurate? 
How expensive is that if I run it on my own project? Should I be triggering it here in the repo instead (cloudbuild/github actions)?

hochoy:
Also, I would like to add a test to check that the files are truly downloaded into the `local machine` or `memory`. 
Would you recommend I give that a shot? Or is that too heavy/invasive of a validation?

src/transfer-manager.ts
```diff
@@ -258,13 +260,11 @@ export class TransferManager {
...
-          options.prefix || '',
-          passThroughOptionsCopy.destination || '',
-          file.name
-        );
-      }
+      passThroughOptionsCopy.destination = path.join(
+        options.prefix || '',
+        passThroughOptionsCopy.destination || '',
+        file.name
+      );
```
hochoy, line 267: I think I will revert this, now that I understand the original design:
1. If `options.prefix` is provided, set `destination`. This causes `files.download` to download files to local machine 
2. If `options.prefix` is not provided, `destination` is not set. This causes `files.download` to return a promise
3.  If `options.prefix` is not provided AND `options.passThroughOptions` is a `function/callback`, then apply the callback onto the results of the file download.

ddelgrosso1:
> @ddelgrosso1 , I tried running `samples-test` and `system-test` but based on the logs, it appears those require an active project in order to do an e2e/live test against real GCP infrastructure.
> 
> Is that accurate? How expensive is that if I run it on my own project? Should I be triggering it here in the repo instead (and leverage your repo's cloudbuild/github actions infra)?

That is accurate. I wouldn't worry about running these on your own, they get run in the CI pipeline each time a commit is pushed. However, unit tests should work without issue locally. 



> Also, I would like to add a test to check that the files are truly downloaded into the `local machine` or `memory`. Would you recommend I give that a shot? Or is that overkill?

If you feel up to it a unit test can probably be created to test this. I can look to see if we have any similar tests elsewhere that might serve as a guide.

One thing I will do is to cleanup the the JS Docs to make it abundantly clear that not supplying a `prefix` will result in the files downloaded to memory.

src/file.ts
```diff
@@ -2098,6 +2099,8 @@ class File extends ServiceObject<File> {
     const fileStream = this.createReadStream(options);
     let receivedData = false;
     if (destination) {
+      fs.mkdirSync(path.dirname(destination), {recursive: true});
```
ddelgrosso1, line 2102: One thing that we might want to handle here is if the directory already exists. I believe in that situation this will throw an error. Additionally since this code path can be used for a single file `path.dirname('foo')` will return `'.'`.

hochoy: Thanks @ddelgrosso1 I had a look at the docs, and `fs.mkdirSync` has the same behaviour as `fs.mkDir`, in which:
- `recursive: false` (default) will throw `EEXIST`
- `recursive: true` will resolve without an error

It's mentioned here [in the docs](https://nodejs.org/api/fs.html#fsmkdirpath-options-callback) [all the way back to Node v12](https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_mkdir_path_options_callback). 

> Calling fs.mkdir() when path is a directory that exists results in an error only when recursive is false

Docs can't always be trusted, so I tried it out:

Recursive: true ===> no error
recursive-true

Recursive: true ===> EEXIST error
recursive-false

hochoy: That being said, we could add a `fs.existsSync()` just to be safe. Let me know, thanks!

hochoy, line 2102:
```suggestion
      const dir = path.dirname(destination)
      if (!fs.existsSync(dir)) {
        fs.mkdirSync(dir,  {recursive: true});
      }
```

src/transfer-manager.ts
```diff
@@ -238,6 +238,8 @@ export class TransferManager {
         prefix: filesOrFolder,
       });
       files = directoryFiles[0];
+      // filter out directories
```
hochoy, line 241:
```suggestion
      // filter out directory objects
```

hochoy: https://cloud.google.com/knowledge/kb/listing-only-objects-using-gsutil-and-not-directories-000004568

Just for the record
Screen Shot 2023-05-28 at 6 43 43 PM

test/file.ts
```diff
@@ -2678,6 +2685,82 @@ describe('File', () => {
...
+                  fs.existsSync(path.dirname(nestedDir)),
+                  false
+                );
+                assert.strictEqual(fs.existsSync(nestedDir), false);
+
+                // Temporary comment: Not very useful since no user would care about an empty buffer. Could remove.
+                assert.strictEqual(buffer.equals(Buffer.alloc(0)), true);
+                done();
+              } catch (e) {
+                done(e);
```
hochoy, line 2758: Any `assert(false)` will throw an Error, meaning `done()` never gets called and the test times out. Adding the `try catch` guarantees `done()` gets called. Similar to what is described in this Stack overflow discussion: https://stackoverflow.com/questions/66461468/mocha-test-false-assert-timeouts

I'm a bit surprised why this doesn't cause issues in the `fs.readFile` callback assertions in the other tests.

test/file.ts
```diff
@@ -2678,6 +2685,82 @@ describe('File', () => {
...
+                // Temporary comment: Option 1: Spy on fs methods
+                assert.strictEqual(createWriteStream.callCount, 0);
+                assert.strictEqual(mkdirSync.callCount, 0);
+
+                // Temporary comment: Option 2: Verify that no folder and/or file were created
+                assert.strictEqual(
+                  fs.existsSync(path.dirname(nestedDir)),
+                  false
+                );
+                assert.strictEqual(fs.existsSync(nestedDir), false);
```
hochoy, line 2752: Wanted to get your thoughts on what checks make sense here:
- Option 1: Spy on `fs` methods
- Option 2: Check the files and folders written into the temporary directory

There's pros and cons. If this library ever stops using `fs` (unlikely), then the spy wouldn't make sense anymore. On the flip side, checking the folder and files feels a lot like a system-test, since it is reliant on an actual working (linux-based) filesystem.

src/file.ts
```diff
@@ -2097,7 +2098,18 @@ class File extends ServiceObject<File> {
...
+
+    // Skip directory objects as they cannot be written to local filesystem
+    /**
+     * Temporary comment:
+     * Could consider returning an error, but TransferManager will have to handle rejected promises in Promises.all.
+     * Example: Promise.allSettled(promises) or Promise.all(promises.map(p => p.catch(e => e)))
+     */
+    if (destination && destination.endsWith('/')) {
+      callback?.(null, Buffer.alloc(0));
+    } else if (destination) {
```
hochoy, line 2110: With how GCS works, empty directory objects get returned to `getFiles`. This causes issues as TransferManager will attempt to write both `folder1/` and `folder1/foo.txt`. 

Writing `folder1/` throws this:
```
Error: EISDIR: illegal operation on a directory, open '/folder1/'
```

In this if statement, we skip writing directory objects, and just return an empty buffer of size zero. Could also choose to return an Error, however, that means TransferManager has to handle rejected promises.

src/file.ts
ddelgrosso1: I think it is fine without the additional check (I probably should have looked at the docs first 😆)

test/file.ts
ddelgrosso1: I would lean towards 1 as I don't think we are in any danger of removing `fs` and I don't think these tests should be doing any real disk IO.

src/file.ts
```diff
@@ -2097,7 +2098,21 @@ class File extends ServiceObject<File> {
 
     const fileStream = this.createReadStream(options);
     let receivedData = false;
-    if (destination) {
+
+    // Skip directory objects as they cannot be written to local filesystem
+    /**
+     * Temporary comment:
+     * Could consider returning an error, but TransferManager will have to handle rejected promises in Promises.all.
+     * Example: Promise.allSettled(promises) or Promise.all(promises.map(p => p.catch(e => e)))
```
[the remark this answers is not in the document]
hochoy: ok

test/file.ts
hochoy: okay

src/transfer-manager.ts
```diff
@@ -258,7 +260,7 @@ export class TransferManager {
         {},
         options.passthroughOptions
       );
-      if (options.prefix) {
+      if (options.prefix || passThroughOptionsCopy.destination) {
```
hochoy, line 263: having iterated on a few approaches to get to this point, I've realized now that (in hindsight), `options.prefix` and `options.passthroughoptions.destination` fulfill the same purpose in `TransferManager`. Both are being prefixed toe the `file.name` via `path.join`. Technically, we only need one of the two.

That being said, I think we should probably keep `options.prefix` for:
- backwards compatibility for users that are currently using it
- users that prefer to set the more visible `options.prefix` over setting `options.passthroughoptions.destination`

ddelgrosso1: Yes, I think for the time let's keep both as to avoid a breaking change. I will make a note to clean this up when we prepare the next major revision.

ddelgrosso1 (approved):
LGTM

ddelgrosso1:
@danielbankhead would you mind just giving this a second set of eyes?

danielbankhead (changes requested):
Looks great! I think we can make it better for `File#download` customers by moving the logic to `TransferManager#downloadManyFiles`

src/file.ts
```diff
@@ -2097,7 +2098,16 @@ class File extends ServiceObject<File> {
...
     const fileStream = this.createReadStream(options);
     let receivedData = false;
-    if (destination) {
+
+    // Skip directory objects as they cannot be written to local filesystem
+    if (
+      destination &&
+      (destination.endsWith('/') || destination.endsWith('\\'))
+    ) {
+      callback?.(null, Buffer.alloc(0));
```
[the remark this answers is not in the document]
hochoy: Hey @danielbankhead , thanks for the review. The issue with object names ending with `/` is that while they may be valid on GCS, it can't actually be written on a Linux-based OS. I don't have a windows to try it on. Perhaps you know of an edge case where names ending with `/` can actually be written.

Screen Shot 2023-06-09 at 10 26 27 AM

The scenario you described is valid, if `destination` is unset/falsy, and a `Buffer` gets created/returned for the GCS object ending with `/`. That behaviour remains unchanged in this PR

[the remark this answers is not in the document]
hochoy: The scenario you described is valid, if destination is unset/falsy, and a Buffer gets created/returned for the GCS object ending with /. That behaviour remains unchanged in this PR

src/file.ts
```diff
@@ -2097,7 +2098,16 @@ class File extends ServiceObject<File> {
...
+    } else if (destination) {
+      fs.mkdirSync(path.dirname(destination), {recursive: true});
```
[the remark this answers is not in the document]
hochoy: Coincidentally, that was the original design I had (making the change on TransferManager). However, I made the change on `File#download` because the root issue is that `File#download` itself cannot write nested paths. And solving it there would solve the problem for both `File` and `TransferManager`.

Screen Shot 2023-06-09 at 10 39 25 AM

That being said, it's not hard to optimize at TransferManager, I can create another branch. What do we do then about the limitation of `File#download` above? Do we warn users not to provide nested destination paths?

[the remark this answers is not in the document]
danielbankhead: I think a simple solution would be to add a flag to gate this functionality and Transfer Manager use this flag.

src/file.ts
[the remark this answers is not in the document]
danielbankhead: For this case, I think allowing the file system to error would be clearer and easier to understand than returning a Buffer.

hochoy:
Hey @danielbankhead @ddelgrosso1 , just wanted to provide a heads up that I'll be revisiting this in about 1-2 months due to project priorities. Thank you for all the feedback and recommendations, they are all valid points for a user like myself.

I do need something like `TransferManager` for a project at work, so I'll very likely be back (no promises of course). But in its current design, it would be more practical for me to build my own version of the TransferManager to reflect the interface I'm looking for. I will likely wrap the existing `File.download` method during my experimentation.

Couple things I would design for, after both your feedback:

- downloads should always be concurrent
- empty objects (like "folders") on gcp are not writeable need special handling
- directory creation should be optimized at the TransferManager level
- unix vs windows-based file handling
- read and write failures should not fail silently. In the event of a partial directory/folder download, we should make it easy for the user to identify which files succeeded vs failed. 
- if similar files exist in the write destination, we could provide a default of either over-writing or error-throwing or user-prompt

I envision users wanting to use TransferManager to download entire buckets or "sub-directories". So, those are some of the considerations I could think of.

src/file.ts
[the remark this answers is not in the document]
ddelgrosso1: @vishwarajanand I think we want to address this per @danielbankhead feedback.

src/file.ts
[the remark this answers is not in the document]
ddelgrosso1: @vishwarajanand same comment here, I think we want to address @danielbankhead idea about flagging / gating this.

test/file.ts
vishwarajanand: This test is remodified into `TransferManager` and this part is no longer needed

vishwarajanand: This test is re-written into `transfer-manager`, so this comment is no longer needed

src/file.ts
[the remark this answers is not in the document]
vishwarajanand: I moved this logic to `transfer-manager`, for clarity and back compat.

src/file.ts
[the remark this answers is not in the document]
vishwarajanand: I moved this logic to `transfer-manager`, for clarity and back compat.
Now, a gating is not required because we can keep "create nested directories as the default behavior".

vishwarajanand:
@ddelgrosso1 I see a windows ci failure due to test code, which i will fix shortly.
Can you review this PR? I also drafted a summary of changes here fro your ease: [go/node-transfermanager-gcs](http://goto.google.com/node-transfermanager-gcs)

src/transfer-manager.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from './file.js';
 import pLimit from 'p-limit';
 import * as path from 'path';
-import {createReadStream, promises as fsp} from 'fs';
+import {createReadStream, mkdirSync, promises as fsp} from 'fs';
```
ddelgrosso1, line 27: As @danielbankhead noted in the doc, probably want the async version of mkdir.

ddelgrosso1 (approved):
LGTM

danielbankhead (approved):

PR #2203, opened 2023-05-26
feat: support iterables in file#save (#2202)

### LINKED ISSUE #2202: File#save should support Readable / AsyncIterable data

I have a small utility function in my codebase that looks like this:
```js
const { pipeline } = require('node:stream/promises')

async function write(file, data, type, metadata = undefined) {
  if (Buffer.isBuffer(data) || typeof data === 'string') {
    await file.save(data, {
      timeout: 9e3,
      contentType: type,
      resumable: false,
      metadata: { metadata },
    })
  } else {
    const writable = file.createWriteStream({
      timeout: 9e3,
      gzip: true,
      contentType: type,
      resumable: false,
      metadata: { metadata },
    })

    await pipeline(data, writable)
  }
}
```

And it seems this could easily be implemented in the lib itself.

I would like `File#save` 's data argument to support `AsyncIterable` (`Readable`)

see code example above

matthieusieben:

Fixes #2202 🦕

BEGIN_COMMIT_OVERRIDE
feat: support iterables in file@save
END_COMMIT_OVERRIDE

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -31,7 +31,14 @@ import * as extend from 'extend';
 import * as fs from 'fs';
 import * as mime from 'mime';
 import * as resumableUpload from './resumable-upload';
-import {Writable, Readable, pipeline, Transform, PassThrough} from 'stream';
+import {
+  Writable,
+  Readable,
+  pipeline,
+  Transform,
+  PassThrough,
+  PipelineSource,
+} from 'stream';
 import * as zlib from 'zlib';
 import * as http from 'http';
 
@@ -92,6 +99,8 @@ export interface PolicyDocument {
   base64: string;
   signature: string;
 }
+
+export type SaveData = string | Buffer | PipelineSource<string | Buffer>;
 
 export type GenerateSignedPostPolicyV2Response = [PolicyDocument];
 
@@ -419,6 +428,7 @@ export enum FileExceptionMessages {
   UPLOAD_MISMATCH = `The uploaded data did not match the data from the server.
     As a precaution, the file has been deleted.
     To be sure the content is the same, you should try uploading the file again.`,
+  STREAM_NOT_READABLE = 'Stream must be readable.',
 }
 
 /**
@@ -3527,13 +3537,9 @@ class File extends ServiceObject<File> {
     this.copy(newFile, copyOptions, callback!);
   }
 
-  save(data: string | Buffer, options?: SaveOptions): Promise<void>;
-  save(data: string | Buffer, callback: SaveCallback): void;
-  save(
-    data: string | Buffer,
-    options: SaveOptions,
-    callback: SaveCallback
-  ): void;
+  save(data: SaveData, options?: SaveOptions): Promise<void>;
+  save(data: SaveData, callback: SaveCallback): void;
+  save(data: SaveData, options: SaveOptions, callback: SaveCallback): void;
   /**
    * @typedef {object} SaveOptions
    * @extends CreateWriteStreamOptions
@@ -3560,7 +3566,7 @@ class File extends ServiceObject<File> {
    * resumable feature is disabled.
    * </p>
    *
-   * @param {string | Buffer} data The data to write to a file.
+   * @param {SaveData} data The data to write to a file.
    * @param {SaveOptions} [options] See {@link File#createWriteStream}'s `options`
    *     parameter.
    * @param {SaveCallback} [callback] Callback function.
@@ -3588,7 +3594,7 @@ class File extends ServiceObject<File> {
    * ```
    */
   save(
-    data: string | Buffer,
+    data: SaveData,
     optionsOrCallback?: SaveOptions | SaveCallback,
     callback?: SaveCallback
   ): Promise<void> | void {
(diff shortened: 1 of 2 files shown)


ddelgrosso1:
@matthieusieben thank you for the contribution, we always appreciate getting them. Let me check regarding the disabling of retries but I think we will have to find another way around as this would not adhere to the documentation and would probably be pretty specific to Node storage when compared to our other languages.

src/file.ts
```diff
@@ -3629,7 +3635,23 @@ class File extends ServiceObject<File> {
...
+            // Streams should not be used more than once. AsyncIterables can be used multiple times.
+            if (
+              typeof data !== 'function' &&
+              'pipe' in data &&
+              typeof data.pipe === 'function'
+            ) {
+              this.storage.retryOptions.autoRetry = false;
+            }
+
+            pipeline(data, writable, _err => {
```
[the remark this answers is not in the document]
matthieusieben: They will trigger a write error and cause the destruction of the writable. I adapted to also account for invalid uses of `file.save` with `data` that is not of the valid type.

src/file.ts
```diff
@@ -3629,7 +3635,23 @@ class File extends ServiceObject<File> {
...
+
+          if (typeof data === 'string' || Buffer.isBuffer(data)) {
+            writable.end(data);
+          } else {
+            // Streams should not be used more than once. AsyncIterables can be used multiple times.
+            if (
+              typeof data !== 'function' &&
+              'pipe' in data &&
+              typeof data.pipe === 'function'
+            ) {
```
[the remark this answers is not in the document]
matthieusieben: Done

matthieusieben:
Here is a summary of the changes that were added:
- Do not attempt to read a destroyed readable as this would yield a never resolving promise
- Do not leave the promise "hanging" when we bail (`reject` will always be called, regardless of wether `bail` was called)
- Do not retry when a failure occurs with a readable stream
- Properly cleanup if `data` is not valid and causes `pipeline` to throw before the pipeline is actually running
- Added more test cases

src/file.ts
```diff
@@ -3608,28 +3615,56 @@ class File extends ServiceObject<File> {
...
+                // If data is not a valid PipelineSource, then pipeline will
+                // fail without destroying the writable stream. If data is a
+                // PipelineSource that yields invalid chunks (e.g. a stream in
+                // object mode or an iterable that does not yield Buffers or
+                // strings), then pipeline will destroy the writable stream.
+                if (!writable.destroyed) writable.destroy();
+
+                if (data instanceof Readable) bail(err);
+
+                handleError(err);
```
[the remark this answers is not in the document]
matthieusieben: (Async)Iterables also need a bail indeed. `PipelineSourceFunction` do not need a bail. Adapted accordingly.

src/file.ts
```diff
@@ -93,6 +93,16 @@ export interface PolicyDocument {
...
 
+type PipelineSourceFunction<T> = (arg?: {
+  signal?: AbortSignal;
+}) => Iterable<T> | AsyncIterable<T>;
+type PipelineSource<T> =
+  | Iterable<T>
+  | AsyncIterable<T>
+  | NodeJS.ReadableStream
+  | PipelineSourceFunction<T>;
+export type SaveData = string | Buffer | PipelineSource<string | Buffer>;
```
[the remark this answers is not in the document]
matthieusieben: Done

matthieusieben:
One approach would be to add the following when `data` is not a buffer or string
```js
// In node 12, pipeline() only supports "Stream"s
if (!('pipe' in data) || typeof data.pipe !== 'function') {
  const nodeVersion = Number(process.versions.node.split('.')[0]);
  if (!nodeVersion || nodeVersion < 14) {
    throw new Error('Saving using an iterable requires Node 14 (or higher)');
  }
}
```

And to catch that particular error in tests. What do you think ?

ddelgrosso1:
> The reason node-12 is failing is that `pipeline` only supported `Streams` (and not `AsyncIterable`, `Iterable` or `PipelineSourceFunction`). Not only did node12 not support those but it does not even throw a proper error in theses cases.
> 
> https://nodejs.org/docs/latest-v12.x/api/stream.html#stream_stream_pipeline_streams_callback
> 
> How should I adapt to this limitation of Node 12 ?

We are preparing to move to Node 14 minimum in the next few weeks and alongside it a new major version of the library. I suggest we not add code to handle Node 12 at this point. Since everything else is passing we can just release this once the new major version goes live.

src/file.ts
```diff
@@ -3608,28 +3614,60 @@ class File extends ServiceObject<File> {
     }
     const returnValue = retry(
       async (bail: (err: Error) => void) => {
-        await new Promise<void>((resolve, reject) => {
+        if (data instanceof Readable && !data.readable) {
+          // Readable streams can only be consumed once and would cause a dead
+          // lock when used with pipeline().
+          return bail(new Error(FileExceptionMessages.STREAM_NOT_READABLE));
+        }
```
danielbankhead, line 3621: Is this true? I've tested this locally and pipeline seems to exit cleanly. Example:

```ts
const readable = stream.Readable({read() {this.push(null)}});
readable.read();
readable.readable // false
stream.pipeline(a, stream.PassThrough, () => { console.log('done')})
```

danielbankhead (approved):
Minor comment, but looks good to me

danielbankhead:
Adding `do not merge` until after the next major

src/file.ts
matthieusieben: When I run your code on Node `v16.20.1`, `done` is not printed. This means that the pipeline callback is never called. If you were to resolve/reject a promise in said callback, you would have a forever hanging promise.

```js
const stream = require('stream');

const readable = stream.Readable({read() {this.push(null)}});
readable.read();
readable.readable // false
stream.pipeline(readable, stream.PassThrough, () => { console.log('Never called!!!')})
```

matthieusieben: Consider the following example:
```js
const stream = require('stream');

const readable = stream.Readable({
  read() {
    this.push('aaa');
    this.push(null);
  },
});
readable.read();
readable.readable; // false
stream.pipeline(
  readable,
  async function* (src) {
    for await (const chunk of src) {
      console.log(chunk);
    }
  },
  err => {
    console.log(err ? 'Pipeline failed.' : 'Pipeline succeeded.');
  }
);
```

Running this results in "Pipeline succeeded" being printed while no chunk was actually "piped". I adapted the comment in my commit to better reflect what happens.

matthieusieben: Note that maybe this is an expected behavior in NodeJS: If a stream is piped twice, the second time it will appear "empty" (as if it contained 0 bytes). If that is the case, maybe this `if` should be removed ? It's kind of an edge case but on the other hand, as a dev, if I mistakenly `file.save(alreadyConsumed)` I would rather see an error than upload an empty file...

matthieusieben: I reported an [issue](https://github.com/nodejs/node/issues/48764) in NodeJS regarding this.

matthieusieben: I also added an await for next tick to better detect wether the stream is not readable:

```js
const stream = require('stream');
const {pipeline} = require('stream/promises');

async function main() {
  const readable = stream.Readable({
    read() {
      this.push('aaa');
      this.push(null);
    },
  });

  console.log(readable.readable); // true

  readable.read();

  console.log(readable.readable); // true
  console.log(readable.destroyed); // false
  console.log(readable._readableState.ended); // true

  await new Promise((resolve) => setImmediate(resolve));

  console.log(readable.readable); // false
  console.log(readable.destroyed); // true
  console.log(readable._readableState.ended); // true

  await pipeline(readable, async function* (source) {
    for await (const chunk of source) {
      console.log('chunk', chunk);
    }
    console.log('done');
  });
}

main();

```

ddelgrosso1 (approved):
LGTM. Thank you for the contribution, we always appreciate them!

ddelgrosso1:
Removed do not merge as `7.0` is live.

ddelgrosso1:
BEGIN_COMMIT_OVERRIDE
feat: support iterables in file@save
END_COMMIT_OVERRIDE

PR #2214, opened 2023-06-07
refactor!: mark bucket.setLabels and associated interfaces as deprecated

### LINKED ISSUE #2141: refactor: deprecate bucket.setLabels
`bucket.setLabels` is nothing more than a wrapper around `bucket.setMetadata`. There was a recent customer issue where the customer was chaining calls to `setLabels` and `setMetadata` when in fact this could be accomplished in a single call. We should deprecate `setLabels` and instead document how to achieve the same functionality utilizing `setMetadata`.

ddelgrosso1:

Fixes #2141 🦕

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -3530,20 +3530,24 @@ class Bucket extends ServiceObject {
     callback: SetLabelsCallback
   ): void;
   /**
+   * @deprecated
    * @typedef {array} SetLabelsResponse
    * @property {object} 0 The bucket metadata.
    */
   /**
+   * @deprecated
    * @callback SetLabelsCallback
    * @param {?Error} err Request error, if any.
    * @param {object} metadata The bucket metadata.
    */
   /**
+   * @deprecated
    * @typedef {object} SetLabelsOptions Configuration options for Bucket#setLabels().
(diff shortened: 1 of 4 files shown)


danielbankhead (approved):

src/bucket.ts
```diff
@@ -3530,20 +3530,24 @@ class Bucket extends ServiceObject {
...
    * @param {object} metadata The bucket metadata.
    */
   /**
+   * @deprecated
    * @typedef {object} SetLabelsOptions Configuration options for Bucket#setLabels().
    * @property {string} [userProject] The ID of the project which will be
    *     billed for the request.
    */
   /**
+   * @deprecated Use setMetadata directly.
```
frankyn, line 3550: Is the expectation to update samples as well in a follow-up PR?

ddelgrosso1: Ah good catch, I will update that in this PR.

ddelgrosso1: Updated `addBucketLabel` sample to use `setMetadata`.

frankyn (approved):

danielbankhead (approved):

PR #2234, opened 2023-07-12
feat!: better typing for metadata

### LINKED ISSUE #1817: Add typing for `File.metadata`
#### Environment details

  - OS: Windows
  - Node.js version: 12.20.0
  - npm version: 6.14.8
  - `@google-cloud/storage` version: 5.18.2

#### Steps to reproduce

  1. Create instance of the `File` class.
  2. Access its `metadata` property in TypeScript.

The issue is that its [type](https://github.com/googleapis/nodejs-storage/blob/ac786c3ff6c516690b2113e7507b39e9da9eeb11/src/file.ts#L444) is [effectively](https://github.com/googleapis/nodejs-storage/blob/ac786c3ff6c516690b2113e7507b39e9da9eeb11/src/file.ts#L20) … [`any`](https://github.com/googleapis/nodejs-common/blob/f0346d5e5c72b1eb15df4d9caf5af6fea6d4fe71/src/service-object.ts#L53), whereas the [reference](https://googleapis.dev/nodejs/storage/latest/File.html#getMetadata-examples) and [example](https://github.com/googleapis/nodejs-storage/blob/ac786c3ff6c516690b2113e7507b39e9da9eeb11/samples/getMetadata.js#L47-L80) demonstrate the structure of the metadata.

Why not to add this structure to the typing?

ddelgrosso1:

Fixes #1817 🦕


src/acl.ts
```diff
@@ -94,6 +94,21 @@ export interface AccessControlObject {
...
+  domain?: string;
+  entity?: string;
+  entityId?: string;
+  generation?: string;
+  object?: string;
+  projectTeam?: {
+    projectNumber?: string;
+    team?: 'editors' | 'owners' | 'viewers';
+  };
+  role?: 'OWNER' | 'READER' | 'WRITER' | 'FULL_CONTROL';
```
frankyn, line 108: is this strict typing in that if new fields are added in the API the node.js client will need to be updated as well?

frankyn (commented):
Had a question on if this is a breaking change; I think so, and trying to understand how Sameena's concerns were concluded. I'm still reviewing the PR so just an initial comment.

src/acl.ts
ddelgrosso1: https://github.com/googleapis/nodejs-storage/pull/2234/files#diff-747d768492a54441f1682714c15429c572d74c031145c71672dec819ff94fbbbR109 this would act as a catch all for anything new so that it flows through without us having to make a change. However, we would obviously want to add anything new to the appropriate metadata interface to continue to support up to date typings.

package.json
```diff
@@ -58,7 +58,8 @@
     "compressible": "^2.0.12",
     "duplexify": "^4.0.0",
     "ent": "^2.2.0",
-    "gaxios": "^5.0.0",
+    "extend": "^3.0.2",
```
danielbankhead, line 61: Re-adding extend or merge conflict?

src/acl.ts
danielbankhead: I like the `unknown` too - encourages customers to validate it

danielbankhead (approved):
This is a great change - customers are gonna love it (including [myself](https://github.com/danielbankhead/website/blob/main/src/functions/types.ts)).

package.json
ddelgrosso1: Ah conflict, let me get rid of this again.

frankyn (commented):
I have a few comments; the types looked appropriate from my review and more wondering about the use of `unknown`

src/bucket.ts
```diff
@@ -79,19 +79,19 @@ interface CreateNotificationQuery {
...
 
-export type GetFilesResponse = [File[], {}, Metadata];
+export type GetFilesResponse = [File[], {}, unknown];
 export interface GetFilesCallback {
   (
     err: Error | null,
     files?: File[],
     nextQuery?: {},
-    apiResponse?: Metadata
+    apiResponse?: unknown
```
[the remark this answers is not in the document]
ddelgrosso1: Here is some [docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). 

The reason I had to go with unknown is sort of twofold, first previously `Metadata` was defined as `any` and was being used to represent ACTUAL metadata as well as a raw API response. Ideally we would have a well defined type for an API response but seeing as we are still using 2 HTTP libraries that is a bit more difficult currently.

src/bucket.ts
```diff
@@ -250,29 +268,91 @@ export interface GetBucketOptions extends GetConfig {
...
-  (err: ApiError | null, bucket: Bucket | null, apiResponse: Metadata): void;
+  (err: ApiError | null, bucket: Bucket | null, apiResponse: unknown): void;
 }
 
 export interface GetLabelsOptions {
   userProject?: string;
 }
 
-export type GetLabelsResponse = [Metadata];
+export type GetLabelsResponse = [unknown];
```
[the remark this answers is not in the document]
ddelgrosso1: See my comments above, the short of it is that previously API responses were `any` and due to the slightly different shapes of responses between Gaxios and teeny-request I chose to go with `unknown` until we can more clearly type API responses.

src/bucket.ts
```diff
@@ -250,29 +268,91 @@ export interface GetBucketOptions extends GetConfig {
...
 export interface GetLabelsCallback {
   (err: Error | null, labels: object | null): void;
 }
 
-export type GetBucketMetadataResponse = [Metadata, Metadata];
+export interface BucketMetadata extends BaseMetadata {
+  acl?: AclMetadata[] | null;
+  autoclass?: {
+    enabled?: boolean;
+    toggleTime?: string;
```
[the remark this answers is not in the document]
frankyn: nvm!

frankyn (approved):

PR #2237, opened 2023-07-13
fix!: Remove extraneous validation in favor of server errors

### LINKED ISSUE #1996: Refactor: remove client side validation in favor of returning server side errors
Per @frankyn an audit was conducted to identify client side validations / exceptions that should be removed in favor of relying server side validation. This will reduce potential issues should the server ever change or add additional checks or types. I will link the appropriate documents to the internal tracking ticket. This is considered a breaking change as users may be relying on the current messages thrown from the client.

ddelgrosso1:

Fixes #1996  🦕

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -53,7 +53,6 @@ import {
   PreconditionOptions,
   IdempotencyStrategy,
   BucketOptions,
-  ExceptionMessages,
 } from './storage';
 import {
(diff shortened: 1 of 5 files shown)


ddelgrosso1:
@frankyn this addresses the 3 findings from your scope of validation checks doc.

danielbankhead (approved):

PR #2291, opened 2023-09-05
feat: export `ApiError`

### LINKED ISSUE #2261: Missing error interface exports
**Environment details**

-  OS: Win x64 22H2
-  Node.js version: 18.14.0
-  npm version: 9.3.1
-  @google-cloud/storage version: 6.9.5 / 7.0.1

**Steps to reproduce**

1. Try to handle errors for bucket.upload with strict typescript or with enabled [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) rule.

2. Can not safely access e.code or e.errors without `ApiError` interface

It appears sdk is not exporting Error interfaces these in the index file: https://github.com/googleapis/nodejs-storage/blob/main/src/nodejs-common/util.ts#L248

**Workaround**

In version 7.0.1 it's possible to import directly from "@google-cloud/storage/build/src/nodejs-common"

danielbankhead:
Fixes #2261 🦕

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,7 +56,7 @@
  * region_tag:storage_quickstart
  * Full quickstart example:
  */
-export * from './nodejs-common';
+export {ApiError} from './nodejs-common';
 export {
   AclMetadata,
   AccessControlObject,
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,6 +56,7 @@
  * region_tag:storage_quickstart
  * Full quickstart example:
  */
+export * from './nodejs-common';
 export {
   AclMetadata,
   AccessControlObject,


src/index.ts
```diff
@@ -56,6 +56,7 @@
  * region_tag:storage_quickstart
  * Full quickstart example:
  */
+export * from './nodejs-common';
```
ddelgrosso1, line 59: Do you think we should export everything from common? It seems like most things have no real use outside of storage and I'm nervous about making it all publicly exposed.

danielbankhead: Fair - exporting `ApiError` only

ddelgrosso1 (approved):

PR #1892, opened 2022-04-27
Build: Native TypeScript Support in Mocha

danielbankhead:
Adding support for `mocha` to directly run TypeScript:
- Saves time by not having a separate `compile` step
- Allows users to run `mocha` more easily in IDEs (e.g. IntelliJ's right-click test functionality)

🦕

PR #2296, opened 2023-09-07
feat: support building in CJS and ESM formats

ddelgrosso1:


danielbankhead (approved):

src/resumable-upload.ts
```diff
@@ -769,7 +775,9 @@ export class Upload extends Writable {
     const headers: GaxiosOptions['headers'] = {
       'x-goog-api-client': `${getRuntimeTrackingString()} gccl/${
         packageJson.version
-      } gccl-invocation-id/${this.currentInvocationId.chunk}`,
+      }-${getModuleFormat()} gccl-invocation-id/${
```
danielduhh, line 778: Not a blocker but it would be easier to cipher through this if the header changes was decoupled in a separate PR

danielduhh (approved):

internal-tooling/performPerformanceTest.ts
```diff
@@ -75,44 +78,44 @@ async function main() {
...
     appBufferSize: NODE_DEFAULT_HIGHWATER_MARK_BYTES,
     crc32cEnabled: false,
     md5Enabled: false,
     api: 'JSON',
     elapsedTimeUs: 0,
     cpuTimeUs: -1,
     status: 'OK',
-    chunkSize: argv.range_read_size,
-    workers: argv.workers,
+    chunkSize: argv.range_read_size as number,
```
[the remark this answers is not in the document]
ddelgrosso1: They are related to updates in TypeScript not ESM.

.jsdoc.json
```diff
@@ -0,0 +1,29 @@
...
+    "destination": "./docs/"
+  },
+  "plugins": ["plugins/markdown", "jsdoc-region-tag"],
+  "source": {
+    "excludePattern": "(^|\\/|\\\\)[._]",
+    "include": ["build/cjs/src"],
+    "includePattern": "\\.js$"
+  },
+  "templates": {
+    "copyright": "Copyright 2019 Google, LLC.",
```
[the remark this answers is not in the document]
ddelgrosso1: This was just a renamed file / moved file. Should I still update the year?

conformance-test/libraryMethods.ts
```diff
@@ -21,6 +21,7 @@ import {
   deleteTestFile,
 } from './testBenchUtil';
 import * as uuid from 'uuid';
+import {getDirName} from '../src/util.js';
```
sofisl, line 24: NB, you could save dirName = getDirName(), that way you improve performance vs. memory (seems like you call it in multiple spots)

sofisl (approved):

sofisl:
Exciting stuff!

src/util.ts
```diff
@@ -192,6 +199,32 @@ export function getRuntimeTrackingString(): string {
   }
 }
 
+export function getDirName() {
```
danielbankhead, line 202: Do you mind testing this with `esbuild`? I'm concerned the fallback logic might break like it broke in a previous PR:
- https://github.com/googleapis/nodejs-storage/issues/1934
- https://github.com/googleapis/nodejs-storage/pull/1892/files#diff-10d5d48c419530b01c345d22a42140d1a726097de84b7525de74f8d5f6fbccc7R619-R625

If it doesn't work perhaps we can have this as a separate .cjs file and import it from both module formats for now.

src/util.ts
```diff
@@ -192,6 +199,32 @@ export function getRuntimeTrackingString(): string {
...
+  }
+
+  return dirToUse;
+}
+
+export function getPackageJSON() {
+  const dirName = getDirName();
+  return JSON.parse(
+    fs
+      .readFileSync(path.join(dirName, '..', '..', '..', 'package.json'))
```
danielbankhead, line 219: I _think_ this might be an issue with `esbuild`; perhaps we can have this as a separate .cjs file and import it from both module formats for now.

I considered suggesting we import this via:

```ts
await import('./package.json', {assert: {type: 'json'}});
```

However, it may produce 'experimental' warnings in some versions of Node.js.

src/util.ts
```diff
@@ -192,6 +199,32 @@ export function getRuntimeTrackingString(): string {
...
 }
 
+export function getDirName() {
+  let dirToUse = '';
+  try {
+    dirToUse = __dirname;
+  } catch (e) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    dirToUse = path.dirname(fileURLToPath(import.meta.url));
```
danielbankhead, line 209: This seems to only be relative to the current file rather than the caller's directory, which is fine but worth noting. [Optional] Alternatively we could point this to the 'root' of the module (or `src`) for consistency and rename this function appropriately.

danielbankhead (commented):
Looks great so far

src/util.ts
ddelgrosso1: I'll give this a test with esbuild and report back.

src/util.ts
ddelgrosso1: I think that import only works with higher versions of node. I remember bumping into this issue with 14. I'll try and dig up the exact version number.

src/util.ts
ddelgrosso1: Added a different helper to fix esbuild problem with package.json.

PR #2303, opened 2023-09-14
feat: add headers option to MPU

ddelgrosso1:

### DIFF
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -93,6 +93,7 @@ export interface UploadFileInChunksOptions {
   uploadId?: string;
   partsMap?: Map<number, string>;
   validation?: 'md5' | false;
+  headers?: {[key: string]: string};
 }
 
 export interface MultiPartUploadHelper {
@@ -100,7 +101,7 @@ export interface MultiPartUploadHelper {
   fileName: string;
   uploadId?: string;
   partsMap?: Map<number, string>;
-  initiateUpload(): Promise<void>;
+  initiateUpload(headers?: {[key: string]: string}): Promise<void>;
   uploadPart(
     partNumber: number,
     chunk: Buffer,
@@ -185,13 +186,14 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
    *
    * @returns {Promise<void>}
    */
-  async initiateUpload(): Promise<void> {
+  async initiateUpload(headers: {[key: string]: string} = {}): Promise<void> {
     const url = `${this.baseUrl}?uploads`;
     return retry(async bail => {
       try {
         const res = await this.authClient.request({
           method: 'POST',
           url,
+          headers,
         });
         if (res.data && res.data.error) {
           throw res.data.error;
@@ -611,6 +613,8 @@ export class TransferManager {
    * @property {string} [uploadId] If specified attempts to resume a previous upload.
    * @property {Map} [partsMap] If specified alongside uploadId, attempts to resume a previous upload from the last chunk
    * specified in partsMap
+   * @property {object} [headers] headers to be sent when initiating the multipart upload.
+   * See {@link https://cloud.google.com/storage/docs/xml-api/post-object-multipart#request_headers| Request Headers: Initiate a Multipart Upload}
    * @experimental
    */
   /**
@@ -665,7 +669,7 @@ export class TransferManager {
     let promises: Promise<void>[] = [];
     try {
       if (options.uploadId === undefined) {
-        await mpuHelper.initiateUpload();
+        await mpuHelper.initiateUpload(options.headers);
       }
       const startOrResumptionByte = mpuHelper.partsMap!.size * chunkSize;
       const readStream = createReadStream(filePath, {
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -91,6 +91,7 @@ export interface UploadFileInChunksOptions {
   uploadName?: string;
   maxQueueSize?: number;
   uploadId?: string;
+  abortExisting?: boolean;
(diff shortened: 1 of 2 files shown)


test/transfer-manager.ts
```diff
@@ -15,219 +15,118 @@
...
-class FakeFile {
-  calledWith_: IArguments;
-  bucket: Bucket;
-  name: string;
-  options: FileOptions;
-  metadata: FileMetadata;
-  createWriteStream: Function;
-  isSameFile = () => false;
-  constructor(bucket: Bucket, name: string, options?: FileOptions) {
-    // eslint-disable-next-line prefer-rest-params
```
danielbankhead, line 74: My favorite part of the PR 💣

danielbankhead (approved):

src/transfer-manager.ts
```diff
@@ -664,8 +696,12 @@ export class TransferManager {
     let partNumber = 1;
     let promises: Promise<void>[] = [];
     try {
+      if (options.abortExisting && options.uploadId) {
```
danielduhh, line 699: How does this option work? If I set abortExisting to true, and have a upload ID, it will cancel the the upload? Shouldn't it only do this if it fails?

test/transfer-manager.ts
```diff
@@ -443,12 +348,69 @@ describe('Transfer Manager', () => {
...
+      };
+
+      await transferManager.uploadFileInChunks(
+        path,
+        {headers: headersToAdd},
+        mockGeneratorFunction
+      );
+    });
+
+    it('should call abortUpload when passed the option and uploadID', async () => {
```
danielduhh, line 387: This seems like a hack? Why not just expose the abort API and ask for an id? We could document that users should catch failures and call this method (since we're not doing it for them)/

src/transfer-manager.ts
```diff
@@ -281,6 +285,30 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
     }, this.retryOptions);
   }
 
+  /**
+   * Aborts an multipart upload that is in progress. Once aborted, any parts in the process of being uploaded fail,
+   * and future requests using the upload ID fail.
+   *
+   * @returns {Promise<void>}
+   */
+  async abortUpload(): Promise<void> {
```
frankyn, line 294: Can you make this private?

src/transfer-manager.ts
frankyn: oh, looks like you can abort an upload using `uploadFileInChunks` when you set abortExisting=True and uploadId is set; based on 

>  [abortExisting] boolean to indicate if an in progress upload should be aborted. uploadID must also be supplied in order to abort the upload.

I guess if this delete in context of this method; the upload id would be known and isn't required like this.

ddelgrosso1: @danielduhh correct the user would have to pass back the `uploadId` and set the `abortExisting` option. The way it works currently is if there is a failure the return has the `uploadId` and mapping of the completed parts that were uploaded.

test/transfer-manager.ts
ddelgrosso1: When we implemented this we decided not to expose the actual XML implementation to avoid users accessing it directly. This was done because at the time we did not know if other MPU implementations would come along and the default would be changed (i.e. JSON).

danielduhh: Yeah it feels odd to have this API do two completely different things based on the parameters.
Instead, I would consider following python and [catching any errors thrown](https://github.com/googleapis/python-storage/pull/1115/files/43fd95c91fdafe61eac65e46a471684586e5a615#diff-39908d1c6ff23e82acdf23e96d17f7237b47a41b53d4de55adddca92b3fa6df2R1045) by MPU with a note:

```
The library will attempt to cancel uploads that fail due to an exception.
    If the upload fails in a way that precludes cancellation, such as a
    hardware failure, process termination, or power outage, then the incomplete
    upload may persist indefinitely. To mitigate this, set the
    `AbortIncompleteMultipartUpload` with a nonzero `Age` in bucket lifecycle
    rules, or refer to the XML API documentation linked above to learn more
    about how to list and delete individual downloads.
    ```

Comment (ddelgrosso1) [reply test/transfer-manager.ts:387]: This goes back to the debate of autodelete vs not. If we want to proceed with autodelete, I'll just move the logic internally to call the delete function.

Comment (danielduhh) [reply test/transfer-manager.ts:387]: @frankyn ?

Comment (frankyn) [reply test/transfer-manager.ts:387]: I set up a discussion with @andrewsg and @ddelgrosso1 to suss this out today; will report back;

File: src/transfer-manager.ts [hunk 5]
```diff
@@ -611,6 +639,10 @@ export class TransferManager {
    * @property {string} [uploadId] If specified attempts to resume a previous upload.
    * @property {Map} [partsMap] If specified alongside uploadId, attempts to resume a previous upload from the last chunk
    * specified in partsMap
+   * @property {object} [headers] headers to be sent when initiating the multipart upload.
+   * See {@link https://cloud.google.com/storage/docs/xml-api/post-object-multipart#request_headers| Request Headers: Initiate a Multipart Upload}
+   * @property {boolean} [autoAbortFailure] boolean to indicate if an in progress upload should be aborted automatically upon failure. If not set,
+   * failures will be automatically aborted.
```
frankyn, line 645: `upload session will be automatically aborted`

frankyn (commented):
one nit;

danielduhh (approved):
Nice!

src/transfer-manager.ts
ddelgrosso1: Updated, thanks @frankyn

frankyn (approved):

PR #2333, opened 2023-10-16
feat: Resume Resumable Uploads

### LINKED ISSUE #2162: Resume an incomplete upload
I have a node server/api that sits in front of cloud storage. It authenticates the user and requires some additional metadata before streaming the upload into cloud storage. 

I'm trying abstract away and hide the fact that we're using GCS from our clients/consumers. I'm therefore building an api that mirrors the resumable upload api and that will allow clients to chunk/resume an interrupted file upload.

Please correct me if I'm wrong but this client library doesn't seem to accommodate this use case.

A client first sends a post request to my api to initiate a resumable upload e.g. `POST /files`

The server then starts a resumable upload and places an object in the bucket. We store the session URI in our database for later retrieval. We then respond with the an upload id that we generate.

A client may then resume an upload via `PATCH /files/{id}` with e.g. `Content-Range: bytes 456-987/1234` but there's no way for me to resume an already started resumable upload using the information in that header given how the storage library is currently written.

Some example code:

```ts
// Resume an interrupted/incomplete upload
server.patch('/files/:id', async (req, res) => {
  const headers = req.headers;
  const contentLength = parseContentLength(headers['content-length']);
  const contentRange = parseContentRange(headers['content-range']);
  const rangeStart = contentRange.rangeStart ?? 0;

  // Nowhere to pass these that I can tell
  // const rangeEnd = contentRange.rangeEnd;
  // const totalSize = contentRange.size;

  const id = req.params.id;
  const record = await db.files.get(id);

  if (!record) {
    throw { statusCode: 404, message: `No such file: ${id}` };
  }

  if (record.uploadSessionURIExpiresAt <= Date.now()) {
    throw { statusCode: 410, message: `No such file: ${id}` };
  }

  const gcsObjectPath = `${gcsObjectPrefix}/${id}`;
  const object = storage.bucket(gcsBucket).file(gcsObjectPath);

  const uploadStream = object.createWriteStream({
    resumable: true,
    // no way for me to say that offset is the next byte after what we expect has already been uploaded
    // no way for me to tell the library the totalSize if it is now known.
    offset: rangeStart,
    uri: record.uploadSessonURI,
  });

  await pipe(req.raw, uploadStream);
  
  // Produces the following error:
  //
  //   The uploaded data did not match the data from the server.
  //   As a precaution, the file has been deleted.
  //   To be sure the content is the same, you should try uploading the file again.

  res.status(200).send({ ok: true });
});
```

Looking at the internals it seems like the behaviour is mainly determined by `startUploading` in [resumable-uploads.ts](https://github.com/googleapis/nodejs-storage/blob/main/src/resumable-upload.ts#L635)

danielbankhead:
Additionally, support uploading partial resumable uploads.

Fixes #2162 🦕


src/file.ts
```diff
@@ -469,6 +470,8 @@ export enum FileExceptionMessages {
   UPLOAD_MISMATCH = `The uploaded data did not match the data from the server.
     As a precaution, the file has been deleted.
     To be sure the content is the same, you should try uploading the file again.`,
+  MD5_RESUMED_UPLOAD = 'MD5 cannot be used with a resumed resumable upload as MD5 cannot be extended from an existing value',
```
ddelgrosso1, line 473: Totally nit picky but should be call resuming something else, i.e. continued resumable upload. Resumed resumable upload sounds tongue twisterish.

ddelgrosso1:
Since this is a larger change let's walk through it when you have some time.

ddelgrosso1 (approved):

PR #2334, opened 2023-10-17
fix: add user-agent header to transfer manager and resumable upload

ddelgrosso1:

### DIFF
--- a/test/nodejs-common/util.ts
+++ b/test/nodejs-common/util.ts
@@ -1869,17 +1869,6 @@ describe('common/util', () => {
     });
   });
 
-  describe('getUserAgentFromPackageJson', () => {
-    it('should format a User Agent string from a package.json', () => {
-      const userAgent = util.getUserAgentFromPackageJson({
-        name: '@google-cloud/storage',
-        version: '0.1.0',
-      });
-
-      assert.strictEqual(userAgent, 'gcloud-node-storage/0.1.0');
-    });
-  });
-
   describe('maybeOptionsOrCallback', () => {
     it('should allow passing just a callback', () => {
       const optionsOrCallback = () => {};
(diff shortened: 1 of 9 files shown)


test/nodejs-common/util.ts
```diff
@@ -1869,17 +1869,6 @@ describe('common/util', () => {
...
   });
 
-  describe('getUserAgentFromPackageJson', () => {
-    it('should format a User Agent string from a package.json', () => {
-      const userAgent = util.getUserAgentFromPackageJson({
-        name: '@google-cloud/storage',
-        version: '0.1.0',
-      });
-
-      assert.strictEqual(userAgent, 'gcloud-node-storage/0.1.0');
```
frankyn, line 1879: Did the client use gcloud-node-storage before?

test/resumable-upload.ts
```diff
@@ -53,6 +53,7 @@ const CHUNK_SIZE_MULTIPLE = 2 ** 18;
 const queryPath = '/?userProject=user-project-id';
 const X_GOOG_API_HEADER_REGEX =
   /^gl-node\/(?<nodeVersion>[^W]+) gccl\/(?<gccl>[^W]+) gccl-invocation-id\/(?<gcclInvocationId>[^W]+) gccl-gcs-cmd\/(?<gcclGcsCmd>[^W]+)$/;
+const USER_AGENT_REGEX = /^gcloud-node-storage\/(?<libVersion>[^W]+)$/;
```
frankyn, line 56: I'm confused, i think this should just be `gcloud-node/<libVersion>`;

ddelgrosso1: So I just moved around the existing code for setting `user-agent` (https://github.com/googleapis/nodejs-storage/blob/f20ef3cb7bf8cbdda988e792edc7abd9a1b516f1/src/nodejs-common/util.ts#L1010). Looks like it has always been set to `gcloud-node-storage` which seems to line up with the examples from the Headers document.

frankyn: TIL that this is different in Node.js thanks!

frankyn (approved):
Similar to idempotency token header; we need a clearer way to verify clients are behaving the way we want them to by testing against testbench (maybe) for presence of expected headers.

danielduhh:
@ddelgrosso1 could you please add the "why" context to the PR description, or link to a bug?

ddelgrosso1:
The rationale for this PR is that currently in certain situations (transfer manager, resumable uploads) `user-agent` is not being set by this library. Instead the auth library sets `user-agent`. This is causing incorrect attribution of metrics to versions of storage that do not exist, i.e. (9.0.0, 9.1.0). This PR should fix this attribution problem.

PR #2338, opened 2023-10-19
fix: Close Open Handle for Empty Objects

danielbankhead:
🦕


src/file.ts
```diff
@@ -2173,8 +2173,14 @@ class File extends ServiceObject<File, FileMetadata> {
         .on('end', () => {
           // In the case of an empty file no data will be received before the end event fires
           if (!receivedData) {
-            fs.openSync(destination, 'w');
```
danielbankhead, line 2176: These open handles were never closed out.

ddelgrosso1: Good catch!

ddelgrosso1 (approved):

PR #2365, opened 2023-11-15
feat: implement object retention lock for bucket / files

### LINKED ISSUE #2349: Add support for object retention
Please implement the object retention feature. The following operations must be supported:

- Create a bucket with object retention enabled by using the `enableObjectRetention` bucket create boolean option at bucket create time
- Get the object retention enabled status of a bucket
- Create an object with retention enabled (on an retention enabled bucket) by populating the `retention` object at object create time (retention object has the fields `mode` which can be Unlocked or Locked, and `retainUntilTime` which is a timestamp)
- Update retention for an existing object with a patch operation, by passing the `overrideUnlockedRetention` boolean patch option
- Get the retention policy of an object
- Remove retention for an existing object the same way as above

Details in go/gcs-object-retention-client-request

You can use the [java implementation](https://github.com/googleapis/java-storage/pull/2277) as a reference.

The kokoro test project should be allowlisted shortly, let me know if you need access to an allowlisted project for your own testing. 

We're aiming to get this in by the November code freeze, let me know if you need any help.

ddelgrosso1:

Fixes #2349  🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -423,6 +423,7 @@ export interface FileMetadata extends BaseMetadata {
   };
   customTime?: string;
   eventBasedHold?: boolean | null;
+  readonly eventBasedHoldReleaseTime?: string;
   generation?: string | number;
   kmsKeyName?: string;
   md5Hash?: string;
@@ -436,6 +437,10 @@ export interface FileMetadata extends BaseMetadata {
     entity?: string;
     entityId?: string;
   };
+  retention?: {
+    retainUntilTime?: string;
+    mode?: string;
+  } | null;
   retentionExpirationTime?: string;
   size?: string | number;
   storageClass?: string;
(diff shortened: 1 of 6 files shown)


JesseLovelace (approved):

src/file.ts
```diff
@@ -423,6 +423,7 @@ export interface FileMetadata extends BaseMetadata {
   };
   customTime?: string;
   eventBasedHold?: boolean | null;
+  readonly eventBasedHoldReleaseTime?: string;
```
danielbankhead, line 426: Where do we use/set this?

ddelgrosso1: It is not user setable. It is a field that will be added as a fast follow that is only returned if an event-based hold was released.

danielbankhead (approved):

PR #2373, opened 2023-11-28
fix: TransferManager#downloadFileInChunks issues

### LINKED ISSUE #2372: Possible TransferManager#downloadFileInChunks issues
I was going to use [`TransferManager#downloadFileInChunks`](https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/src/transfer-manager.ts#L633) to download a 20GB+ file and thought I'd investigate how it works (as [the docs](https://googleapis.dev/nodejs/storage/latest/global.html#DownloadFileInChunksOptions) aren't that great around it - it's not the most clear where the file ends up).

But I think there's multiple issues around the way the response is returned, the CRC32C validation and also memory usage.

https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/src/transfer-manager.ts#L680

First the type of the `result` isn't correct, it'll return `Buffer[]` and not `[Buffer]` (which is what `DownloadResponse` is).

https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/src/transfer-manager.ts#L682

Second the result of the `CRC32C ` validation isn't used - shouldn't it be compared against the [`fileInfo`](https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/src/transfer-manager.ts#L648C11-L648C19) metadata's CRC32 value?

Finally, related to the first issue, the function returns the complete contents of the file into memory (as an array of `Buffer`s`)? Won't this just cause OOM issues with any large GB sized file?

Happy to do a PR to address the issues if the above assumptions are correct. In relation to the complete contents of the file being returned, could have a flag on `DownloadFileInChunksOptions` to signal that function should just resolve to `void`? Or a lot simpler (but breaking), just always resolve to `void` and let the user load the contents of the downloaded file into memory if that's what they want...

Note, I've checked the tests out, and they're [pretty basic](https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/test/transfer-manager.ts#L238-L293) only checking that `CRC32C.fromFile` is called, and not that the actual result is used in validation.

EDIT:

https://github.com/googleapis/nodejs-storage/blob/7a96ce6f764076a14f0961623e2dec2ce8893dd7/src/transfer-manager.ts#L686

I've also noticed that the `FileHandle#close()` isn't `await`ed and should probably be done _after_ all the chunks are downloaded (ie. after `await Promise.all(promises)`) and _before_ the CRC32C is computed. Is this assumption correct and also needs to be changed?

rhodgkins:
Fixes #2372 🦕

### DIFF
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -111,7 +111,7 @@ export interface DownloadFileInChunksOptions {
   chunkSizeBytes?: number;
   destination?: string;
   validation?: 'crc32c' | false;
-  noReturnData?: true;
+  noReturnData?: boolean;
 }
 
 export interface UploadFileInChunksOptions {
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -15,7 +15,13 @@
  */
 
 import {Bucket, UploadOptions, UploadResponse} from './bucket.js';
-import {DownloadOptions, DownloadResponse, File} from './file.js';
+import {
+  DownloadOptions,
+  DownloadResponse,
+  File,
+  FileExceptionMessages,
+  RequestError,
+} from './file.js';
 import pLimit from 'p-limit';
 import * as path from 'path';
 import {createReadStream, promises as fsp} from 'fs';
@@ -105,6 +111,7 @@ export interface DownloadFileInChunksOptions {
   chunkSizeBytes?: number;
   destination?: string;
   validation?: 'crc32c' | false;
+  noReturnData?: true;
 }
 
 export interface UploadFileInChunksOptions {
@@ -639,7 +646,8 @@ export class TransferManager {
     let limit = pLimit(
       options.concurrencyLimit || DEFAULT_PARALLEL_CHUNKED_DOWNLOAD_LIMIT
     );
-    const promises: Promise<{bytesWritten: number; buffer: Buffer}>[] = [];
+    const noReturnData = Boolean(options.noReturnData);
+    const promises: Promise<Buffer | void>[] = [];
     const file: File =
       typeof fileOrName === 'string'
         ? this.bucket.file(fileOrName)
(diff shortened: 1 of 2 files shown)


src/transfer-manager.ts
```diff
@@ -105,6 +111,7 @@ export interface DownloadFileInChunksOptions {
   chunkSizeBytes?: number;
   destination?: string;
   validation?: 'crc32c' | false;
+  noReturnData?: true;
```
ddelgrosso1, line 114: To prevent this from altering existing behavior, i.e. backwards compatible change it will need to default to false.

ddelgrosso1: Also we will need to add this to the JSDoc fields above the `downFileInChunks` method.

rhodgkins: Ah, I was just going with no value at all or `true` so existing users won't be affected? (Sort of matching the style of the `validation` field above.)

src/transfer-manager.ts
```diff
@@ -667,24 +675,39 @@ export class TransferManager {
...
+        const mismatchError = new RequestError(
+          FileExceptionMessages.DOWNLOAD_MISMATCH
+        );
+        mismatchError.code = 'CONTENT_DOWNLOAD_MISMATCH';
+        throw mismatchError;
+      }
     }
+    if (noReturnData) return;
+    return [Buffer.concat(chunks as Buffer[], size)];
   }
```
rhodgkins, line 711: I forgot to add this is possible a breaking change (because of the incorrect return before).

Before it returned `Buffer[]` which didn't match the declared return type `[Buffer]`, so to match it I've combined all the chunks into a single buffer and matched the return type. Happy to leave it returning `chunks` (ie. `Buffer[]`) and correct the return type instead but thought returning the correct/expected thing made more sense...!

ddelgrosso1: We wouldn't really consider this to be breaking as it is documented with the correct intention but actually has a bug and returns the wrong type. This change is OK.

src/transfer-manager.ts
ddelgrosso1: Being as we currently return data, albeit the incorrect type per the definition, we want this to default to returning the data to maintain that compatibility.

ddelgrosso1 (approved):
LGTM. We appreciate the contribution!

src/transfer-manager.ts
```diff
@@ -604,6 +611,7 @@ export class TransferManager {
    * to use when downloading the file.
    * @property {number} [chunkSizeBytes] The size in bytes of each chunk to be downloaded.
    * @property {string | boolean} [validation] Whether or not to perform a CRC32C validation check when download is complete.
+   * @property {boolean} [noReturnData] Wether or not to return the downloaded data. A `true` value here would be useful for files with a size that will not fit into memory.
```
danielbankhead, line 614:
```suggestion
   * @property {boolean} [noReturnData] Whether or not to return the downloaded data. A `true` value here would be useful for files with a size that will not fit into memory.
```

src/transfer-manager.ts
```diff
@@ -667,24 +676,39 @@ export class TransferManager {
...
-    let results: DownloadResponse;
+    let chunks: Array<Buffer | void>;
     try {
-      const data = await Promise.all(promises);
-      results = data.map(result => result.buffer) as DownloadResponse;
-      if (options.validation === 'crc32c') {
-        await CRC32C.fromFile(filePath);
-      }
-      return results;
+      chunks = await Promise.all(promises);
```
danielbankhead, line 695: [Optional]: It might be cleaner to only set `chunks` if `!noReturnData` to avoid casting later, but no biggie.

danielbankhead (commented):
LGTM - one nit, one suggestion

src/transfer-manager.ts
rhodgkins: Going to leave that as is if that's OK, imo it's clear on the return with `if (!noReturnData)` (and not returning anything) rather than checking `if (chunks)` and then returning the `Buffer.concat`.

rhodgkins:
Added some extra test cases too...

src/transfer-manager.ts
```diff
@@ -604,15 +611,16 @@ export class TransferManager {
...
    */
   /**
    * Download a large file in chunks utilizing parallel download operations. This is a convenience method
    * that utilizes {@link File#download} to perform the download.
    *
-   * @param {object} [file | string] {@link File} to download.
+   * @param {File | string} fileOrName {@link File} to download.
    * @param {DownloadFileInChunksOptions} [options] Configuration options.
-   * @returns {Promise<DownloadResponse>}
+   * @returns {Promise<void | DownloadResponse>}
```
rhodgkins, line 623: ...and fixed JSDoc declaration

danielbankhead (approved):

Issue #2197, opened 2023-05-13
"Could not load the default credentials" when using custom apiEndpoint with getSignedUrl.


psigen:
#### Environment details

  - OS: Ubuntu 22.04
  - Node.js version: v18.15.0
  - npm version: 9.5.0
  - `@google-cloud/storage` version: 6.10.0

#### Steps to reproduce

Run the following code using a storage emulator host setting in the environment:
```
STORAGE_EMULATOR_HOST=http://localhost:9199/
```

```
import { Storage } from "@google-cloud/storage";

const storage = new Storage({
  useAuthWithCustomEndpoint: false,
});

const expiration = new Date();
const expirationSecs = 60 * 10; // 10 minutes
expiration.setSeconds(expirationSecs);

const options = {
  version: "v4" as const,
  action: "read" as const,
  expires: expiration,
};

const [url] = await storage.bucket(BUCKET_NAME)
  .file(path)
  .getSignedUrl(options);
```

This will return the following error about credentials, even though the emulator doesn't use credentials, so they aren't needed to sign the URL:
```
Error: Could not load the default credentials. Browse to https://cloud.google.com/docs/authentication/getting-started for more information.
    at GoogleAuth.getApplicationDefaultAsync (/app/api/index.js:65733:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async GoogleAuth.getClient (/app/api/index.js:66050:20)
    at async GoogleAuth.getCredentialsAsync (/app/api/index.js:66009:24)
    at async sign (/app/api/index.js:93784:33)
```

This was previously mentioned in a comment on this previous ticket: https://github.com/googleapis/nodejs-storage/issues/1373#issuecomment-1140351554

ddelgrosso1:
@danielbankhead I poked into this a bit and it appears that the problem comes in that there doesn't be a way to tell the auth library to "faux" sign a request. Is there a clean way we can bypass this in storage when the user does not want to use authentication?

danielbankhead:
[discussed offline, sharing here] We may need to take a similar approach as https://github.com/googleapis/gcs-resumable-upload/pull/416 for this issue.

BrennaEpp:
Hijacking this issue to track the implementation of this feature:

When a custom endpoint is set on the client, Sign URLs and Post Policies should automatically use that domain.

Optional: allow users to set a custom hostname for Sign URL and Post policy methods

See: go/signurl-custom-endpoint-gcs

[...]

PR #2397, opened 2024-01-29
feat: Base TPC Support

danielbankhead:
Related: #2197

🦕


src/signer.ts
```diff
@@ -50,6 +54,11 @@ export interface GetSignedUrlConfigInternal {
   contentType?: string;
   bucket: string;
   file?: string;
+  /**
+   * @example
+   * 'https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/'
+   */
+  signingEndpoint?: string;
```
BrennaEpp, line 61: Interesting addition! What use-case do you envision here?

src/signer.ts
```diff
@@ -137,7 +143,7 @@ export class URLSigner {
     if (cfg.cname) {
       customHost = cfg.cname;
     } else if (isVirtualHostedStyle) {
-      customHost = `https://${this.bucket.name}.storage.googleapis.com`;
+      customHost = `https://${this.bucket.name}.storage.${this.universeDomain}`;
```
BrennaEpp, line 146: Not sure I follow the logic here; is there a way to provide something like "localhost:9000" and create a signed URL that way (either through an endpoint you give the client or otherwise)?

BrennaEpp (commented):
I'd like to see some tests around signed URLs, but that may or may not apply to this PR depending if I'm understanding it correctly. Left some questions to figure it out. Thanks!

src/signer.ts
danielbankhead: Customers can use this to set customize the signed URL endpoint for the signed URL V2 & V4 flows, which is passed to auth via `auth.sign`/`authClient.sign`

src/signer.ts
danielbankhead: Yep, `localhost:9000` can be used via the `signingEndpoint` param to be called to generate a signed URL. Note that a part of the V4 flow the custom host (`cname`) is included in the signed body itself:

https://github.com/googleapis/nodejs-storage/blob/0d3e2111df5c3f7570c07b986c7b4f6d560e888a/src/signer.ts#L251-L254

https://github.com/googleapis/nodejs-storage/blob/0d3e2111df5c3f7570c07b986c7b4f6d560e888a/src/signer.ts#L279

https://github.com/googleapis/nodejs-storage/blob/0d3e2111df5c3f7570c07b986c7b4f6d560e888a/src/signer.ts#L303-L310

https://github.com/googleapis/nodejs-storage/blob/0d3e2111df5c3f7570c07b986c7b4f6d560e888a/src/signer.ts#L312-L328

src/signer.ts
```diff
@@ -102,20 +111,17 @@ const SEVEN_DAYS = 7 * 24 * 60 * 60;
 
 /**
  * @const {string}
- * @private
+ * @deprecated - unused
```
ddelgrosso1, line 114: Same question as above

src/file.ts
```diff
@@ -302,7 +312,7 @@ export enum ActionToHTTPMethod {
 }
 
 /**
- * @private
+ * @deprecated - no longer used
```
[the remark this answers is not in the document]
danielbankhead: I would suggest no since we've previously marked it as `private`

src/signer.ts
danielbankhead: Same here

src/storage.ts
```diff
@@ -692,7 +693,9 @@ export class Storage extends Service {
    * @param {StorageOptions} [options] Configuration options.
    */
   constructor(options: StorageOptions = {}) {
-    let apiEndpoint = 'https://storage.googleapis.com';
+    const universe = options.universeDomain || DEFAULT_UNIVERSE;
```
danielbankhead, line 696: I’ll add a test for this.

danielbankhead:
> I'd like to see some tests around signed URLs, but that may or may not apply to this PR depending if I'm understanding it correctly. Left some questions to figure it out. Thanks!

Added some tests

JesseLovelace (approved):

src/signer.ts
BrennaEpp: I guess I'm confused by the example - it seems like you are passing in what to call to create the signURL, not where the signed URL endpoint will point to.

danielbankhead: Yea, the `cname` is what the url will point to

BrennaEpp: `cname` works for all styles, not just Bucketbound?

danielbankhead: Yep, see:

https://github.com/googleapis/nodejs-storage/blob/0d3e2111df5c3f7570c07b986c7b4f6d560e888a/src/signer.ts#L251-L254

BrennaEpp: I guess I'm confused here...
`cname` is a specific concept that is tied to [bucket bound hostnames]( https://cloud.google.com/storage/docs/request-endpoints#cname), and while it does seem to allow arbitrary hosts, the naming is confusing. Ultimately it is not a requirement that a hostname can be set directly when signing, so I think this is okay though.

What I think this PR misses addressing is the FR of when a custom endpoint is set on the client, Sign URLs (v2 and v4) and Post Policies automatically use that endpoint. I think if it did that, specifically setting the URL in the universe domain case in the quoted code may be unnecessary (I guess it depends on how you implement it).

danielbankhead: I figured out the gap in understanding - should the `hostname` be included in the the signature?

src/file.ts
```diff
@@ -2763,18 +2775,23 @@ class File extends ServiceObject<File, FileMetadata> {
...
         const signatureHex = Buffer.from(signature, 'base64').toString('hex');
+        const universe = this.parent.storage.universeDomain;
         fields['policy'] = policyBase64;
         fields['x-goog-signature'] = signatureHex;
 
         let url: string;
+
         if (options.virtualHostedStyle) {
-          url = `https://${this.bucket.name}.storage.googleapis.com/`;
+          url = `https://${this.bucket.name}.storage.${universe}/`;
```
BrennaEpp, line 2790: So I think this still doesn't address the custom endpoint on client use-case (ie. when an endpoint is set on the client, then any signURL created using that client will point to that endpoint instead of the default storage.googleapis.com).

However, that can be a separate FR as this works for TPC. I'm happy as long as the TPC portion of the recently added conformance tests pass, or there's a similar unit test which verifies that calling sign URL will produce a url pointing to the TPC universe (you can extract the signature from the conformance test to check that as well)

danielbankhead: > So I think this still doesn't address the custom endpoint on client use-case (ie. when an endpoint is set on the client, then any signURL created using that client will point to that endpoint instead of the default storage.googleapis.com).

Ah, for post policies too. I'll add that.

src/resumable-upload.ts
```diff
@@ -356,10 +365,34 @@ export class Upload extends Writable {
...
+        `.${DEFAULT_UNIVERSE}`;
+
+      if (
+        !isDomain &&
+        !isDefaultUniverseDomain &&
+        !isSubDomainOfUniverse &&
+        !isSubDomainOfDefaultUniverse
+      ) {
+        // a custom, non-universe domain,
+        // use gaxios
```
danielbankhead, line 395: Going to create a follow-up issue to clean this up - we should have a cleaner, explicit mechanism for disabling/bypassing auth; even for emulator contexts.

danielbankhead: Created:
- https://github.com/googleapis/nodejs-storage/issues/2400

BrennaEpp (approved):

PR #2431, opened 2024-03-26
feat: allow setting contentEncoding during compose

### LINKED ISSUE #2430: combining gzipped files
I have few processes uploading parts of a big json-lines file and I upload them compressed, as described in [transparent encoding](https://cloud.google.com/storage/docs/transcoding). After those partial uploads I'm combining the file into the final big compressed file, and I would like to set the `metadata.contentEncoding` during the `/compose` request.

Concatenated gzip files creates valid gzip file; I verified that if I add content encoding manually to combined file then the decompressive transcoding works as intended.

I'll provide pull-request

zuozp8:
Fixes #2430

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -1658,6 +1658,7 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
         json: {
           destination: {
             contentType: destinationFile.metadata.contentType,
+            contentEncoding: destinationFile.metadata.contentEncoding,
           },
           sourceObjects: (sources as File[]).map(source => {
             const sourceObject = {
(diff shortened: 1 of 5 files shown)


ddelgrosso1:
Thanks for the contribution @zuozp8. It looks like the test 
```
Bucket
       combine
         should make correct API request:
```
needs to be updated.

zuozp8:
@ddelgrosso1 sure, I updated the test

I made sure my `undefined` doesn't look and caused content-type to be undefined too in that test

zuozp8:
@ddelgrosso1 linter wants to change files that I didn't touch, should i change them?

wouldn't it be better to commit `package-lock.json` into repo to avoid unexpected version bumps?

ddelgrosso1:
@zuozp8 don't worry about them, I have another PR open that will fix them. As a general rule, we don't normally commit package locks but sometimes this is the outcome 😆

ddelgrosso1:
Thanks again for the contribution @zuozp8. I will get this merged / released after Cloud Next is over.

ddelgrosso1 (approved):

PR #2432, opened 2024-03-26
feat: add ability to create a File object from URL

### LINKED ISSUE #1898: Proposal: Instantiate a File using Public URL

Right now [File](https://googleapis.dev/nodejs/storage/latest/Bucket.html#file) can only be instantiated using the file name. However, is common for systems to save the entire file url in databases, in these cases, to later delete the file from the bucket you would need to extract the file name (including parent directories) from the public url. Which means that you would have to spend some time thinking how to properly do this and write unit tests for it. I would prefer if the library handles this for me.

1 - Switch from `file(name: string, options?: FileOptions): File` to just `file(options: FileOptions): File`

2 - Extend [FileOptions](https://cloud.google.com/nodejs/docs/reference/storage/latest/storage/fileoptions) like:

```typescript
interface FileOptions {
    name?: string,
    publicUrl?: string,
    ...
}
```

3- Extract the name from the public url:

If `name` is not filled, we can proceed and check `publicUrl` -> From `https://storage.googleapis.com/${firebase-project}.appspot.com/images%2Ftest.jpeg` extract just `images%2Ftest.jpeg`, decode it as `images/test.jpeg` and use it as `name`. This would require doing a [split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split): We already know the host is a static `https://storage.googleapis.com` and the bucket name (`${firebase-project}.appspot.com`) we already have it from [here](https://googleapis.dev/nodejs/storage/latest/Storage.html#bucket).

**Example of usage of proposed API**

`file({ name: "images/test.jpeg" })` and `file({ publicUrl: "https://storage.googleapis.com/${firebase-project}.appspot.com/images%2Ftest.jpeg" })`

A backwards compatible way to achieve this would be to leave the API as it is but automatically recognise if `name` is a public url, in this case you would apply the same extraction logic mentioned before.

I can offer myself to open the PR if you guys agree with one solution

ddelgrosso1:

Fixes #1898  🦕

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -324,7 +324,7 @@ export const STORAGE_POST_POLICY_BASE_URL = 'https://storage.googleapis.com';
 /**
  * @private
  */
-const GS_URL_REGEXP = /^gs:\/\/([a-z0-9_.-]+)\/(.+)$/;
+const GS_URL_REGEXP = /^gs:\/\/([a-z0-9_.-]+)\/(.+)$/g;
 
 export interface FileOptions {
   crc32cGenerator?: CRC32CValidatorGenerator;
@@ -479,6 +479,9 @@ export class RequestError extends Error {
 }
 
 const SEVEN_DAYS = 7 * 24 * 60 * 60;
+const GS_URL_REGEX = /(gs):\/\/([a-z0-9_.-]+)\/(.+)/g;
+const HTTPS_PUBLIC_URL_REGEX =
+  /(https):\/\/(storage\.googleapis\.com)\/([a-z0-9_.-]+)\/(.+)/g;
 
 export enum FileExceptionMessages {
   EXPIRATION_TIME_NA = 'An expiration time is not available.',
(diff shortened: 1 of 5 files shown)


src/file.ts
```diff
@@ -2358,6 +2361,35 @@ class File extends ServiceObject<File, FileMetadata> {
...
+      return new File(bucket, gsMatches[0][2], options);
+    } else if (httpsMatches.length > 0) {
+      const bucket = new Bucket(storageInstance, httpsMatches[0][2]);
+      return new File(bucket, httpsMatches[0][3], options);
+    } else {
+      throw new Error(
+        'URL string must be of format gs://bucket/file or https://storage.googleapis.com/bucket/file'
+      );
+    }
+  }
```
danielbankhead, line 2391: Alternatively, we could parse the URL, then use `protocol` to determine the logic. Here's an example with an optional `storage` param:

```suggestion
  static from(
    publicUrlOrGsUrl: string | URL,
    storage?: Storage,
    options?: FileOptions
  ): File {
    const url = new URL(publicUrlOrGsUrl);

    if (url.protocol === 'gs:') {
      const storageInstance = storage ?? new Storage();
      const bucket = new Bucket(storageInstance, url.host);
      return new File(bucket, url.pathname.slice(1), options);
    } else if (url.protocol === 'http:' || url.protocol === 'https:') {
      const {groups: {bucketName, fileName}} = /\/(?<bucket>[^/]*)\/(?<filename>.*)/.exec(u.pathname);
      const storageInstance = storage ?? new Storage({apiEndpoint: url.hostname});
      const bucket = new Bucket(storageInstance, bucketName);
      return new File(bucket, fileName, options);
    } else {
      throw new Error(
        'URL string must be of format gs://bucket/file or https://storage.googleapis.com/bucket/file'
      );
    }
  }
```

If we do this we'll need to make a small change to `storage.ts` on line ~180 to prevent accidental custom `apiEndpoint` when provided.
From:
```ts
if (options.apiEndpoint) {
```
To:
```ts
if (options.apiEndpoint && options.apiEndpoint !== apiEndpoint) {
```

danielbankhead (approved):
Looks great! One optional suggestion

src/file.ts
ddelgrosso1: Good suggestion, I'm apt to leave this using the regexs to make it a bit more concise and avoid the apiEndpoint logic. It also matches how it was done in the python client (not that consistency is SUPER important).

shaffeeullah:
let's gooooo 🚀  so excited to see this feature get implemented!
appreciate you @ddelgrosso1 & @danielbankhead  !!

PR #2435, opened 2024-04-04
fix: remove extraneous mime-types package in favor of mime

### LINKED ISSUE #2152: refactor: remove mime-types / mime-db dependency
Currently we utilize `mime` to detect content types and set metadata accordingly. Mime uses mime-db which is essentially a lookup table from JSON files. This dependency is accounting for ~13% of our package size. We should remove this dependency in favor of something smaller for example we could write a MIME sniffer based on the [spec](https://mimesniff.spec.whatwg.org/#matching-a-mime-type-pattern) if one does not already exist for Node.

ddelgrosso1:

Fixes #2152 🦕


danielbankhead (approved):

ddelgrosso1:
Going to add the additional logic to remove `compressible`. Adding do not merge in the meantime.

ddelgrosso1:
Added additional logic and removed compressible from the dependencies.

src/file.ts
```diff
@@ -326,6 +325,24 @@ export const STORAGE_POST_POLICY_BASE_URL = 'https://storage.googleapis.com';
  */
 const GS_URL_REGEXP = /^gs:\/\/([a-z0-9_.-]+)\/(.+)$/;
 
+/**
+ * @private
```
danielbankhead, line 329: Perhaps a comment or [@privateRemarks](https://tsdoc.org/pages/tags/privateremarks/) on how to regenerate this regex over time would keep this maintainable in the future

danielbankhead (approved):
LGTM, one comment

PR #2439, opened 2024-04-17
fix: change copyoptions type

### LINKED ISSUE #2389: Apparently wrong type of `CopyOptions.metadata`
It seems that the type of `CopyOptions.metadata` is wrong.

Before https://github.com/googleapis/nodejs-storage/pull/2234 (landed in f48dcd2d00081aea8990f35b68a91248f3862abe) `CopyOptions.metadata` was of type `Metadata=any`: https://github.com/googleapis/nodejs-storage/pull/2234/files#diff-bc9705d0f7a567399044dfc66ccc82d4d9aa1cff116842a0094d54e463c9ecbcL313, https://github.com/googleapis/nodejs-storage/pull/2234/files#diff-dd08f09f3839e8051415914bfa81750cc0f28ae2d9b1d6b95eb87bb0efff2577L48. This means that it refers to the custom metadata, as `cacheControl`, `contentType` and other properties are declared directly on the `CopyOptions` interface.

After that PR was merged, `CopyOptions.metadata` got type `FileMetadata`, which itself contains `cacheControl`, `contentType` as well as a nested `metadata`, which is `Record<string, string>` and means “custom metadata”.

It seems to me that `CopyOptions.metadata` should be of type `Record<string, string>` directly. This is confirmed by code below.

#### Steps to reproduce

```js
file.copy(to,
  {
    contentType: 'outer', // A property from CopyOptions
    metadata: { // Typed as FileMetadata
      contentType: 'internal',
      custom1: 'outer', // Not a part of FileMetadata, but allowed by TypeScript
      metadata: { // Intended to hold custom metadata, but ignored
        custom2: 'internal'
      }
    },
  }
);
```

The copied file gets the following metadata:

- `Content-Type: outer`
- Custom metadata:
  - `contentType: inner` - note how a property of `FileMetadata.contentType` it became a custom metadata.
  - `custom1: outer`

Whereas `custom2` is ignored.

image

----
#### Environment details

  - OS: any
  - Node.js version: any
  - npm version: any
  - `@google-cloud/storage` version: `7.7.0`

vishwarajanand:
Fixes https://github.com/googleapis/nodejs-storage/issues/2389

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -361,7 +361,9 @@ export interface CopyOptions {
   contentType?: string;
   contentDisposition?: string;
   destinationKmsKeyName?: string;
-  metadata?: FileMetadata;
+  metadata?: {
+    [key: string]: string;
+  };
   predefinedAcl?: string;
   token?: string;
   userProject?: string;
--- a/src/file.ts
+++ b/src/file.ts
@@ -362,7 +362,7 @@ export interface CopyOptions {
   contentDisposition?: string;
   destinationKmsKeyName?: string;
   metadata?: {
-    [key: string]: string;
+    [key: string]: string | boolean | number | null;
   };
   predefinedAcl?: string;
   token?: string;
--- a/test/file.ts
+++ b/test/file.ts
@@ -545,12 +545,17 @@ describe('File', () => {
 
     it('should accept an options object', done => {
       const newFile = new File(BUCKET, 'name');
+      const METADATA = {
+        metadataKey: 'metadataValue',
+      };
       const options = {
         option: true,
+        metadata: METADATA,
       };
 
       file.request = (reqOpts: DecorateRequestOptions) => {
         assert.deepStrictEqual(reqOpts.json, options);
+        assert.strictEqual(reqOpts.json.metadata, METADATA);
         done();
       };
 


ddelgrosso1:
The change itself looks good. I think the only thing I would suggest is perhaps we add a system test and / or unit test to make sure editable metadata is set correctly during a `copy`.

ddelgrosso1 (approved):

src/file.ts
```diff
@@ -361,7 +361,9 @@ export interface CopyOptions {
   contentType?: string;
   contentDisposition?: string;
   destinationKmsKeyName?: string;
-  metadata?: FileMetadata;
+  metadata?: {
+    [key: string]: string;
```
ddelgrosso1, line 365: Apologies, realized this after I made some similar changes yesterday. We probably need to expand the types of this, similar to: https://github.com/googleapis/nodejs-storage/pull/2442/files#diff-bc9705d0f7a567399044dfc66ccc82d4d9aa1cff116842a0094d54e463c9ecbcR473

ddelgrosso1 (changes requested):
Added a comment about expanding the types of custom metadata

ddelgrosso1 (approved):

PR #2460, opened 2024-05-07
fix: Do not set `customEndpoint` if `apiEndpoint === default`

danielbankhead:
🦕


ddelgrosso1 (commented):
LGTM. Was there an associated problem with this?

ddelgrosso1 (approved):

danielbankhead:
> LGTM. Was there an associated problem with this?

Nope! I think it would be a minor, but useful feature for `universeDomain` users

PR #2466, opened 2024-05-12
Fix: improve GetFilesResponse interface

### LINKED ISSUE #2464: Missing interface's property
Why doesn't the second element of the array have the property pageToken?
https://github.com/googleapis/nodejs-storage/blob/b4dbd73189b9fae4c23d614753670ee562bc717b/src/bucket.ts#L90
Screenshot 2024-05-12 at 21 11 41

 This property will be present if you pass `{ autoPaginate: false }` into `Storage.getFiles`.

olishevskii:

Fixes https://github.com/googleapis/nodejs-storage/issues/2464 🦕

### DIFF
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -87,7 +87,11 @@ interface MetadataOptions {
   ifMetagenerationNotMatch?: number | string;
 }
 
-export type GetFilesResponse = [File[], (GetFilesOptions | {}) & Partial<Pick<GetFilesOptions, 'pageToken'>>, unknown];
+export type GetFilesResponse = [
+  File[],
+  (GetFilesOptions | {}) & Partial<Pick<GetFilesOptions, 'pageToken'>>,
+  unknown,
+];
 export interface GetFilesCallback {
   (
     err: Error | null,
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -87,7 +87,11 @@ interface MetadataOptions {
   ifMetagenerationNotMatch?: number | string;
 }
 
-export type GetFilesResponse = [File[], {}, unknown];
+interface NextPageFilesOptions {
+  pageToken?: string;
+}
+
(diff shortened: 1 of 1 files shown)


ddelgrosso1:
Hi @olishevskii thank you for opening a PR and trying to clean this up. 

If you take a look at https://github.com/googleapis/nodejs-storage/blob/b4dbd73189b9fae4c23d614753670ee562bc717b/src/bucket.ts#L2840 you will see that what is actually being returned is a cross between `GetFilesOptions & {pageToken?: string}`. This PR omits the `GetFilesOptions` portion. 

I'm also not sure we want to define a new interface that will need to be exported just to hold the page token portion.

olishevskii:
Hi @ddelgrosso1. I see your point of view regarding defining a new interface `NextPageFilesOptions`. I've rewritten my solution using TypeScript utilities. Additionally, I noticed that a query can be not only like `GetFilesOptions` but also like `GetFilesCallback`.
https://github.com/googleapis/nodejs-storage/blob/b4dbd73189b9fae4c23d614753670ee562bc717b/src/bucket.ts#L2798-L2802

src/bucket.ts
```diff
@@ -87,7 +87,7 @@ interface MetadataOptions {
   ifMetagenerationNotMatch?: number | string;
 }
 
-export type GetFilesResponse = [File[], {}, unknown];
+export type GetFilesResponse = [File[], (GetFilesOptions | GetFilesCallback) & Partial<Pick<GetFilesOptions, 'pageToken'>>, unknown];
```
ddelgrosso1, line 90: You can omit `GetFilesCallback`, `query` can only be `GetFilesOptions` or an empty object. It will never get assigned the callback.

olishevskii: I fixed it.

ddelgrosso1 (approved):

Issue #2092, opened 2022-10-21
feat: Align Support for Storage Emulation


danielbankhead:
Today, `baseUrl` is equivalent to the `STORAGE_EMULATOR_HOST`, when available. This leads to [limitations](https://github.com/googleapis/nodejs-storage/issues/1314) when using the client. Here's the relevant code snippet:

https://github.com/googleapis/nodejs-storage/blob/14472691121aa6af0a316c0bf78728edb164c66c/src/storage.ts#L650-L668

Ideally, `STORAGE_EMULATOR_HOST` should only be a host (without any path associations, e.g. `http://localhost:9199`) while the client calls the appropriate path on the host (e.g. `/storage/v1`). That implementation should look something like this: https://github.com/googleapis/nodejs-storage/commit/c75b8b82262dddb794304a71f648bd6e03cafb30

This requires coordination with Firebase and updates to [Firebase Storage Emulator](https://firebase.google.com/docs/emulator-suite/connect_storage). Additionally, we should explore practices to ensure reliability and a consistent experience between Firebase and Google Cloud Storage (e.g. code consolidation and integration testing).

Background:
- https://github.com/googleapis/nodejs-storage/issues/1314
- https://github.com/googleapis/nodejs-storage/issues/2069
- https://github.com/googleapis/nodejs-storage/pull/2070
- https://github.com/googleapis/nodejs-storage/issues/2086
- https://github.com/googleapis/nodejs-storage/pull/2087
- https://github.com/googleapis/nodejs-storage/pull/2089

mgabeler-lee-6rs:
The Go GCS client looks like it had some similar issues in the past. There's a fair bit in that repo if you [search issues for `STORAGE_EMULATOR_HOST`](https://github.com/googleapis/google-cloud-go/search?q=STORAGE_EMULATOR_HOST&type=issues), but this ticket is the closest to this and has two PRs linked to fix it: https://github.com/googleapis/google-cloud-go/issues/2476

The changes there look _functionally_ equivalent to what I proposed in #2070, so I'm wondering if the firebase emulator may already have problems with the Go client? I have no experience with the firebase emulator, so I couldn't say for sure.

A thought: as a stepping stone to avoid a "hard break", might it be useful to temporarily permit a second environment variable indicating how to interpret `STORAGE_EMULATOR_HOST`, with the initial default being the current behavior, and some specific value indicating to use the "new style", and at some point that default changes, and later that second env var is dropped?

[...]

PR #2470, opened 2024-05-16
fix: Improve STORAGE_EMULATOR_HOST usage

le0pard:
## Problem

 - if provided `STORAGE_EMULATOR_HOST` env, `apiEndpoint == baseUrl`. This lead to invalid url to fetch data, because no `storage/v1` in `baseUrl`
 - if provided `apiEndpoint`, in this case all works

## Expected behaviour

 - `STORAGE_EMULATOR_HOST` env works same way as `apiEndpoint` property

## Fix

 - Apply same logic for `STORAGE_EMULATOR_HOST` as for `apiEndpoint` to generate `baseUrl`

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -724,7 +724,7 @@ export class Storage extends Service {
     options = Object.assign({}, options, {apiEndpoint});
 
     // Note: EMULATOR_HOST is an experimental configuration variable. Use apiEndpoint instead.
-    const baseUrl = EMULATOR_HOST || `${options.apiEndpoint}/storage/v1`;
+    const baseUrl = `${EMULATOR_HOST || options.apiEndpoint}/storage/v1`;
 
     const config = {
       apiEndpoint: options.apiEndpoint!,
--- a/test/index.ts
+++ b/test/index.ts
@@ -493,7 +493,7 @@ describe('Storage', () => {
         });
 
         const calledWith = storage.calledWith_[0];
-        assert.strictEqual(calledWith.baseUrl, EMULATOR_HOST);
+        assert.strictEqual(calledWith.baseUrl, `${EMULATOR_HOST}/storage/v1`);
         assert.strictEqual(
           calledWith.apiEndpoint,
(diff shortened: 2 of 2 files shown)


ddelgrosso1:
Hi @le0pard we appreciate the contribution. However, there is a reason we do not append pathing to `STORAGE_EMULATOR_HOST`. For more details see: https://github.com/googleapis/nodejs-storage/pull/2087 and https://github.com/googleapis/nodejs-storage/issues/2092

le0pard:
Thanks @ddelgrosso1 Should we just introduce new env variable, which will work the same as `apiEndpoint`, while team busy make `STORAGE_EMULATOR_HOST` work with Firebase Storage emulator ?

ddelgrosso1:
@le0pard fair question. Let me check internally with the firebase team as to what the current status is. The problem with adding a new variable is it is one more thing to maintain and deprecating something else becomes a long task. 

Is there a problem with using apiEndpoint for your use case?

le0pard:
Thanks @ddelgrosso1 

`apiEndpoint` is exactly how I resolve this issue for now. Problem, that in project in use node.js and python code. In python  sdk `STORAGE_EMULATOR_HOST` work as expected ( https://github.com/googleapis/python-storage/blob/main/google/cloud/storage/client.py#L168-L171 ) , but not here. So in python code I dont need create any additional code for sdk to switch endpoint for local development. Just pass env variable in docker container. In JS codebase I need create separate env variable like `GCLOUD_API_ENDPOINT` and write code, which will check this env variable to make sdk work in same way, as python:

```js
const {
  Storage: GoogleStorage
} = require('@google-cloud/storage')

const STORAGE_ENDPOINT = process.env.GCLOUD_API_ENDPOINT ?? undefined

const storage = new GoogleStorage({
  apiEndpoint: STORAGE_ENDPOINT
})
```

And of course I cannot share same `STORAGE_EMULATOR_HOST` env variable, because it is "break" node.js sdk right now

PR #2489, opened 2024-06-14
fix: make CreateBucketRequest extend from BucketMetadata to allow all…

### LINKED ISSUE #2454: CreateBucketRequest type is misaligned with JSON API
Currently the [`CreateBucketRequest` type](https://github.com/googleapis/nodejs-storage/blob/c267e6b81e8d0d19f6c7ef886c6897e4bfc7897b/src/storage.ts#L119) is missing some fields [that are supported by the JSON API ](https://cloud.google.com/storage/docs/json_api/v1/buckets/insert). 

Short term: add the missing fields
Long term: Use a single metadata / resource representation type

ddelgrosso1:
… fields to be set

Fixes #2454  🦕

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -48,9 +48,23 @@ export interface GetServiceAccountCallback {
 }
 
 export interface CreateBucketQuery {
-  project: string;
-  userProject: string;
   enableObjectRetention: boolean;
+  predefinedAcl?:
+    | 'authenticatedRead'
+    | 'private'
+    | 'projectPrivate'
+    | 'publicRead'
+    | 'publicReadWrite';
+  predefinedDefaultObjectAcl?:
+    | 'authenticatedRead'
+    | 'bucketOwnerFullControl'
+    | 'bucketOwnerRead'
+    | 'private'
+    | 'projectPrivate'
+    | 'publicRead';
+  project: string;
+  projection?: 'full' | 'noAcl';
+  userProject: string;
 }
 
 export enum IdempotencyStrategy {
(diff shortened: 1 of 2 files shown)


ddelgrosso1:
Making `CreateBucketReqest` extend `BucketMetadata` so that the missing fields that are accepted by https://cloud.google.com/storage/docs/json_api/v1/buckets/insert get carried over. I've left the custom fields alone as that would be a breaking change.

src/storage.ts
```diff
@@ -117,30 +131,30 @@ export interface AutoclassConfig {
...
+    | 'authenticatedRead'
+    | 'bucketOwnerFullControl'
+    | 'bucketOwnerRead'
+    | 'private'
+    | 'projectPrivate'
+    | 'publicRead';
+  projection?: 'full' | 'noAcl';
   regional?: boolean;
   requesterPays?: boolean;
-  retentionPolicy?: object;
```
danielduhh, line 143: will removing this cause any issues?

danielduhh (commented):
tests?

src/storage.ts
ddelgrosso1: It isn't removed. This interface now extends `BucketMetadata` https://github.com/googleapis/nodejs-storage/blob/18eef67892af6c7311a12d4c22579e240149eb02/src/bucket.ts#L347

ddelgrosso1:
@danielduhh all tests are existing. Didn't make much sense to add unit tests to test each individual field is settable.

danielduhh (approved):

PR #2495, opened 2024-07-05
fix: Handle unhandled error in `startResumableUpload_`

### LINKED ISSUE #2494: Unhandled Error in startResumableUpload_ Function
Cannot handle the error by `try...catch` nor `writable.on('error')`

#### Environment details

  - OS: All
  - Node.js version: All
  - npm version: All
  - `@google-cloud/storage` version: 7.11.2

#### Steps to reproduce

  1.  partial upload without url provided
  2. run `file.createWriteStream`
  

```javascript
const opitons = {
  offset: 1,
  isPartialUpload: true,
};
let writable;
try {
   writable = file.createWriteStream(opitons);
} catch (error) {
  // cannot catch the error
  return;
}

writable.on('error', (err: RangeError) => {
   // cannot catch the error
});

writable.write('data');
```

cuyl:
* Added error handling using `stream.destroy`

Fixes #2494

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -4141,8 +4141,7 @@ class File extends ServiceObject<File, FileMetadata> {
     ) {
       retryOptions.autoRetry = false;
     }
-
-    const uploadStream = resumableUpload.upload({
+    const cfg = {
       authClient: this.storage.authClient,
       apiEndpoint: this.storage.apiEndpoint,
       bucket: this.bucket.name,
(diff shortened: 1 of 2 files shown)


src/file.ts
```diff
@@ -4168,7 +4167,17 @@ class File extends ServiceObject<File, FileMetadata> {
...
-    });
+    };
+
+    let uploadStream: resumableUpload.Upload;
+
+    try {
+      uploadStream = resumableUpload.upload(cfg);
+    } catch (error) {
+      dup.destroy(error as Error);
+      this.storage.retryOptions.autoRetry = false;
```
ddelgrosso1, line 4178: There is no need to set this here.

cuyl: Seems this is a unrecoverable error.

https://github.com/googleapis/nodejs-storage/blob/18eef67892af6c7311a12d4c22579e240149eb02/src/resumable-upload.ts#L350-L354

ddelgrosso1: Retries for resumable uploads do not happen at this layer. Turning this to false here is shutting off retries for other operations as well.

cuyl: OK, NP

ddelgrosso1:
Thank you for the contribution @cuyl, we appreciate it.

ddelgrosso1 (approved):

PR #2519, opened 2024-09-16
feat(storage): add support for 'fields' query parameter to getFiles

### LINKED ISSUE #2138: feature: add support for 'fields' query parameter to getFiles
Currently `getFiles` retrieves all information / metadata about a File object. In some situations, for example deleting files, much of this information is unnecessary and can eat up memory if the list of files is large. Since `getFiles` is a precursor call for `deleteFiles` we should support the [fields](https://cloud.google.com/storage/docs/json_api/v1/parameters#query) query string parameter so that the information returned can be reduced.

thiyaguk09:

Fixes #2138


samples/listFilesWithFields.js
ddelgrosso1, line 1: This sample isn't necessary and can be removed.

thiyaguk09: I agree that the sample isn't necessary in this context. It can be removed to streamline the content and improve clarity.

PR #2521, opened 2024-09-17
feat(storage): add support for 'fields' query parameter to getFiles

### LINKED ISSUE #2138: feature: add support for 'fields' query parameter to getFiles

thiyaguk09:

Fixes #2138


system-test/common.ts
thiyaguk09, line 1: We adjusted the test placement to address the previous failures.

package.json
```diff
@@ -122,6 +122,8 @@
     "pack-n-play": "^2.0.0",
     "proxyquire": "^2.1.3",
     "sinon": "^18.0.0",
+    "nise": "6.0.0",
+    "path-to-regexp": "6.2.2",
```
ddelgrosso1, line 126: Let's just add a tacking issue to remove this once we upgrade to Node 18 minimum so we don't forget.

ddelgrosso1 (approved):
LGTM

PR #2538, opened 2024-09-30
feat: adds integration tests for Universe Domain configuration

thiyaguk09:

Fixes


ddelgrosso1:
Code looks good, think the `after` hook needs some adjustment as it is failing.

thiyaguk09:
> Code looks good, think the `after` hook needs some adjustment as it is failing.

The `after` hook has been adjusted. Please review the changes.

system-test/storage.ts
```diff
@@ -3866,6 +3866,63 @@ describe('storage', function () {
...
+      const [bucket] = await universeDomainStorage.createBucket(bucketName, {
+        location: TEST_UNIVERSE_LOCATION,
+      });
+
+      file = bucket.file('LogoToSign.jpg');
+      fs.createReadStream(FILES.logo.path).pipe(file.createWriteStream());
+    });
+
+    after(async () => {
+      await deleteFileAsync(file);
```
ddelgrosso1, line 3896: Since we are creating a bucket in the `before` hook. We will want to clean it up as well (or at least attempt to). You can leverage the `deleteBucketAsync` function in this file. It may need to be moved but that should be fine. If the function fails, likely safe to ignore the error so the test doesn't fail.

thiyaguk09: I've implemented changes to the after hook to delete the bucket using the deleteBucketAsync function. Please review the updated code and let me know if the issue is resolved.

ddelgrosso1 (approved):

ddelgrosso1 (changes requested):
What happened to the kokoro changes? I saw them at one point.

system-test/storage.ts
```diff
@@ -3866,6 +3866,64 @@ describe('storage', function () {
     });
   });
 
+  describe('universeDomainTests', () => {
+    const TEST_UNIVERSE_DOMAIN = process.env.TEST_UNIVERSE_DOMAIN;
```
ddelgrosso1, line 3870: @JesseLovelace brought up a good point, we should probably assert that these credentials and environment variables are not undefined / null.

thiyaguk09:
> What happened to the kokoro changes? I saw them at one point.

gcf-owl-bot removed those changes https://github.com/googleapis/nodejs-storage/pull/2538/commits/31ca74401772ad48f1d119850328fd78451623d8.

system-test/storage.ts
thiyaguk09: Thanks for bringing that up, I've implemented validations to ensure that the credentials and environment variables are not undefined or null.  Added kokoro changes again. 

Kindly review and rerun kokoro.

.kokoro/samples-test.sh
```diff
@@ -22,6 +22,12 @@ export NPM_CONFIG_PREFIX=${HOME}/.npm-global
 export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/secret_manager/long-door-651-kokoro-system-test-service-account
 export GCLOUD_PROJECT=long-door-651
 
+# For universe domain testing
+export TEST_UNIVERSE_DOMAIN_CREDENTIAL=$(realpath ${KOKORO_GFILE_DIR}/secret_manager/client-library-test-universe-domain-credential)
+export TEST_UNIVERSE_DOMAIN=$(gcloud secrets versions access latest --project cloud-devrel-kokoro-resources --secret=client-library-test-universe-domain)
```
ddelgrosso1, line 27: Are we sure this is the right secret name? Looks like the test is failing because this envvar isn't set.

thiyaguk09: Due to the previous test failure, I've identified and resolved the issue by moving the code to the appropriate file. And now tests are passing successfully.

ddelgrosso1 (approved):

PR #2548, opened 2024-10-25
feat: add support for restore token

thiyaguk09:

Fixes #restoreToken


src/file.ts
```diff
@@ -547,6 +551,7 @@ class File extends ServiceObject<File, FileMetadata> {
   name: string;
 
   generation?: number;
+  restoreToken?: string;
```
ddelgrosso1, line 554: Is there a reason to have this field as both a class level and metadata level field?

ddelgrosso1:
My understanding is that the restore token would be returned on a call to delete. Is that going to be a separate PR or should it be included here?

thiyaguk09:
> My understanding is that the restore token would be returned on a call to delete. Is that going to be a separate PR or should it be included here?

I'm currently focused on implementing support for restore tokens in the get, list, and restore operations.

src/file.ts
thiyaguk09: I've noticed that in some scenarios, a refresh token is associated with a specific generation ID. To ensure consistent behavior and avoid potential issues, I've implemented the refresh token handling in a way that aligns with the generation ID concept. This approach will help maintain data integrity and security.

ddelgrosso1 (approved):

PR #2570, opened 2025-01-14
fix(getFiles): add nextPageToken to fields for autoPaginate

jjaareet:
`bucket.getFiles()` - returns all files as expected
`bucket.getFiles({fields: 'items(name)'})` - unexpectedly returns only 1,000 files

i noticed that `autoPaginate` and `fields` do not play nicely together unless `nextPageToken` is added to `fields`. hoping to prevent someone else from banging their head on that wall with this change

edit: sorry for the force push after requesting review, just noticed a typo in the commit message


ddelgrosso1:
@jjaareet thank you for the contribution!

PR #2586, opened 2025-03-20
feat: add moveFileAtomic method

thiyaguk09:


src/file.ts
```diff
@@ -282,6 +282,14 @@ export interface MakeFilePublicCallback {
   (err?: Error | null, apiResponse?: unknown): void;
 }
 
+interface MoveQuery {
```
ddelgrosso1, line 285: To avoid confusion with the existing `move` function, I would rename this MoveFileQuery.

src/file.ts
```diff
@@ -3426,6 +3434,188 @@ class File extends ServiceObject<File, FileMetadata> {
...
   }
 
+  moveObj(
+    destination: string | File,
+    options?: MoveOptions
+  ): Promise<MoveResponse>;
+  moveObj(destination: string | File, callback: MoveCallback): void;
+  moveObj(
+    destination: string | File,
+    options: MoveOptions,
```
ddelgrosso1, line 3444: I would probably create additional aliases for `MoveOptions` and `MoveCallback` called `MoveFileOptions` and `MoveFileCallback`. We want to make this very explicit that it is not the same as the existing `move`.

src/file.ts
```diff
@@ -3426,6 +3434,188 @@ class File extends ServiceObject<File, FileMetadata> {
...
+   *   const newFile = data[0];
+   *   const apiResponse = data[1];
+   * });
+   *
+   * ```
+   * @example <caption>include:samples/files.js</caption>
+   * region_tag:storage_move_file_hns
+   * Another example:
+   */
+  moveObj(
```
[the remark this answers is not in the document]
thiyaguk09: I understand the importance of appropriate naming for moveFile. However, we already have a sample named [moveFile.js](https://github.com/googleapis/nodejs-storage/blob/c338e0c4c56496065404f3834307f2935029f0b8/samples/moveFile.js) for the [move](https://github.com/googleapis/nodejs-storage/blob/c338e0c4c56496065404f3834307f2935029f0b8/src/file.ts#L3759) function, which may confuse the consumer. 

Therefore, we need to update the samples accordingly. Could I make a copy of moveFiles.js and rename it to move.js, and use moveFile.js for the new function called moveFile? Do you have any suggestions?

samples/moveFileAtomic.js
```diff
@@ -0,0 +1,76 @@
...
+ * at https://cloud.google.com/storage/docs.
+ */
+
+function main(
+  bucketName = 'my-bucket',
+  srcFileName = 'test.txt',
+  destFileName = 'test2.txt',
+  destinationGenerationMatchPrecondition = 0
+) {
+  // [START storage_move_file_atomic]
```
ddelgrosso1, line 29: This sample tag is likely incorrect. I would imagine the correct is `storage_move_object` but please double check.

thiyaguk09: Thanks for the suggestion! I'll investigate the sample tag and see if storage_move_object is indeed the correct one. This is a good learning point for me regarding naming conventions.

ddelgrosso1 (commented):
Please update the title and commit message

src/file.ts
```diff
@@ -297,6 +305,10 @@ export interface MoveOptions {
   preconditionOpts?: PreconditionOptions;
 }
 
+export type MoveFileAtomicOptions = MoveOptions;
```
ddelgrosso1, line 308: Please make sure we are exporting these from index.ts

src/file.ts
thiyaguk09: It has been renamed to MoveFileAtomicQuery.

src/file.ts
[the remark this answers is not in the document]
thiyaguk09: moveFileAtomicQuery was used for a new function.

src/file.ts
thiyaguk09: I missed adding those in index.ts. Thank you for identifying it; I have added them now.

src/file.ts
thiyaguk09: As discussed, the new function name has been changed to moveFileAtomic. The corresponding options, callback, and response have been updated to MoveFileAtomicOptions, MoveFileAtomicCallback, and MoveFileAtomicResponse, respectively.

ddelgrosso1 (approved):

PR #2592, opened 2025-04-21
chore!: migrate to Node 18

thiyaguk09:


src/bucket.ts
```diff
@@ -543,6 +542,11 @@ export enum BucketExceptionMessages {
   INVAILD_CHANNEL_RESPONSE = 'Response data was null',
 }
 
+// const createLimit = async (concurrency: number) => {
```
ddelgrosso1, line 545: Let's remove this instead of commenting it out.

src/bucket.ts
```diff
@@ -4547,8 +4553,8 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
     this.getFiles(options)
       .then(([files]) => {
         const limit = pLimit(MAX_PARALLEL_LIMIT);
-        const promises = files.map(file => {
-          return limit(() => processFile(file));
+        const promises = files.map(async file => {
+          return (await limit)(() => processFile(file));
```
ddelgrosso1, line 4557: I don't think we want to await here. Promises is supposed to be an array containing the executing promises.

src/file.ts
```diff
@@ -544,6 +555,16 @@ export enum FileExceptionMessages {
   MISSING_RESUME_CRC32C_FINAL_UPLOAD = 'The CRC32C is missing for the final portion of a resumed upload, which is required for validation. Please provide `resumeCRC32C` if validation is required, or disable `validation`.',
 }
 
+const getFileType = async (filename: string) => {
```
ddelgrosso1, line 558: From the issue and the above code, it sounded like there was a problem with dynamic imports. Should this have been reverted removed as well?

src/file.ts
```diff
@@ -1335,13 +1356,19 @@ class File extends ServiceObject<File, FileMetadata> {
 
     newFile = newFile! || destBucket.file(destName);
 
-    const headers: {[index: string]: string | undefined} = {};
+    // const headers: {[index: string]: string | undefined} = {};
```
ddelgrosso1, line 1359: Please remove if no longer needed.

src/file.ts
```diff
@@ -2009,10 +2037,11 @@ class File extends ServiceObject<File, FileMetadata> {
       !options!.metadata!.contentType ||
       options!.metadata!.contentType === 'auto'
     ) {
-      const detectedContentType = mime.getType(this.name);
-      if (detectedContentType) {
-        options!.metadata!.contentType = detectedContentType;
-      }
+      void getFileType(this.name).then(detectedContentType => {
```
ddelgrosso1, line 2040: I have some concerns about this, what will happen if getFileType takes time to resolve and the code below continues to execute?

src/file.ts
```diff
@@ -3432,6 +3463,193 @@ class File extends ServiceObject<File, FileMetadata> {
     }/${encodeURIComponent(this.name)}`;
   }
 
+  moveFileAtomic(
```
ddelgrosso1, line 3466: Can you please rebase the `node-18` branch with main so that these existing changes drop out of this PR.

src/transfer-manager.ts
```diff
@@ -256,21 +265,26 @@ class XMLMultiPartUploadHelper implements MultiPartUploadHelper {
...
+        const res = await this.authClient.request<
+          string | MultiPartUploadErrorResponse
+        >({
+          headers: this.#setGoogApiClientHeaders(headersObject),
           method: 'POST',
           url,
         });
 
-        if (res.data && res.data.error) {
+        if (typeof res.data === 'object' && res.data.error) {
```
ddelgrosso1, line 281: If we are going to change this check, please just use optional chaining. `res?.data?.error`

src/transfer-manager.ts
```diff
@@ -130,6 +130,10 @@ export interface UploadFileInChunksOptions {
   headers?: {[key: string]: string};
 }
 
+interface MultiPartUploadErrorResponse {
```
ddelgrosso1, line 133: Why are we adding new types to this file?

src/transfer-manager.ts
```diff
@@ -508,7 +527,7 @@ export class TransferManager {
       }
 
       promises.push(
-        limit(() =>
+        (await limit)(() =>
```
ddelgrosso1, line 530: I think this is changing the intention here. Promises should be an array of executing promises. This looks like it will push the result of each promise into the array.

src/transfer-manager.ts
```diff
@@ -616,7 +635,7 @@ export class TransferManager {
       }
 
       promises.push(
-        limit(async () => {
+        (await limit)(async () => {
```
ddelgrosso1, line 638: Same comment.

src/bucket.ts
thiyaguk09: This was missed during the merge but has been removed now.

src/file.ts
thiyaguk09: This was missed during the merge but has been removed now.

src/file.ts
thiyaguk09: This was missed during the merge but has been removed now.

src/bucket.ts
thiyaguk09: I agree, you are correct. During the dynamic import, the async/await was initially applied for debugging purposes, but unfortunately, it was forgotten and not removed. Thank you for finding it; it has now been removed.

src/transfer-manager.ts
thiyaguk09: I agree, you are correct. During the dynamic import, the async/await was initially applied for debugging purposes, but unfortunately, it was forgotten and not removed. Thank you for finding it; it has now been removed.

src/file.ts
thiyaguk09: Understood, and the rebase has been completed.

src/transfer-manager.ts
thiyaguk09: Updated to `(res?.data as MultiPartUploadErrorResponse)?.error`

src/transfer-manager.ts
thiyaguk09: To fix the "Property 'error' does not exist on type '{}'" issue, TypeScript defaults to {} without a specific type, which means no properties are guaranteed. Use MultiPartUploadErrorResponse to tell TypeScript that res.data may have an error property.

src/bucket.ts
```diff
@@ -4547,7 +4548,7 @@ class Bucket extends ServiceObject<Bucket, BucketMetadata> {
     this.getFiles(options)
       .then(([files]) => {
         const limit = pLimit(MAX_PARALLEL_LIMIT);
-        const promises = files.map(file => {
+        const promises = files.map(async file => {
```
ddelgrosso1, line 4551: The async can be removed here.

src/storage-transport.ts
```diff
@@ -124,29 +123,51 @@ export class StorageTransport {
...
+          projectId = _projectId;
+          this.projectId = projectId;
+        }
+        return projectId;
+      } catch (e) {
+        if (callback) return callback(e as GaxiosError);
+        throw e;
+      }
+    };
+    const requestPromise = prepareRequest().then(() => {
```
ddelgrosso1, line 156: Please use `await` syntax as opposed to `.then` particularly for new code.

test/storage-transport.ts
```diff
@@ -110,11 +113,13 @@ describe('Storage Transport', () => {
...
     assert.ok(
-      calledWith.headers['x-goog-api-client'].includes('gccl-gcs-cmd/test-key'),
+      calledWith.headers
+        .get('x-goog-api-client')
+        .includes('gccl-gcs-cmd/test-key'),
     );
   });
 
-  it('should clear and add interceptors if provided', async () => {
+  it.skip('should clear and add interceptors if provided', async () => {
```
ddelgrosso1, line 122: Was it intentional to skip this test? Was something wrong?

thiyaguk09: Gaxios interceptors are not working properly after the latest upgrade. The fix is part of the conformance test update, which is why this test was skipped here.

src/storage-transport.ts
thiyaguk09: The code has been revised with async/await for improved handling.

ddelgrosso1 (approved):

organize:
What's the schedule on this / anything that can be helped with? I'm facing various issues in the current `gaxios` / `node-fetch` versions that have been fixed later.

ddelgrosso1 (approved):

ddelgrosso1:
> Forked this for testing, I'll update this comment in case I find anything.
> 
> * Is the change in `file.delete` intended? The request contains a double forward-slash: `DELETE /storage/v1/b/$some_bucket//o/$some_object`
> * a similar change can be seen with `file.setMetadata`
> * `file.delete` seems to hang as the callback is never invoked (see `service-object.ts#L304`)

cc: @thiyaguk09

thiyaguk09:
> Forked this for testing, I'll update this comment in case I find anything.
> 
> * Is the change in `file.delete` intended? The request contains a double forward-slash: `DELETE /storage/v1/b/$some_bucket//o/$some_object`
> * a similar change can be seen with `file.setMetadata`
> * `file.delete` seems to hang as the callback is never invoked (see `service-object.ts#L304`)

Thanks for forking and testing! I appreciate you digging into that.

Regarding the double forward-slash in `DELETE /storage/v1/b/$some_bucket//o/$some_object` for `file.delete` and `file.setMetadata`, that's definitely not intended and sounds like a bug. The hanging callback on `file.delete` is also a critical issue.

I'm currently tied up with some urgent work, but I'll make sure to prioritize looking into these issues as soon as I can. Your detailed findings are really helpful!

thiyaguk09:
> can you address the comments from @organize. Also, do you know why the docs CI check is 429'ing on a bunch of samples now?

Sure, I’ll address the comments from @organize. Regarding the docs CI check returning 429 errors on several samples, I’ll investigate the cause—could be rate limiting or a recent config change. Will update shortly.

ddelgrosso1 (approved):

PR #2599, opened 2025-06-08
fix: resolving location header on resumable upload

orgads:
axios headers are no longer a plain object.

The fix is straightforward, and opening an issue requires too much effort.

- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)


orgads:
Sorry, this is only needed when I use:
```json
  "overrides": {
    "google-auth-library": "$google-auth-library"
  }
```

I guess the google-auth-library versions are incompatible. Closing for now.

Issue #2367, opened 2023-11-21
Application hangs when destroying a `File.createWriteStream()` stream to abort the upload


vbshnsk:
#### Environment details

  - OS: Mac OS Ventura 13.5
  - Node.js version: 16.19
  - npm version: 8.19
  - `@google-cloud/storage` version: 7.6.0

#### Steps to reproduce

So I am trying to implement an endpoint on my server that would pass the incoming data to a File write stream and save it to GCS, and I want to halt the upload if during streaming file turns out to be bigger than expected. Here's a snippet of how it is implemented:

```typescript
const MAX_TRANSACTION_IMAGE_SIZE_BYTES = 10 * 1024 * 1024;

async function test(body: NodeJS.ReadableStream) {
  let receivedBytes = 0;
  const cancelStream = new AbortController();

  const bucketName = 'bucket';
  const file = this.storage
    .bucket(bucketName)
    .file('image.png')
    .createWriteStream();

  const validateStream = new PassThrough();
  validateStream.on('data', (chunk) => {
    receivedBytes += chunk.length;
    if (receivedBytes > MAX_TRANSACTION_IMAGE_SIZE_BYTES) {
      // this correctly aborts everything in a pipeline, and destroys the streams, but
      // the underlying streams still seem to hang
      cancelStream.abort(new PayloadTooLargeException('Image too large'));
    }
  });

  await pipeline(body, validateStream, file, {
    signal: cancelStream.signal,
  });
}
```

I have checked the code and I see that calling `.destroy()` on the returned write stream does not really result in anything being passed to the underlying `emitStream` or `writeFileStream`, so they stay hanging. Is this a bug on the SDK side? Is there a way to abort the upload request?

vbshnsk:
I have since tried to `DELETE` the resumable upload by `crc32` when I destroy my write/validate streams but it also doesn't affect the underlying streams, so the handle stays open.

xXACCEXx:
Not sure if this is the same behaviour I'm seeing or not.

I'm seeing streams aren't closing when using WebStream apis in Node. 
To me, it feels like the same problem, the handler on the web stream side completes correctly, but the event is not emitted on the GCS lib side.

This has the effect of any means of closing the stream does not work... Either destroy or end... Neither actually closes the handle.

xXACCEXx:
Here's a little example of what I mean.

I tested this on node v20.10.0 using ts-node.

```ts
import { join } from 'path';
[...]

PR #2624, opened 2025-08-06
fix: propagate errors when using pipelines (#2560)

### LINKED ISSUE #2560: Uncatchable error when upload stream is used in failed pipeline.
### Please make sure you have searched for information in the following guides.

- [X] Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-node/issues
- [X] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
- [X] Check our Troubleshooting guide: https://github.com/googleapis/google-cloud-node/blob/main/docs/troubleshooting.md
- [X] Check our FAQ: https://github.com/googleapis/google-cloud-node/blob/main/docs/faq.md
- [X] Check our libraries HOW-TO: https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md
- [X] Check out our authentication guide: https://github.com/googleapis/google-auth-library-nodejs
- [X] Check out handwritten samples for many of our APIs: https://github.com/GoogleCloudPlatform/nodejs-docs-samples

### A screenshot that you have tested with "Try this API".

N/A. This issue seems to be related to the client library's error handling ability and not the underlying API.

### Link to the code that reproduces this issue. A link to a **public** Github Repository or gist with a minimal reproduction.

https://gist.github.com/thebenperson/c6c1e66653fdc8c05cab52726b2f0664

### A step-by-step description of how to reproduce the issue, based on the linked reproduction.

1. Create a readable input stream.
2. Create a writable output stream by calling File.createWriteStream().
3. Call Stream.promises.pipeline(input, output).
4. Have the input stream emit() an error before it can provide any data.

### A clear and concise description of what the bug is, and what you expected to happen.

Expected behavior: Stream.promises.pipeline() promise is rejected.
Actual behavior: Program ends due to an unhandled "error" event.

I might have found the problem.

The error from the input stream is passed on to the output stream.
But, no data has been provided from the input stream yet.

This means that this callback hasn't been called yet:
https://github.com/googleapis/nodejs-storage/blob/740d30d09e8a2ff5a85c7f12831b03284d483fa6/src/file.ts#L2121

Meaning pipeline() hasn't been called here yet:
https://github.com/googleapis/nodejs-storage/blob/740d30d09e8a2ff5a85c7f12831b03284d483fa6/src/file.ts#L2128

So that means that there is nothing to handle errors when 
emitStream.destroy(e) is called here:
https://github.com/googleapis/nodejs-storage/blob/740d30d09e8a2ff5a85c7f12831b03284d483fa6/src/file.ts#L2075

### A clear and concise description WHY you expect this behavior, i.e., was it a recent change, there is documentation that points to this behavior, etc. **

Other types of writable streams to not cause errors like this one.
The writable stream from this library is interfering with the pipeline() function's ability to handle errors.

wvanderdeijl:

This Pull Request add two tests relating to issues #2367 and #2560.

The first one is about the application hanging when a pipeline is used that writes to a File stream and that pipeline errors. This was fixed in PR #2437 but without any unit tests. That change introduced a regression that is logged as #2560.
Since both issues are related I tried to add unit tests for both scenarios.

If you undo the changes form PR #2437 the "should close upstream when pipeline fails" test fails as we reintroduce the bug. But the "should error pipeline if source stream emits error before any data" test succeeds as this problem did not exist before v7.10.0.

If you run the tests with the latest version (so with #2437 changes in place) the "should close upstream when pipeline fails" test now succeeds, but the "should error pipeline if source stream emits error before any data" fails with an "Uncaught Error: Error before first chunk".

Here is what Gemini thinks of this:

> ## The Cause of the "Uncaught Error"
> The test fails with an "Uncaught Error" because an internal stream within your createWriteStream method emits an 'error' event that has no listener attached to it. In Node.js, an unhandled 'error' event on an EventEmitter (which streams are) will crash the process.
>
> Here's a step-by-step breakdown of what's happening in your test:
>
> 1. Your test creates a writable stream by calling file.createWriteStream().
> 2. It then uses stream.pipeline() to pipe a source stream that is designed to fail immediately (before producing any data) into this writable stream.
> 3. The pipeline utility correctly detects the error from the source stream and, as part of its cleanup, calls writable.destroy(error).
> 4. This causes the writable stream to emit an 'error' event. Your test's pipeline callback correctly catches this, and the test should pass.
> 5. Here's the problem: Inside createWriteStream, there's a listener on the writable stream's 'error' event: writeStream.once('error', e => { emitStream.destroy(e); });. This is intended to propagate the error to an internal emitStream.
> 6. However, the error handling for emitStream is set up inside a writeStream.on('writing', ...) listener. This 'writing' event is only emitted when data starts to flow.
> 7. In your test's scenario, the source stream fails before any data is sent. The 'writing' event never fires, so the error handling for emitStream (which would be the inner pipeline) is never attached.
> 8. When emitStream.destroy(error) is called, emitStream emits an 'error' event. With no listeners attached, this becomes an uncaught exception, and Mocha reports it as a test failure.

The code changes were also suggested by Gemini.

Fixes #2560

### DIFF
--- a/src/file.ts
+++ b/src/file.ts
@@ -2100,6 +2100,10 @@ class File extends ServiceObject<File, FileMetadata> {
 
     const emitStream = new PassThroughShim();
 
+    // If `writeStream` is destroyed before the `writing` event, `emitStream` will not have any listeners. This prevents an unhandled error.
+    const noop = () => {};
+    emitStream.on('error', noop);
+
     let hashCalculatingStream: HashStreamValidator | null = null;
 
     if (crc32c || md5) {
@@ -2137,6 +2141,9 @@ class File extends ServiceObject<File, FileMetadata> {
       } else {
         this.startResumableUpload_(fileWriteStream, options);
       }
+
+      // remove temporary noop listener as we now create a pipeline that handles the errors
+      emitStream.removeListener('error', noop);
 
       pipeline(
         emitStream,
--- a/test/file.ts
+++ b/test/file.ts
@@ -2303,8 +2303,11 @@ describe('File', () => {
         }
       });
 
-      file.startResumableUpload_ = (dup: duplexify.Duplexify) =>
+      file.startResumableUpload_ = (dup: duplexify.Duplexify) => {
         dup.setWritable(uploadStream);
+        uploadStream.emit('error', error);
+      };
+
       let closed = false;
       uploadStream.on('close', () => {
         closed = true;
--- a/test/file.ts
+++ b/test/file.ts
@@ -23,7 +23,14 @@ import {
 } from '../src/nodejs-common/index.js';
 import {describe, it, before, beforeEach, afterEach} from 'mocha';
 import {PromisifyAllOptions} from '@google-cloud/promisify';
-import {Readable, PassThrough, Stream, Duplex, Transform} from 'stream';
+import {
+  Readable,
+  PassThrough,
+  Stream,
+  Duplex,
+  Transform,
+  pipeline,
+} from 'stream';
 import assert from 'assert';
 import * as crypto from 'crypto';
 import duplexify from 'duplexify';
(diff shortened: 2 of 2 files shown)


wvanderdeijl:
I don't understand why the new test fails in CI for node 14. I only have a apple sillicon mac and nodejs 14 is not available for that architecture, so I can't test locally.

ddelgrosso1 (approved):

PR #2659, opened 2025-10-07
fix: Transform network failures into specific TLS timeout

thiyaguk09:
## Description

> This change refines error handling in `util.makeRequest` to intercept and transform common low-level network failures (`ECONNRESET`, `ETIMEDOUT`, `"timed out"`, `"TLS handshake"`) into **three distinct and specific standard `Error` messages**.
>
>The following raw errors are now intercepted:
>* `ECONNRESET`
>* `ETIMEDOUT`
>* Generic messages containing `"timed out"`
>* Generic messages containing `"TLS handshake"`
>
>Instead of a single message, developers now receive tailored diagnostic information: a specific message for **TLS/CPU starvation issues**, one for **network timeouts**, and one for **connection resets**. This provides clearer guidance for debugging. The original stack trace is preserved on the new `Error` object. This is an additive change with no breaking impacts, validated by a consolidated, data-driven unit test suite.
>
>By transforming these errors and preserving the original stack trace, we prevent the propagation of cryptic, low-level network codes and provide developers with a clear, unified diagnostic message tailored to the type of connection failure experienced.
>

## Impact

>The impact of this change is primarily positive, improving the developer experience:
>
>* **Improved Error Diagnostics (Granular):** Developers receive **three distinct, specific messages** instead of a single ambiguous one. For instance, a `TLS handshake` error receives a special message related to **CPU starvation** to guide performance debugging.
>* **Consistent Error Handling:** Facilitates easier integration with custom error retry and logging mechanisms by providing a predictable, standard `Error` structure (augmented with the original stack trace) rather than a raw, non-standard network error object.
>* **No Breaking Changes:** This is a purely additive fix that catches and transforms errors that would have been thrown anyway. It does not alter the successful path for requests.
>

## Testing

>**Yes, unit tests were added.**
>
>* A dedicated unit test suite, **`Network Connectivity Errors`**, was created under `makeAuthenticatedRequestFactory` to validate the new transformation logic.
>* The test structure was consolidated into a single, **data-driven `forEach` loop** that verifies the mapping of the four common failure modes to the **three distinct output messages** defined in the `UtilExceptionMessages` enum.
>
>* A single test loop replaced four separate tests, covering all conditions:
>    * `should transform raw ECONNRESET into specific network error`
>    * `should transform raw "TLS handshake" into specific network error`
>    * `should transform raw generic "timed out" into specific network error`
>    * `should transform raw ETIMEDOUT into specific network error`
>
>**Tests Changed?** No existing tests were modified.
>
>**Breaking Changes?** No breaking changes are necessary.
>

## Checklist

Fixes #

### DIFF
--- a/src/nodejs-common/util.ts
+++ b/src/nodejs-common/util.ts
@@ -256,6 +256,12 @@ export interface ParsedHttpResponseBody {
   err?: Error;
 }
 
+export enum UtilExceptionMessages {
+  TLS_TIMEOUT_ERROR_MESSAGE = 'Request or TLS handshake timed out. This may be due to CPU starvation or a temporary network issue.',
+  ETIMEDOUT_ERROR_MESSAGE = 'Connection timed out. This suggests a network issue or the server took too long to respond.',
+  ECONNRESET_ERROR_MESSAGE = 'Connection reset by peer. The connection was closed unexpectedly, possibly by a firewall or network issue.',
+}
+
 /**
  * Custom error type for API errors.
  *
@@ -914,15 +920,29 @@ export class Util {
         (err: Error | null, response: {}, body: any) => {
           if (err) {
             const lowerCaseMessage = err.message.toLowerCase();
-            const isTLsTimeoutOrConnReset =
-              /tls handshake|timed out|etimedout|econnreset/.test(
-                lowerCaseMessage
+            let newError: Error | undefined;
+
+            if (lowerCaseMessage.includes('tls handshake')) {
+              newError = new Error(
+                UtilExceptionMessages.TLS_TIMEOUT_ERROR_MESSAGE
               );
-            if (isTLsTimeoutOrConnReset) {
-              const TLS_TIMEOUT_ERROR_MESSAGE =
-                'Request or TLS handshake timed out. This may be due to CPU starvation or a temporary network issue.';
-              const timeOutError = new Error(TLS_TIMEOUT_ERROR_MESSAGE);
-              err = timeOutError;
+            } else if (
+              lowerCaseMessage.includes('etimedout') ||
+              lowerCaseMessage.includes('timed out')
+            ) {
+              newError = new Error(
+                UtilExceptionMessages.ETIMEDOUT_ERROR_MESSAGE
+              );
+            } else if (lowerCaseMessage.includes('econnreset')) {
+              newError = new Error(
+                UtilExceptionMessages.ECONNRESET_ERROR_MESSAGE
+              );
+            }
+
+            if (newError) {
+              // Preserve the original stack trace for better debugging
+              newError.stack = err.stack;
+              err = newError;
             }
           }
           util.handleResp(err, response as {} as r.Response, body, callback!);
--- a/src/nodejs-common/util.ts
+++ b/src/nodejs-common/util.ts
@@ -912,6 +912,25 @@ export class Util {
         options,
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         (err: Error | null, response: {}, body: any) => {
+          // Check for a TLS handshake timeout.
+          // This is a conceptual check, the exact error format may vary.
+          if (
+            err &&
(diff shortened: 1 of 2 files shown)


src/nodejs-common/util.ts
```diff
@@ -912,6 +912,38 @@ export class Util {
         options,
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         (err: Error | null, response: {}, body: any) => {
+          // Check for a TLS handshake timeout.
+          // This is a conceptual check, the exact error format may vary.
+          if (
+            err &&
+            (err.message?.toLowerCase().includes('tls handshake') ||
+              err.message?.toLowerCase().includes('timed out') ||
```
v-pratap: The err.message.toLowerCase() call is repeated multiple times inside this block. It would be slightly more efficient to call it once, store the result in a variable, and then use that variable in the if conditions.

test/nodejs-common/util.ts
```diff
@@ -1189,6 +1189,100 @@ describe('common/util', () => {
...
       });
 
+      describe('TLS handshake errors', () => {
+        const error = new Error('🤮');
+
+        beforeEach(() => {
+          authClient.authorizeRequest = async () => {
+            throw error;
+          };
+        });
```
v-pratap: The new tests are very similar in code, with a lot of duplicated code. Is it possible to reuse the code?

thiyaguk09: Thank you for your input. I will review it and respond to you later.

src/nodejs-common/util.ts
```diff
@@ -912,6 +912,38 @@ export class Util {
...
+            if (err.message.toLowerCase().includes('econnreset')) {
+              // ECONNRESET (Connection reset by peer) implies temporary service unavailability
+              code = 503;
+              message =
+                'Connection reset by peer. This suggests the remote service is temporarily unavailable or overloaded.';
+            } else {
+              // TLS Handshake, ETIMEDOUT, or generic 'timed out'
+              // Mapped to 408 (Request Timeout) as a signal for retrying
+              code = 408;
+              message =
```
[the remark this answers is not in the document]
thiyaguk09: Thank you for your input. I will review it and respond to you later.

src/nodejs-common/util.ts
```diff
@@ -912,6 +912,38 @@ export class Util {
...
+            (err.message?.toLowerCase().includes('tls handshake') ||
+              err.message?.toLowerCase().includes('timed out') ||
+              err.message?.toLowerCase().includes('etimedout') ||
+              err.message?.toLowerCase().includes('econnreset'))
+          ) {
+            let code: number;
+            let message: string;
+            if (err.message.toLowerCase().includes('econnreset')) {
+              // ECONNRESET (Connection reset by peer) implies temporary service unavailability
+              code = 503;
```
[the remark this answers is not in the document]
thiyaguk09: That's an insightful point! You're right, TCP/socket errors like ECONNRESET are fundamental network issues and shouldn't be incorrectly treated as HTTP status codes. They reflect a connection failure, not necessarily an application-level server error.

I will check on it and come back.

ddelgrosso1:
I'm not really sure why we are forcing things such as `ECONNRESET`, `ETIMEDOUT` into a TLS error. They may or may not be related to TLS. I think this gives a false impression to the end user. I think we need to rethink what it is we are trying to accomplish here.

thiyaguk09:
> I'm not really sure why we are forcing things such as `ECONNRESET`, `ETIMEDOUT` into a TLS error. They may or may not be related to TLS. I think this gives a false impression to the end user. I think we need to rethink what it is we are trying to accomplish here.

I agree completely—the current grouping is misleading. Thanks for the feedback! 

I've updated the error handling to **distinguish these failures**:
* `"tls handshake"` for specific **protocol issues**.
* `"etimedout"` / `"timed out"` for **network timeouts**.
* `"econnreset"` for **connection resets**.

This provides a much more accurate diagnosis for the end-user.

PR #2707, opened 2026-01-28
Feat: Implement robust path validation and structured skip reporting

### LINKED ISSUE #2660: `downloadManyFiles` can't write to tempdir outside of `cwd`
### Please make sure you have searched for information in the following guides.

- [x] Search the issues already opened: https://github.com/GoogleCloudPlatform/google-cloud-node/issues
- [x] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
- [x] Check our Troubleshooting guide: https://github.com/googleapis/google-cloud-node/blob/main/docs/troubleshooting.md
- [x] Check our FAQ: https://github.com/googleapis/google-cloud-node/blob/main/docs/faq.md
- [x] Check our libraries HOW-TO: https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md
- [x] Check out our authentication guide: https://github.com/googleapis/google-auth-library-nodejs
- [x] Check out handwritten samples for many of our APIs: https://github.com/GoogleCloudPlatform/nodejs-docs-samples

### A screenshot that you have tested with "Try this API".

Not related to the API, just the JS implementation of `TransferManager.downloadManyFiles`.

### Link to the code that reproduces this issue. A link to a **public** Github Repository or gist with a minimal reproduction.

https://gist.github.com/faultyserver/baed2362e93fb2eb8feda06ad2b77813

### A step-by-step description of how to reproduce the issue, based on the linked reproduction.

The above gist can be run just by adding a valid `Storage` reference at the top. In general, the process is:

1. Make a temporary directory, like `path.join(os.tmpdir(), 'some-namespace')`
2. Call `downloadManyFiles` with `passthroughOptions.destination` as that tempdir
3. Run the code, observe a rejected promise with the `TRAVERSAL_OUTSIDE_BASE_DESTINATION` error message.

### A clear and concise description of what the bug is, and what you expected to happen.

`downloadManyFiles` doesn't allow writing to a temporary directory.

I would expect to be able to write a large directory of content into a new directory that I created, especially a temporary directory, to avoid having to write into the current working directory first. Additional use cases this is affecting for me are working in a monorepo where a command may write to a separate cache folder in the same repo that is specified as an absolute path rather than relative from the root of the repository.

### A clear and concise description WHY you expect this behavior, i.e., was it a recent change, there is documentation that points to this behavior, etc. **

This was added in a recent version (~2 weeks ago) in [this PR](https://github.com/googleapis/nodejs-storage/pull/2654/files#diff-8d244cccd9a8777fa7e003c8123ae6bb9638c9c28bae3ac22c730027f317b812R938) and is now preventing downloads to anywhere other than the current working directory, without an option to override or opt out of the behavior for knowably-safe operations, like writing to a temporary directory.

For scripts that are just downloading data to use in other processing, it's common to write to a temporary directory to avoid populating randomly into a project's repository or somewhere persistent and visible to the user.

thiyaguk09:
## Description

> This PR implements a robust security layer for the `downloadManyFiles` method in the `TransferManager`. It addresses potential path traversal vulnerabilities and introduces a structured result reporting system.
>
>**Key Changes:**
>
>* **Structured Reporting:** Changed the return type from `Promise<DownloadResponse[]>` to `Promise<DownloadManyFilesResult>`. This new object includes both successful `responses` and a `skippedFiles` array containing detailed metadata (reason, message, local path).
>* **Strict Security Validation:** Implemented a "Strict Blocking" model for file paths. It prevents:
>   * **Path Traversal:** Blocking dot-segments (`../`) that resolve outside the intended target.
>   * **Absolute Escape:** Blocking object names that start with `/`, `\`, or Windows drive letters to prevent root-level directory escapes.
>   * **Filename Poisoning:** Filtering illegal characters such as Null bytes (`\0`), control characters, and Windows volume separators (`:`).
>
>
>* **Parity Guarantee:** Updated the internal batch loop to ensure every input file is accounted for. If a file is not downloaded due to security or a network error, it is explicitly added to the `skippedFiles` list.

## Impact

> **This is a BREAKING CHANGE.**
Existing implementations that expect an array return from `downloadManyFiles` will need to be updated to destructure the new result object.

## Testing

> 
>**Unit Tests:**
>
>* Added a "Parity Check" test to ensure `inputCount === responses.length + skippedFiles.length`.
>* Verified behavior across simulated Unix and Windows path structures.
>
>**Integration Tests:**
>
>* Updated existing storage samples to handle the new return type and log skipped files.
>* Verified that successful downloads still function as expected with the new parallel limit logic.

## Checklist

Fixes #2660

### DIFF
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -22,6 +22,8 @@ import {
   File,
   FileExceptionMessages,
   RequestError,
+  SkippedFileInfo,
+  SkipReason,
 } from './file.js';
 import pLimit from 'p-limit';
 import * as path from 'path';
@@ -572,11 +574,7 @@ export class TransferManager {
       options.concurrencyLimit || DEFAULT_PARALLEL_DOWNLOAD_LIMIT
     );
     const promises: Promise<DownloadResponse>[] = [];
-    const skippedFiles: Array<{
-      fileName: string;
-      localPath: string;
-      reason: string;
-    }> = [];
+    const skippedFiles: SkippedFileInfo[] = [];
     let files: File[] = [];
 
     const baseDestination = path.resolve(
@@ -622,14 +620,19 @@ export class TransferManager {
 
       const resolvedPath = path.resolve(dest || file.name);
       const relativePath = path.relative(baseDestination, resolvedPath);
-      const isOutside =
-        relativePath.startsWith('..') || path.isAbsolute(relativePath);
+      const isOutside = relativePath.startsWith('..');
+      const isAbsoluteAttempt = path.isAbsolute(relativePath);
 
-      if (isOutside || file.name.includes(':')) {
+      if (isOutside || isAbsoluteAttempt || file.name.includes(':')) {
+        let reason: SkipReason = SkipReason.ILLEGAL_CHARACTER;
+        if (isOutside) reason = SkipReason.PATH_TRAVERSAL;
+        else if (isAbsoluteAttempt) reason = SkipReason.ABSOLUTE_PATH_BLOCKED;
+
         skippedFiles.push({
           fileName: file.name,
           localPath: resolvedPath,
-          reason: 'Path Traversal Detected',
+          reason: reason,
+          message: `File ${file.name} was skipped due to path validation failure.`,
         });
         continue;
       }
@@ -644,15 +647,30 @@ export class TransferManager {
 
       promises.push(
         limit(async () => {
-          const destination = passThroughOptionsCopy.destination;
-          if (destination && destination.endsWith(path.sep)) {
-            await fsp.mkdir(destination, {recursive: true});
-            return Promise.resolve([
-              Buffer.alloc(0),
-            ]) as Promise<DownloadResponse>;
-          }
+          try {
+            const destination = passThroughOptionsCopy.destination;
+            if (
+              destination &&
+              (destination.endsWith(path.sep) || destination.endsWith('/'))
+            ) {
+              await fsp.mkdir(destination, {recursive: true});
+              return [Buffer.alloc(0)] as DownloadResponse;
+            }
 
-          return file.download(passThroughOptionsCopy);
+            return file.download(passThroughOptionsCopy);
+          } catch (err) {
+            const error = err instanceof Error ? err : new Error(String(err));
+            skippedFiles.push({
+              fileName: file.name,
+              localPath: path.resolve(
+                passThroughOptionsCopy.destination || file.name
+              ),
+              reason: SkipReason.DOWNLOAD_ERROR,
+              message: error.message,
+              error: error as Error,
+            });
+            return [Buffer.alloc(0)] as DownloadResponse;
+          }
         })
       );
     }
--- a/src/transfer-manager.ts
+++ b/src/transfer-manager.ts
@@ -16,6 +16,7 @@
 
 import {Bucket, UploadOptions, UploadResponse} from './bucket.js';
 import {
+  DownloadManyFilesResult,
   DownloadOptions,
   DownloadResponse,
   File,
@@ -566,12 +567,21 @@ export class TransferManager {
   async downloadManyFiles(
     filesOrFolder: File[] | string[] | string,
     options: DownloadManyFilesOptions = {}
-  ): Promise<void | DownloadResponse[]> {
+  ): Promise<DownloadManyFilesResult | DownloadResponse[]> {
     const limit = pLimit(
       options.concurrencyLimit || DEFAULT_PARALLEL_DOWNLOAD_LIMIT
     );
     const promises: Promise<DownloadResponse>[] = [];
+    const skippedFiles: Array<{
+      fileName: string;
+      localPath: string;
+      reason: string;
+    }> = [];
     let files: File[] = [];
+
+    const baseDestination = path.resolve(
+      options.prefix || options.passthroughOptions?.destination || '.'
+    );
 
     if (!Array.isArray(filesOrFolder)) {
       const directoryFiles = await this.bucket.getFiles({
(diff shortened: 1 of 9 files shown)


src/transfer-manager.ts
```diff
@@ -598,16 +606,39 @@ export class TransferManager {
...
       }
       if (options.stripPrefix) {
-        passThroughOptionsCopy.destination = file.name.replace(regex, '');
+        dest = file.name.replace(regex, '');
+      }
+
+      const resolvedPath = path.resolve(dest || file.name);
+      const relativePath = path.relative(baseDestination, resolvedPath);
+      const isOutside = relativePath.startsWith('..');
+      const isAbsoluteAttempt = path.isAbsolute(relativePath);
```
krishnamd-jkp: We do not want to block an absolute path download if it is still inside user's input download directory.

test/transfer-manager.ts
```diff
@@ -350,6 +351,163 @@ describe('Transfer Manager', () => {
         true
       );
     });
+
+    it('skips files that attempt path traversal via dot-segments (../) and returns them in skippedFiles', async () => {
+      const prefix = 'safe-directory';
```
krishnamd-jkp: nit: Can rename this to download directory.

test/transfer-manager.ts
```diff
@@ -350,6 +351,163 @@ describe('Transfer Manager', () => {
...
+      const downloadStub = sandbox.stub(file, 'download');
+
+      const result = await transferManager.downloadManyFiles([file], options);
+
+      assert.strictEqual(downloadStub.called, false);
+      const results = result as DownloadManyFilesResult;
+      assert.strictEqual(results.skippedFiles.length, 1);
+    });
+
+    it('should skips files containing Windows volume separators (:) to prevent drive-injection attacks', async () => {
```
krishnamd-jkp: nit: should skip

src/transfer-manager.ts
```diff
@@ -598,16 +606,39 @@ export class TransferManager {
...
+
+      if (isOutside || isAbsoluteAttempt || file.name.includes(':')) {
+        let reason: SkipReason = SkipReason.DOWNLOAD_ERROR;
+        if (isOutside) reason = SkipReason.PATH_TRAVERSAL;
+        else if (file.name.includes(':')) reason = SkipReason.ILLEGAL_CHARACTER;
+        else if (isAbsoluteAttempt) reason = SkipReason.ABSOLUTE_PATH_BLOCKED;
```
krishnamd-jkp: Are we disallowing if the path is absolute here?

test/transfer-manager.ts
thiyaguk09: Fixed

test/transfer-manager.ts
thiyaguk09: Fixed

src/transfer-manager.ts
thiyaguk09: The code had a dead end, so I’ve removed the unnecessary check. Now, we're not blocking absolute paths except for Windows drive paths like `c:`. I've also added a partial check and included more path scenarios in the test cases.

src/transfer-manager.ts
thiyaguk09: The code had a dead end, so I’ve removed the unnecessary check. Now, we're not blocking absolute paths except for Windows drive paths like `c:`. I've also added a partial check and included more path scenarios in the test cases.

krishnamd-jkp (approved):

src/transfer-manager.ts
```diff
@@ -598,16 +606,38 @@ export class TransferManager {
...
       if (options.stripPrefix) {
-        passThroughOptionsCopy.destination = file.name.replace(regex, '');
+        dest = file.name.replace(regex, '');
+      }
+
+      const resolvedPath = path.resolve(dest || file.name);
+      const relativePath = path.relative(baseDestination, resolvedPath);
+      const isOutside = relativePath.startsWith('..');
+      const hasIllegalChars = (dest || file.name).includes(':');
+
```
Pulkit0110: I think `startsWith('..')` will also flag files like `..temp.txt` and `includes(':')` may break GCS filenames with timestamps.

Please double check if that's the intention here

src/transfer-manager.ts
```diff
@@ -566,13 +569,18 @@ export class TransferManager {
   async downloadManyFiles(
     filesOrFolder: File[] | string[] | string,
     options: DownloadManyFilesOptions = {}
-  ): Promise<void | DownloadResponse[]> {
+  ): Promise<DownloadManyFilesResult | DownloadResponse[]> {
```
Pulkit0110: I think we should not change the signature as it may be a breaking change for existing users

src/transfer-manager.ts
```diff
@@ -617,20 +647,39 @@ export class TransferManager {
...
         })
       );
     }
+    const results = await Promise.all(promises);
 
-    return Promise.all(promises);
+    return {
+      responses: results,
+      skippedFiles: skippedFiles,
+    };
```
Pulkit0110: It may break users application as earlier they were supposed to get a list of results but now they'll receive an object.

Pulkit0110 (changes requested):

src/transfer-manager.ts
thiyaguk09: Sure, let me check

src/transfer-manager.ts
thiyaguk09: I’ve reverted the return type change. The method still returns a `Promise<void | DownloadResponse[]>`. Following the Go implementation's pattern, I’ve augmented the response objects with `skipped`, `reason`, and `fileName` properties. This avoids breaking existing user applications while providing the visibility we need.

src/transfer-manager.ts
thiyaguk09: I’ve implemented a more robust traversal check using `path.relative` combined with a `split(path.sep).includes('..')` check. This prevents `..` escapes (including Windows-style backslashes on Linux) while safely allowing files like `..temp.txt`.

I switched the illegal character check to a `regex /^[a-zA-Z]:/`. This specifically blocks Windows drive-letter injections while allowing colons in the middle of filenames (like GCS timestamps) on supported systems."

src/transfer-manager.ts
thiyaguk09: I have reversed the change to the return type, as explained in the comment above.

